PR #1099, opened 2014-12-18
DB2 support

timowest:
Fixes https://github.com/querydsl/querydsl/issues/647

PR #1105, opened 2014-12-20
Reorganize *UseLiterals* tests

timowest:


timowest:
@Shredder121 Have you succeeded with the custom runner? Maybe the custom runner could be handled in a separate pull request or issue?

Shredder121:
I have not yet succeeded, but I am content with putting it off for now.
So shall I merge?
Then the addition of #1099 should be updated as well.

timowest:
Merging is ok.

PR #1109, opened 2015-01-07
Allow arbitrary table types for export (codegen)

### LINKED ISSUE #1108: SQL codegen allow exporting arbitrary table types
Although querydsl has "exportTables" and "exportViews" options, there is no way to export other types for code generation.  Our specific use case is that Postgres recently added support for materialized views  and these are in information schema with the type "MATERIALIZED VIEW", so even though exportViews is set to true we do not get them generated.  It would be useful to be able to specify something like "TABLE, VIEW, MATERIALIZED VIEW".

dharaburda:
Fixes #1108

### DIFF
--- a/querydsl-sql-codegen/src/main/java/com/mysema/query/sql/codegen/MetaDataExporter.java
+++ b/querydsl-sql-codegen/src/main/java/com/mysema/query/sql/codegen/MetaDataExporter.java
@@ -109,6 +109,9 @@ public class MetaDataExporter {
     private boolean exportForeignKeys = true;
 
     private boolean spatial = false;
+    
+    @Nullable
+    private String tableTypesToExport; 
 
     public MetaDataExporter() {}
 
@@ -200,7 +203,18 @@ public class MetaDataExporter {
         }
 
         String[] typesArray = null;
-        if (!exportAll) {
+        
+        if(tableTypesToExport != null && !tableTypesToExport.isEmpty()) {
+        	List<String> types = new ArrayList<String>();
+        	if(tableTypesToExport.contains(",")) {
+        		for(String tableType : tableTypesToExport.split(",")) {
+        			types.add(tableType.trim());
+        		}
+        	} else {
+        		types.add(tableTypesToExport.trim());
+        	}
+        	typesArray = types.toArray(new String[types.size()]);
+        } else if (!exportAll) {
             List<String> types = new ArrayList<String>(2);
             if (exportTables) {
                 types.add("TABLE");
@@ -644,5 +658,12 @@ public class MetaDataExporter {
     public void setSpatial(boolean spatial) {
         this.spatial = spatial;
     }
+
+    /**
+     * @param tableTypesToExport
+     */
+	public void setTableTypesToExport(String tableTypesToExport) {
+		this.tableTypesToExport = tableTypesToExport;
+	}
 
 }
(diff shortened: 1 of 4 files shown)


timowest:
Thanks for the PR! I added some commit comments.

querydsl-sql-codegen/src/main/java/com/mysema/query/sql/codegen/MetaDataExporter.java
```diff
@@ -109,6 +109,9 @@
     private boolean exportForeignKeys = true;
 
     private boolean spatial = false;
+    
+    @Nullable
+    private String tableTypesToExport;
```
timowest: Could this be `List<String>` typed instead, empty list as default

querydsl-sql-codegen/src/main/java/com/mysema/query/sql/codegen/MetaDataExporter.java
```diff
@@ -200,7 +203,18 @@ public void export(DatabaseMetaData md) throws SQLException {
         }
 
         String[] typesArray = null;
-        if (!exportAll) {
+        
+        if(tableTypesToExport != null && !tableTypesToExport.isEmpty()) {
```
timowest: Fix formatting and indentation (4 spaces)

querydsl-sql-codegen/src/main/java/com/mysema/query/sql/codegen/MetaDataExporter.java
dharaburda: The property comes in as a comma-separated list from the pom:

`<tableTypesToExport>TABLE, VIEW</tableTypesToExport>`

and AbstractMetaDataExportMojo just calls the setter.  If we change it to a List<String> here, we would have to tokenize the string in that class instead.  I basically replicated what was done for tableNamePattern, which also allows a comma separated string.

querydsl-sql-codegen/src/main/java/com/mysema/query/sql/codegen/MetaDataExporter.java
```diff
@@ -200,7 +203,18 @@ public void export(DatabaseMetaData md) throws SQLException {
         }
 
         String[] typesArray = null;
-        if (!exportAll) {
+
+        if (tableTypesToExport != null && !tableTypesToExport.isEmpty()) {
+            List<String> types = new ArrayList<String>();
+            if (tableTypesToExport.contains(",")) {
```
timowest: This check seems unnecessary, since split(",") will also work if no comma is contained.

querydsl-sql-codegen/src/main/java/com/mysema/query/sql/codegen/MetaDataExporter.java
timowest: That's true. For consistency let's keep it like you suggested.

timowest:
Could you also update the reference docs and AntMetaDataExporter?

querydsl-sql-codegen/src/main/java/com/mysema/query/sql/codegen/MetaDataExporter.java
Shredder121: If you don't want to tokenize the String, you can also use [this way to automatically receive a `List<String>`](http://maven.apache.org/guides/plugin/guide-java-plugin-development.html#Collections)
I don't know if specifying only one value (`<tableTypesToExport>TABLE</tableTypesToExport>`) would work after that, but if it does, then I think this would improve readability of the pom as well.

timowest:
@Shredder121 Do you think this is ok to merge? The comma separated argument style is consistent with tableNamePattern, so it might be a better choice.

Shredder121:
I think the text of `tableTypesToExport` being mutually exclusive to `exportTables` and `exportViews` would serve well in the Javadoc of AbstractMetaDataExportMojo.

Other than that it is good to merge in my opinion.

PR #1131, opened 2015-01-22
Make QueryMetadata.projection a single Expression

### LINKED ISSUE #610: Make QueryMetadata.projection a single Expression instance
Since multiple projection args are now wrapped into a QTuple expression, the QueryMetadata.projection property should be of type Expression.

timowest:
Fixes #610


querydsl-sql/src/main/java/com/querydsl/sql/DB2Templates.java
```diff
@@ -233,7 +233,7 @@ public void serialize(QueryMetadata metadata, boolean forCountRow, SQLSerializer
                 for (OrderSpecifier<?> os : metadata.getOrderBy()) {
                     rn.orderBy(os);
                 }
-                metadata.addProjection(rn.as("rn"));
+                //metadata.addProjection(rn.as("rn")); FIXME
```
timowest: Needs a fix!

querydsl-sql/src/main/java/com/querydsl/sql/SQLServer2005Templates.java
```diff
@@ -90,7 +90,7 @@ public void serialize(QueryMetadata metadata, boolean forCountRow, SQLSerializer
                 for (OrderSpecifier<?> os : metadata.getOrderBy()) {
                     rn.orderBy(os);
                 }
-                metadata.addProjection(rn.as("rn"));
+                //metadata.addProjection(rn.as("rn")); FIXME
```
timowest: Needs a fix!

timowest:
@Shredder121 Could you take a look at the changes?

Shredder121:
Yes I can, and I will.

querydsl-core/src/main/java/com/querydsl/core/DefaultQueryMetadata.java
```diff
@@ -59,7 +61,8 @@
 
     private List<OrderSpecifier<?>> orderBy = ImmutableList.of();
 
-    private List<Expression<?>> projection = ImmutableList.of();
+    @Nullable
```
Shredder121: I think it's a better idea to pull this annotation up to the interface, if you want to get hints from tools.

querydsl-collections/src/main/java/com/querydsl/collections/DefaultQueryEngine.java
```diff
@@ -188,7 +188,7 @@ private List evaluateSingleSource(QueryMetadata metadata, Map<Expression<?>,
                 order(metadata, sources, list);
             }
             // projection
-            if (metadata.getProjection().size() > 1 || !metadata.getProjection().get(0).equals(source)) {
+            if (!metadata.getProjection().equals(source)) {
```
Shredder121: If you move the annotation to the interface, this would highlight as a possible null pointer dereference.

querydsl-core/src/main/java/com/querydsl/core/DefaultQueryMetadata.java
timowest: I moved the getter annotations now to the interface, but left the `@Nullable` annotations for the fields.

timowest:
I will in the end again try to combine the commits.

querydsl-core/src/main/java/com/querydsl/core/DefaultQueryMetadata.java
```diff
@@ -306,6 +301,9 @@ public void setLimit(Long limit) {
 
     @Override
     public void setModifiers(@Nullable QueryModifiers restriction) {
+        if (restriction == null) {
+            throw new NullPointerException();
```
Shredder121: Why is it marked as `@Nullable`, but must be nonnull? I think that it makes more sense to use `QueryModifiers.EMPTY` if null is specified, than to throw a `NullPointerException`.

Also, if this annotation must carry meaning, consider moving this to the interface as well.

timowest: I removed the `@Nullable` annotation from the argument. I think getters/setters should be mostly side-effect free, that's why  I'd rather not do a silent conversion. This way it is maybe more clear what the valid values for this property are.

Shredder121: Then I think it's best to document this in `QueryBase.restrict()` and others, or annotate the argument as `@Nonnull` to help tools assist the user.

timowest:
I combined now the commits.

timowest:

Shredder121:
Can you maybe remove all `@SuppressWarnings` in UnionImpl?
Because of the raw type `Query` the compiler highlights all the calls, but it is easily fixed.

timowest:
Done

Shredder121:
Looks good now.
I'll merge it when the build succeeds.

PR #1133, opened 2015-01-24
Add javadoc `{@link }`s to hyperlink the original source

Shredder121:
When coding, I often see myself having to click on a property to get to a QObject, and then click through to get to the original model object, but then having to look up the property.
This hyperlinks the original source.


timowest:
Does this PR take into account that some properties are generated from fields and others from getters?

Shredder121:
It should at least.

That reminds me of something.
I also had some trouble with adding links for constructors.
So for now I'll add the progress tag.

timowest:
For constructors the primitive/wrapper differentiation might be a challenge.

Shredder121:
Closing in favor of #1289

Issue #1128, opened 2015-01-22
CaseBuilder .otherwise() exception on .add


renanleandrof:
I wanna do a sort of a sum of two columns.
I need to check if one of the values is null.
Doing a case, i found that the OTHERWISE function bugs when i put an ADD expression on it.

```java
 NumberExpression<BigDecimal> casex = new CaseBuilder()
                .when(e.value1.isNull().and(e.value2.isNull())).then(BigDecimal.ZERO)
                .when(e.value1.isNull()).then(e.value2)
                .when(e.value2.isNull()).then(e.value1)
                .otherwise(e.value1.add(e.value2));

        if (sortOrder.equals(ASC)) {
            query.orderBy(casex.asc());
        } else {
            query.orderBy(casex.desc());
        }
```

17:47:32.154 [main] ERROR org.hibernate.hql.internal.ast.ErrorCounter - line 9:341: expecting "end", found '+' line 9:341: expecting "end", found '+' at antlr.Parser.match(Parser.java:211) ...

I'm not sure if its QueryDSL fault or if it's hibernate.

PS: if i put the add expression on THEN clause, it works fine.

timowest:
Looks like a operator precedence issue.

timowest:
Which version of Querydsl and Hibernate do you use?

renanleandrof:
<versao.hibernate>4.2.16.Final</versao.hibernate>
<versao.querydsl>3.3.3</versao.querydsl>

____________________________
Renan Leandro Ferreira
Tel: (61) 8436-0600
http://www.ipgii.org.br
Sola Scriptura, Sola Gratia, Sola Fide, Solus Christus, Soli Deo Gloria!

timowest:
Could you try upgrading to Querydsl 3.6.0? With the query in the PR I haven't been able to replicate the issue.

renanleandrof:
Nice! It worked on 3.6.0.

I had some work to update it, but now is better ;)

Thanks!

____________________________
Renan Leandro Ferreira
Tel: (61) 8436-0600
http://www.ipgii.org.br
Sola Scriptura, Sola Gratia, Sola Fide, Solus Christus, Soli Deo Gloria!

PR #1135, opened 2015-01-24
Add test case for CaseBuilder usage

timowest:
Related to #1128


timowest:

Shredder121:
Yes, this is good.
We can also backport this, since case expressions should also be validated for Querydsl 3.

PR #1136, opened 2015-01-25
Fix javadoc grouping

timowest:


querydsl-root/pom.xml
```diff
@@ -257,10 +257,18 @@
               <packages>com.querydsl.core*</packages>
             </group>
             <group>
+              <title>Codegem</title>
```
Shredder121: Codegen is probably what you meant.

querydsl-root/pom.xml
```diff
@@ -285,12 +293,16 @@
...
+              <packages>com.querydsl.lucene3</packages>
+            </group>
+            <group>
+              <title>Lucene 4</title>
+              <packages>com.querydsl.lucene4</packages>
             </group>
             <group>
               <title>Hibernate Search</title>
-              <packages>com.querydsl.search</packages>
+              <packages>com.querydsl.hibernate.search</packages>
```
Shredder121: Shouldn't these all have a wildcard?
I admit, there are no more packages, but for consistency (and errors?) it might be better to add them.
What do you think?

timowest: Makes sense.

Shredder121:
Do you also see 2 occurrences of the Spatial group?

Shredder121:
It looks good to me now.
Shall we also backport it to QUERYDSL_3?
Not the package structure of course, but the grouping.

timowest:
Yes, sounds good.

timowest:

Shredder121:
Yes, both this and the backport are good.

PR #1137, opened 2015-01-25
Replace List with Map for updates in SQLUpdateClause

### LINKED ISSUE #1129: Replace List of Pair with LinkedHashMap in SQLUpdateClause

timowest:
Fixes #1129


timowest:
@Shredder121 Could you take at this one as well? Using a map is simpler and guarantees that no duplicate bindings appear.

Shredder121:
Yes, I agree with you.
I think that it is good to merge.
So I'll merge it then.

PR #1140, opened 2015-01-27
Use Q type parameter in MongodbQuery for subclass

### LINKED ISSUE #591: Use Q parameter in MongodbQuery for query subclass
Use Q parameter in MongodbQuery for query subclass

This change shouldn't be applied to the 3.* series since it breaks backwards compatibility

timowest:
Fixes #591


timowest:

timowest:
@Shredder121 Could you take a look at this PR? Two others are kind of depending on this one.

Shredder121:
Yes, looks good to me.
I will merge it now.

Issue #706, opened 2014-04-09
Limit the public API of Querydsl


timowest:
The public API of Querydsl should be more limited then it is currently and internal functionality should be better hidden.

The public runtime API could be something like
* query classes: JPAQuery, SQLQuery etc
* projections: Projections.*
* dynamic expression creation: BooleanBuilder, PathBuilder, Expressions.*
* query factory classes

The following classes should be better hidden
* [x] projection types: QBean, QTuple, ConstructorExpression
* [x] specialized operation types: SimpleOperation etc (at least construction)
* [x] specialized path types: SimplePath etc (at least construction)

timowest:
Here is another change suggestion

* Rename com.querydsl.core.types.expr to com.querydsl.core.types.dsl
* Combine contents of com.querydsl.core.types.expr, path, query and template
* Move com.querydsl.support.Expressions to com.querydsl.core.types.dsl
* Make constructors of SimpleExpression subclasses in those packages protected
* Remove static factory methods from the same classes

@Shredder121 What do you think?

Shredder121:
>- Rename com.querydsl.core.types.expr to com.querydsl.core.types.dsl
>- Combine contents of com.querydsl.core.types.expr, path, query and template
>- Move com.querydsl.support.Expressions to com.querydsl.core.types.dsl

That sounds good.

>- Make constructors of SimpleExpression subclasses in those packages protected

Yes, that sounds good, as long as they are all still instantiable in some way.

>- Remove static factory methods from the same classes

What do you mean by this? Do you mean that all instantiation is done in Projections and Expressions?

timowest:
> What do you mean by this? Do you mean that all instantiation is done in Projections and Expressions?

Yes, that's the reason.

Shredder121:
Then that sound good as well.

PR #1141, opened 2015-01-28
Hide Projection constructors

timowest:
Related to #706


timowest:

Shredder121:
I only now just noticed.
`AppendingFactoryExpression` doesn't define a `serialVersionUID`.
I admit, its implementation might not change, but do you want to add it before it is released?

timowest:
Added it right now.

Shredder121:
Then these changes are also done.
So I shall merge this then.

PR #1143, opened 2015-01-28
Move NullExpression creation to Expressions

### LINKED ISSUE #1130: Move NullExpression creation to Expressions

timowest:
Fixes #1130


timowest:

Shredder121:
You can make the `NullExpression` a singleton, then you can make the `nullExpression()` method generic. Like so:

```java
@SuppressWarnings("unchecked")//does not produce non-null instances of T
public static <T> NullExpression<T> nullExpression() {
    return (NullExpression<T>) NullExpression.DEFAULT;
}
```
You can keep the type token generic methods for ease of use, since some people don't like the verbose syntax of explicit type arguments. But you could even go as far as this:
```java
@SuppressWarnings("unchecked")//does not produce non-null instances of T
public static <T> NullExpression<T> nullExpression() {
    return (NullExpression<T>) NullExpression.DEFAULT;
}

public static <T> NullExpression<T> nullExpression(Class<T> type) {
    return nullExpression(); //return type captured via the type token
}

public static <T> NullExpression<T> nullExpression(Path<T> path) {
    return nullExpression(); //return type captured via the type token
}
```
But that is up to you.
My point is, all compiler inserted casts are valid, since null can be cast to any object.

```java
public class SelectBase extends AbstractBaseTest {
...
    @Test
    @ExcludeIn(DERBY)
    public void Null() {
        Number result = query().from(employee).singleResult(Expressions.<Integer>nullExpression());
        assertNull(result);
    }
...
}
```

timowest:
Yes, that makes sense. In Querydsl SQL the actual type of the Null argument is relevant, but it is obtained differently.

timowest:
@Shredder121 This contains now as well your requested changes.

Shredder121:
Yes, the only thing I would like is change the Javadoc for the `nullExpression()` methods.
It might both be just `get the nullexpression for the specified type`.
I don't know if you feel that way as well? But at least functionality wise we can merge.
Tip of the day: [[ci skip]](http://docs.travis-ci.com/user/how-to-skip-a-build/).

timowest:
Updated the docs. And thanks for the tip! ;)

Shredder121:
Yes I found it interesting and thought you would think so as well.
It even works in the body part of the commit message, so it shouldn't have to appear in normal `git log`s, which is nice.

PR #1145, opened 2015-01-28
PathMetadata improvements

### LINKED ISSUE #893: Impossible to use Jackson into an EntitybasePath
I tried to use jackson into a EntityBasePath :

```java 
ObjectMapper sMapper = new ObjectMapper();
QContent rootNode = QContent.content;
String json = sMapper.writeValueAsString(rootNode);
```
It seems does not work. There is a getter conflict into class com.mysema.query.types.PathMetadata (full stack bellow). Do you know why ?

Caused by: java.lang.IllegalArgumentException: Conflicting getter definitions for property "root": com.mysema.query.types.PathMetadata#isRoot(0 params) vs com.mysema.query.types.PathMetadata#getRoot(0 params)

timowest:
Fixes #893

### DIFF
--- a/querydsl-codegen/src/main/java/com/querydsl/codegen/EntitySerializer.java
+++ b/querydsl-codegen/src/main/java/com/querydsl/codegen/EntitySerializer.java
@@ -13,12 +13,15 @@
  */
 package com.querydsl.codegen;
 
-import javax.annotation.Generated;
-import javax.inject.Inject;
-import javax.inject.Named;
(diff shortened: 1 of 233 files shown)


timowest:
@Shredder121 How does this look like? It touches so many files that it would be nice to merge before new work.

Shredder121:
Yes, I agree and it looks good. Nicely done.
So I will merge it then. (assuming the CI builds won't break)

PR #1149, opened 2015-01-31
Fix maven artifacts.

iliya-gr:


timowest:
The master branch contains already the 4.0.0 contents.

PR #1152, opened 2015-02-02
Fix Column handling

### LINKED ISSUE #1151: @Column annotation with no name property causes incorrect query to be created. Using JPASQLQuery.
For example I have a Company class which I want to select the name from (this is a contrived example but displays the bug):

```java
@Entity
public class Company {

  @Id
  @GeneratedValue
  private Long id;
  @Column(unique = true, nullable = false)
  private String name;
}
```
```java
JPASQLQuery sql = query.from(company).limit(10);
sql.list(company.name);
```
Will generate a query like
```sql
select company. from company limit ?
```

What I would have expected would be to use the field name as if the @Column annotation didn't exist in this case.

timowest:
Fixes #1151

### DIFF
--- a/querydsl-jpa/src/main/java/com/querydsl/jpa/NativeSQLSerializer.java
+++ b/querydsl-jpa/src/main/java/com/querydsl/jpa/NativeSQLSerializer.java
@@ -55,10 +55,9 @@ public final class NativeSQLSerializer extends SQLSerializer {
     @Override
     protected void appendAsColumnName(Path<?> path, boolean precededByDot) {
         if (path.getAnnotatedElement().isAnnotationPresent(Column.class)) {
-            SQLTemplates templates = getTemplates();
             Column column = path.getAnnotatedElement().getAnnotation(Column.class);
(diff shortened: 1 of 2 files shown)


querydsl-jpa/src/main/java/com/querydsl/jpa/NativeSQLSerializer.java
```diff
@@ -57,7 +57,11 @@ protected void appendAsColumnName(Path<?> path, boolean precededByDot) {
         if (path.getAnnotatedElement().isAnnotationPresent(Column.class)) {
             SQLTemplates templates = getTemplates();
```
Shredder121: These templates are only used in one branch of the if block, so you *could* move them in there.

Shredder121:
I think that the fix is good, so I'll merge it, but I first want to hear your opinion on my suggestion.

timowest:
Yes, your suggestion was valid and I made the change.

Shredder121:
Then I'll merge them when the build completes.

PR #1154, opened 2015-02-03
Add test case for CaseBuilder usage

timowest:
Related to #1128


Shredder121:
This looks good to me, and since it is only a backport, it can be merged.

PR #1158, opened 2015-02-05
Add GroupBy.sortedSet and sortedMap

### LINKED ISSUE #1116: GroupBy.sortedSet
GroupBy.sortedSet: GSortedSet seems to be missing.

timowest:
Fixes #1116

### DIFF
--- a/querydsl-core/src/main/java/com/querydsl/core/group/GroupBy.java
+++ b/querydsl-core/src/main/java/com/querydsl/core/group/GroupBy.java
@@ -13,6 +13,7 @@
  */
 package com.querydsl.core.group;
 
+import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -125,13 +126,22 @@ public final class GroupBy {
      * @param expression
      * @return
      */
-    public static <E> AbstractGroupExpression<E, Set<E>> sortedSet(Expression<E> expression) {
+    public static <E extends Comparable<? super E>> AbstractGroupExpression<E, Set<E>> sortedSet(Expression<E> expression) {
         return GSet.createSorted(expression);
     }
 
-    public static <E, F> GroupExpression<E, Set<F>> sortedSet(GroupExpression<E, F> groupExpression) {
+    public static <E, F extends Comparable<? super F>> GroupExpression<E, Set<F>> sortedSet(GroupExpression<E, F> groupExpression) {
         return new MixinGroupExpression<E, F, Set<F>>(groupExpression, GSet.createSorted(groupExpression));
     }
+
+    public static <E> AbstractGroupExpression<E, Set<E>> sortedSet(Expression<E> expression, Comparator<? super E> comparator) {
+        return GSet.createSorted(expression, comparator);
+    }
+
+    public static <E, F> GroupExpression<E, Set<F>> sortedSet(GroupExpression<E, F> groupExpression, Comparator<? super F> comparator) {
+        return new MixinGroupExpression<E, F, Set<F>>(groupExpression, GSet.createSorted(groupExpression, comparator));
+    }
+
 
     /**
      * Create a new aggregating map expression using a backing LinkedHashMap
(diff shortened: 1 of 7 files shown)


Shredder121:
You beat me to the implementation.

Notes and ideas:
- NEED to enforce the generic type of the `GroupBy.sorted*()` to `Comparable<? super E>`, otherwise ClassCastExceptions can occur.
- Add another pair of `GroupBy.sorted*()` that takes a comparator, this method doesn't have to be bounded to `Comparable`.
- Are those bridge methods necessary for newly added methods?

Other than that, it looks very nice already.

Shredder121:
Is it possible to add a `BASIC_RESULTS_UNORDERED` in `AbstractGroupByTest`, which is `BASIC_RESULTS` with the rows shuffled?
Then you can add SortedSet and SortedMap tests, and verify they work.

timowest:
Yes, I added the tests now.

Shredder121:
Maybe add some tests with a comparator? like reverse natural ordering?

timowest:
Good point. Added them.

querydsl-core/src/main/java/com/querydsl/core/group/GroupBy.java
```diff
@@ -106,21 +107,44 @@
...
+     */
+    public static <E extends Comparable<? super E>> AbstractGroupExpression<E, Set<E>> sortedSet(Expression<E> expression) {
+        return GSet.createSorted(expression);
+    }
+
+    public static <E, F extends Comparable<? super F>> GroupExpression<E, Set<F>> sortedSet(GroupExpression<E, F> groupExpression) {
+        return new MixinGroupExpression<E, F, Set<F>>(groupExpression, GSet.createSorted(groupExpression));
+    }
+
+    public static <E> AbstractGroupExpression<E, Set<E>> sortedSet(Expression<E> expression, Comparator<? super E> comparator) {
```
Shredder121: Does this sound good?
'Create a new aggregating set expression using a backing TreeSet using the given comparator'

Shredder121:
The last idea I have is not possible using the current logic, but I don't know how valuable it is to end users.
When you see `GroupBy.sortedSet()` you would think it's result type would be `SortedSet`. The same goes for `SortedMap`.

@ssaarela What do you think? Do you think this is necessary (for your use case at least)?

ssaarela:
I would definitely expect the result to be at least SortedSet/Map, or maybe even NavigableSet/Map which is the most specialized interface implemented by TreeSet/Map and assignable to SortedSet/Map. No need to support older Java than 1.6, right?

Having an alternative with custom Comparable is also great addition as that is also part of SortedSet/Map contract.

Shredder121:
I think that `SortedSet` and `SortedMap` makes the most sense, since it's in the method name.

Yes, we added the custom Comparator one already, I also agree that since it's part of the contract it should be included.

timowest:
Now these changes are also in.

querydsl-core/src/test/java/com/querydsl/core/group/GroupByMapTest.java
```diff
@@ -45,6 +36,30 @@ public void Group_Order() {
...
+        Group group = results.get(1);
+        Iterator<Integer> it = group.getSet(commentId).iterator();
+        assertEquals(1, it.next().intValue());
+        assertEquals(2, it.next().intValue());
+        assertEquals(3, it.next().intValue());
+    }
+
+    @Test
+    public void Set_By_Sorted_Reverse() {
+        Map<Integer, Group> results = BASIC_RESULTS
```
Shredder121: Could you maybe use `BASIC_RESULTS_UNORDERED` here instead of the already sorted `BASIC_RESULTS`?

The same for `Set_By_Sorted()`.

Shredder121:
I think that at least the functionality works good, and the type parameters also work nice.
After all of this, the only thing that remains is having `getSortedSet()` and `getSortedMap()` in interface `Group`, for retrieval of the actual result.

timowest:
@Shredder121  Do you think this is good to merge now?

Shredder121:
Are the `getSorted*()` methods not necessary?
If there is a composite `Group` I think it is still nice to get a `SortedSet` out of the results.

timowest:
Added them.

Shredder121:
Yes. Looks good now.
So it is good to merge now in my opinion.

timowest:
Good. In this case a backport might not be so easy, because the signature of some classes has changed. So maybe we can skip it in this case?

Shredder121:
You'll be surprised. The erasures actually stay the same.
If you run these commands:

```bash
#export signatures (-s) of all class files in group package
$ javap -s querydsl-core/target/classes/com/querydsl/core/group/* > newSignatures

#go back to the status quo
$ git reset (--hard) cc27741cf311cf5db11aca5c656ab2217dbb614b

#compile, then export again
$ javap -s querydsl-core/target/classes/com/querydsl/core/group/* > oldSignatures

#compare word by word, the contents of each export
git diff --no-index --color-words oldSignatures newSignatures

```

You can see that the `Signature:  ` lines are all the same, or new.
This means that the erasures are the same.

Shredder121:
`M` is erased to `java.util.Map` in `get()`.
`S` is erased to `java.util.Set` in `get()`.
Aside from that, all methods are new, so no changes whatsoever.

Shredder121:
@timowest have you looked at the signatures?
It is fine if you still want to skip the backport, but I recommend at least taking a look at the diff.

timowest:
Yes, I can do the backport.

PR #1170, opened 2015-02-07
Switch to jdk6 for tests

### LINKED ISSUE #1169: Make sure that the code is compatible with JDK 1.6
Travis builds with different JDK versions doesn't cover this requirement, since Querydsl can't be built anymore with JDK 1.6. Still the result should be compatible with JDK 1.6.

This looks like a promising solution http://eclipsesource.com/blogs/2014/05/09/ensure-class-library-compatibility-in-a-maven-build/

timowest:
Fixes #1169

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,5 +15,6 @@ before_script:
   - ./travis/cubrid.sh
   - ./travis/firebird.sh
 script:
+  - jdk_switcher use openjdk6
   - sh -c 'cd querydsl-root && mvn -B test -Pall,travis'
 


Shredder121:
Sun JDK6 is not provided.
You could use OpenJDK 6 instead.

Shredder121:
Hmm, it still compiles the project.

Then the maven plugin signature check sounds like a good solution.
Shall we close this then?

Shredder121:
It still tries to compile the project using JDK6.
I think that the Maven plugin that checks the signatures is a good alternative.

It also runs client-side, so it's validated early.

timowest:
Ok, let's try that instead.

PR #1171, opened 2015-02-07
Sync Maven and Ant plugin configuration

### LINKED ISSUE #1165: Documentation around ant based sql code generation needs some love
I'm using gradle so I'm using the ant codegen stuff. Three things in particular tripped me up.

1. The `querydsl-sql-codegen` module should be mentioned since that's where the `AntMetaDataExporter` resides.
2. The parameters for `com.mysema.query.sql.ant.AntMetaDataExporter` are not the same as the maven parameters. `dbDriverClass` instead of `jdbcDriver`, `dbUrl` instead of `jdbcUrl`, etc.
3. The parameter `targetSourceFolder` is required. A `NullPointerException` is thrown without it.

I've opened a few bugs now but I haven't said thanks to you guys for this project. It's going to be super helpful in my projects and it works really well once I've figured some things out.

timowest:
Fixes #1165

### DIFF
--- a/querydsl-sql-codegen/src/main/java/com/querydsl/sql/codegen/ant/AntMetaDataExporter.java
+++ b/querydsl-sql-codegen/src/main/java/com/querydsl/sql/codegen/ant/AntMetaDataExporter.java
@@ -19,13 +19,14 @@ import java.sql.DriverManager;
 import java.sql.SQLException;
 import java.util.Comparator;
 
+import org.apache.tools.ant.BuildException;
+import org.apache.tools.ant.Task;
+
 import com.mysema.codegen.model.SimpleType;
 import com.querydsl.codegen.BeanSerializer;
 import com.querydsl.sql.codegen.DefaultNamingStrategy;
 import com.querydsl.sql.codegen.MetaDataExporter;
 import com.querydsl.sql.codegen.NamingStrategy;
-import org.apache.tools.ant.BuildException;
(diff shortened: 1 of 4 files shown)


Shredder121:
[Tutorial: Writing Tasks](http://ant.apache.org/manual/tutorial-writing-tasks.html) doesn't mention `@parameter` doclet tags, so I think these do nothing.

querydsl-sql-codegen/src/main/java/com/querydsl/sql/codegen/ant/AntMetaDataExporter.java
```diff
@@ -37,190 +38,250 @@
...
+
 
     @Override
     @SuppressWarnings({ "unchecked", "rawtypes" })
     public void execute() {
         Connection dbConn = null;
-        File targetPackagePath = new File(targetSourceFolder);
+        File targetPackagePath = new File(targetFolder);
+
+        if (targetFolder == null) {
```
Shredder121: You should move this null check to before you use the variable.

timowest:
Made the required changes.

Shredder121:
there are a few occurrences of suffix changed to prefix.

timowest:
Fixed those as well.

timowest:
@Shredder121 Is this now good to merge?

Shredder121:
I believe so yes.

PR #1172, opened 2015-02-07
Verify jdk6 compatibility in test phase

### LINKED ISSUE #1169: Make sure that the code is compatible with JDK 1.6

timowest:
Fixes #1169

### DIFF
--- a/querydsl-root/pom.xml
+++ b/querydsl-root/pom.xml
@@ -366,6 +366,27 @@
           </dependency>
         </dependencies>
       </plugin>
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>animal-sniffer-maven-plugin</artifactId>
+        <version>1.13</version>
+        <configuration>
+          <signature>
+            <groupId>org.codehaus.mojo.signature</groupId>
+            <artifactId>java16</artifactId>
+            <version>1.1</version>
+          </signature>
+        </configuration>
+        <executions>
+          <execution>
+            <id>ensure-java-1.6-class-library</id>
+            <phase>test</phase>
+            <goals>
+              <goal>check</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
     </plugins>
   </build>
 


Shredder121:
(*Remember that the build must fail for it to succeed, so it's a little counterintuitive, when #1164 is merged, remerge it to test for successful behavior*)

Shredder121:
Strange.. You would think that `JaveSE7SQLExceptionWrapper` would trigger as well.

I think we need to put the annotations of that plugin on the classpath as well. Then we could annotate elements like those with `@IgnoreJRERequirement`.
Do you want to do that?

timowest:
Yes, we need to wait for the #1164 to be merged into it.

timowest:
I can confirm that it doesn't catch the `Throwable.addSuppressed(Throwable)` usage.

Shredder121:
Maybe we should submit a bug report, linked to our code as being a test case?
Maybe it launches a JVM under JDK6 and since it IS runtime compatible doesn't break?

At least, users don't encounter that method anyway, so it is not that big of a deal, but maybe it was supposed to discover this.

Shredder121:
Could you keep me updated if the new version works?

timowest:
It doesn't work better, but I felt it would be good to use a newer version anyway.

timowest:
I tried it with a few others from this list and it also doesn't complain about them: http://www.javaworld.com/article/2074087/core-java/jdk-7--new-interfaces--classes--enums--and-methods.html

e.g.
```
Collections.emptyIterator();
Byte.compare((byte)1, (byte)2);
```

Shredder121:
Agreeably so.
Then I think it's valuable for the developers to find out exactly why it is not picked up.

-------- Original message --------

Shredder121:
I wanted to say it only checks static methods, but it might be that it only checks class definitions?
It doesn't make sense, since it says with precision that it is java.util.Objects.equals().
Maybe the definitions are not updated?



-------- Original message --------

timowest:
I was wrong. I tested with a class under src/test/java and that was not picked up. The throwable issue might be caused by this https://jira.codehaus.org/browse/MANIMALSNIFFER-20

timowest:
I can confirm that it is caused by it.

Shredder121:
So since the statement is actually `QueryException.addSuppressed()` it is not picked up?


-------- Original message --------

timowest:
Yes, maybe this approach is still safer than running the tests with JDK 1.6?

Shredder121:
Of course it is, and I CAN run the tests just fine, I dont compile the project again.
That's how I made the exception wrapper.

This is much faster, and notifies you straight away, before you push.
So I am all for it.

-------- Original message --------

timowest:
Yes, and eventually the bug might be fixed anyway.

Issue #1157, opened 2015-02-05
Make all mapping options of com.querydsl.sql.Configuration available in the Maven plugin


timowest:
Renaming (via RenameMapping)
* schema -> schema
* table  -> table
* schema, table -> table
* schema, table -> schema, table
* schema, table, column -> column
* table, column -> column

Numerics (via NumericMapping)
* total, decimal -> java type

Custom types (via TypeMapping)
* table, column -> java type
* schema, table, column -> java type

jbrackett:
Hi Timo, is this ticket related to what we discussed on stackoverflow?

http://stackoverflow.com/questions/28331489/querydsl-join-on-same-table-multiple-times/28349997?noredirect=1#comment45050781_28349997

Shredder121:
I don't believe it is.
Your issue is more related to the usage of entities where it generally expects other types.
You can open a new ticket for your issue, probably named `document the use of non-interchangeable types`?

jbrackett:
Thanks, will do.

timowest:
@jbrackett I can confirm that this not related to your SO question. It is more about what configuration options to expose via the maven plugin.

PR #1175, opened 2015-02-07
Extract NameMapping out of Configuration

timowest:
Related to https://github.com/querydsl/querydsl/issues/1157


timowest:
@Shredder121 Could you take a look at this issue as well? I don't want to run into conflicts when I tackle this https://github.com/querydsl/querydsl/issues/1194

querydsl-sql/src/main/java/com/querydsl/sql/Configuration.java
```diff
@@ -84,7 +75,7 @@ public Configuration(SQLTemplates templates) {
             javaTypeMapping.register(customType);
         }
         for (Map.Entry<SchemaAndTable, SchemaAndTable> entry : templates.getTableOverrides().entrySet()) {
-            schemaTables.put(entry.getKey(), entry.getValue());
+            nameMapping.registerTableOverride(entry.getKey(), entry.getValue());
```
Shredder121: The only thing I see is since you make delegate methods to NameMapping, this should use the `registerTableOverride()` method for consistency.

Shredder121:
For the rest it looks good, only a few very compact (and slightly questionable)  containsKey alternatives, but that was already that way?
Then I don't have anything else, and it looks good.

timowest:
Changed to the delegate method usage.

Shredder121:
Then I think it's good to merge when the build completes.

PR #1176, opened 2015-02-07
Combine QueryMixin.convert and normalize

### LINKED ISSUE #1067: Combine QueryMixin.convert and normalize methods
Combine QueryMixin.convert and normalize methods

timowest:
Fixes #1067


timowest:
@Shredder121 This touches also a few files, so would be good to merge before bigger changes. Could you take a glimpse?

Shredder121:
It looks good, but maybe try to keep unrelated classes out of the commits.
As you said, it touches a few files, and the more it touches(and the more unrelated things it solves), the bigger chance you fix it twice across different branches.

That said, it still looks good.
I will merge it.

PR #1177, opened 2015-02-07
Simplify Operator usage

timowest:
The benefits of this simplified Operator usage are
* simpler instance control through enums
* less code for declaring Operator instances

The downsides are
* No generic type for operation return type anymore


timowest:
@Shredder121 What do you think of this solution approach?

In addition to Date/Time types also Spatial types (JTS vs Geolatte) make Operator type usage difficult. It seems to be useful for `String`, `Boolean` and `Number`.

Shredder121:
I think it is at least not as bad as I thought, although I know of some improvements.

Since there is no generic return type anymore, I think it is important to realize there must be some validation.
Maybe there can be at creation time (in the static factory methods sounds good enough) a validation that the type of the operator matches (roughly for Date/Time and Spatial types) the type the expression wants.

This roughly translates into an interface like this:
```java
public interface Operator extends Serializable{

    String name();

    Class<?> getType();
    
}
```
This allows us to query the Operator like this:
```java
public static <D extends Number & Comparable<?>> NumberExpression<D> create(Class<? extends D> type, Operator op, Expression<?> one) {
    Preconditions.checkArgument(op.getType().isAssignableFrom(type));
    return new NumberOperation<D>(type, op, ImmutableList.<Expression<?>>of(one));
}
```
This prevents operations like this:

```java
NumberExpression<Integer> num = Expressions.numberOperation(Integer.class, Ops.UPPER, one);
```

I like fail-fast behavior better than undefined behavior at runtime, since this only fails at `AbstractNumberType.getValue():40`, whereas it could have failed with any ClassCastException elsewhere.

Of course, as for third-party types, this is really difficult, but at least this way we don't lose all safety that generics 'promised' us.
What do you think?

timowest:
Sounds good, I will do the necessary changes.

timowest:
@Shredder121 How about now?

Shredder121:
Does `com.querydsl.scala.Operations.number()` declare a new operator? Or how do I read that?

timowest:
`Operations` contains factory methods for `Operation` creation.

timowest:
I fixed now the remaining issue on the Scala side. Expressions on the Scala side use also primitive types, since Scala supports primitive types as generic type parameters.

The fix normalizes the type of OperationImpl to the wrapper type before comparison.

Shredder121:
Ah I see, that indeed sounds like the problem.

timowest:
For consistency it would be nice to have the same constraints for expression types for both Java and Scala based Querydsl usage, but on the other hand they are different languages and this way Querydsl matches better the expressivity of Scala with a little added complexity in some places.

Shredder121:
Since there is no type parameter anymore, we now have a few SuppressWarnings less.

timowest:
@Shredder121 Do you think this is good to merge now, or do you still see other improvements that could be applied?

Shredder121:
I think we are there. I don't know of any further (important) improvements.
So then I will merge it, if you didn't come up with anything in the meantime?
You can merge it as well for once, if you think it's good.

timowest:
I might still make some changes to the Scala operation usage, but that can be independent of this PR.

PR #1182, opened 2015-02-07
Fix String cast mapping

### LINKED ISSUE #1106: String cast generates invalid SQL using EclipseLink and MySQL
Eclipselink uses the template 'cast({0} varchar(255))' for the STRING_CAST function.  This does not work when the underlying database is MySQL.  MySQL uses the type 'char' rather than 'varchar'.  The correct template in this situation would be 'cast({0} char(255))'.

timowest:
Fixes #1106

### DIFF
--- a/querydsl-jpa/pom.xml
+++ b/querydsl-jpa/pom.xml
@@ -25,7 +25,7 @@
   <properties>
     <hibernate.version>4.3.7.Final</hibernate.version>
     <hibernate.validator.version>4.3.1.Final</hibernate.validator.version>
-    <eclipselink.version>2.5.1</eclipselink.version>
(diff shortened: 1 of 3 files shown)


timowest:
The ToChar operator doesn't seem to work as documented.

Shredder121:
Now `.append("test")` chains don't work at all anymore.
Maybe add a trim?

timowest:
I added now a test for `append` usage.

timowest:
And added now the necessary trim as well. At least HSQLDB and Derby required it.

Shredder121:
That is indeed what I noticed.

Shredder121:
So, are you content with the outcome? Shall I merge then?

timowest:
Yes, this is fine with me.

PR #1184, opened 2015-02-07
Fix querydsl-docs dependencies

### LINKED ISSUE #1183: querydsl-docs fails maven build
[ERROR] Failed to execute goal org.jboss.maven.plugins:maven-jdocbook-plugin:2.3.5:resources (default-resources) on project querydsl-docs: Unexpected problem: Unable to locate artifact [org.jboss.jdocbook:jdocbook-core:jar:1.0.4]: Could not find artifact org.jboss.jdocbook:jdocbook-core:jar:1.0.4 in querydsl (http://www.querydsl.com/maven)

timowest:
Fixes #1183

### DIFF
--- a/querydsl-docs/pom.xml
+++ b/querydsl-docs/pom.xml
@@ -91,4 +91,27 @@
     </pluginRepository>
   </pluginRepositories>
 
+  <repositories>
+    <repository>
+      <id>jboss</id>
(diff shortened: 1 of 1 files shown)


Shredder121:
Yes, now that I see it, we indeed miss these artifacts now.

PR #1186, opened 2015-02-08
Optimize surefire config in travis profile

timowest:


timowest:
@Shredder121 Could you take a look at this PR? It should make the handling of Travis logs simpler.

Shredder121:
There is also a `<usefile>` parameter, do you know by chance if travis outputs the console? it looks like it.
I think disabling as much output as possible could(?) help.

Shredder121:
Other than that it looks good.

timowest:
Good catch. `<usefile>` disabled `<printSummary>false</printSummary>`, when I tried it, that's why I didn't include it.

Shredder121:
If that is the case, then it doesn't matter.
A little more I/O isn't that bad.
In my opinion it is good to merge.

PR #1188, opened 2015-02-08
Fix Id handling

### LINKED ISSUE #1187: mongodb - eq expression with objectId generates wrong path
If one has an entity person with a manual handled reference (without the morphia ```@Reference``` annotation) saving an ObjectId in the object for a reference to an Address entity as shown below.

```java
class Person {
   @Id
   ObjectId id;

   String name;

   // manual reference to an address
   ObjectId addressId;
}
```

the morphia serializer doesn't respect the query:
```java
QPerson.person.addressId.eq(<objectId>)
```

as shown in the code in the com.mysema.query.mongodb.MongodbSerializer#getKeyForPath:
```java
    protected String getKeyForPath(Path<?> expr, PathMetadata<?> metadata) {
        if (expr.getType().equals(ObjectId.class)) {
            return "_id";
        } else {
            return metadata.getElement().toString();
        }
    }
```
every path with an objectId type on field automatically gets translated to ```_id``` instead of the ```addressId``` as shown in the query path expression.

timowest:
Fixes #1187

### DIFF
--- a/querydsl-mongodb/src/main/java/com/querydsl/mongodb/MongodbSerializer.java
+++ b/querydsl-mongodb/src/main/java/com/querydsl/mongodb/MongodbSerializer.java
@@ -18,30 +18,13 @@ import java.util.List;
 import java.util.regex.Pattern;
 
 import org.bson.BSONObject;
-import org.bson.types.ObjectId;
 
 import com.google.common.collect.Sets;
 import com.mongodb.BasicDBList;
 import com.mongodb.BasicDBObject;
 import com.mongodb.DBObject;
 import com.mongodb.DBRef;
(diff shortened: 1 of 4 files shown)


timowest:
@Shredder121 Could you also take a look at this pair of issues? The querydsl3 version is slightly different because of Spring Data compatibility issues.

Shredder121:
Yes you said that indeed.
The only thing I noticed is that the new file isn't terminated by a newline.

Not that it matters much in test files, but just so you know.

Shredder121:
But it looks good to me nonetheless. I will merge them.

timowest:
You are right. I didn't notice.

PR #1195, opened 2015-02-10
Make template methods abstract

timowest:


timowest:
@Shredder121 Here is a small one to review to make the PR list shorter ;)

Shredder121:
Yes, indeed.
And it looks good as well ;)
I'll merge it.

PR #1199, opened 2015-02-12
Simplify visitor signature

timowest:
The `Nullable` usage here caused lots of unneeded warnings in the implementation classes.


Shredder121:
It's a shame that type annotations were added so late, it would have worked perfectly here.

Anyway, how do you see the nullability now? Do Visitors return nonnull references? Or is it implementation dependent(via Void)?

timowest:
Implementation dependent. It shouldn't be centrally defined.

Issue #1025, opened 2014-10-31
Numeric type mappings should be used for all numeric types, not only Types.NUMERIC


timowest:
Numeric type mappings should be used for all numeric types, not only Types.NUMERIC

Related post https://groups.google.com/forum/#!topic/querydsl/MCZUCQG7ico

PR #1200, opened 2015-02-12
Improve numeric mappings

### LINKED ISSUE #1194: Default mapping for numeric types are incorrect
`NUMERIC` and `DECIMAL` should map to `BigDecimal` by default, rather than the current mapping of

17-...,?   -> BigDecimal
0-16,?     -> Double

which causes non-floating pointing numbers to be mapped to floating points number, which should never be used to represent exact values. See 
http://stackoverflow.com/questions/28371179/floating-point-types-returned-in-orm-dsl, http://docs.oracle.com/cd/B19306_01/java.102/b14188/datamap.htm, http://www.cs.mun.ca/~michael/java/jdk1.1.5-docs/guide/jdbc/getstart/mapping.doc.html

timowest:
Fixes https://github.com/querydsl/querydsl/issues/1194 and https://github.com/querydsl/querydsl/issues/1025


querydsl-sql/src/main/java/com/querydsl/sql/JDBCTypeMapping.java
```diff
@@ -117,11 +129,8 @@ public void registerNumeric(int total, int decimal, Class<?> javaType) {
...
-    private Class<?> getNumericClass(int total, int decimal) {
-        Pair<Integer,Integer> key = Pair.of(total, decimal);
-        if (numericTypes.containsKey(key)) {
-            return numericTypes.get(key);
-        } else if (decimal <= 0) {
+    private static Class<?> getNumericClass(int total, int decimal) {
+        if (decimal <= 0) {
             if (total > 18 || total == 0) {
                 return BigInteger.class;
             } else if (total > 9 || total == 0) {
```
Shredder121: Total can only be 0 once.
Is this what was intended?

querydsl-sql/src/main/java/com/querydsl/sql/JDBCTypeMapping.java
```diff
@@ -136,19 +145,21 @@ public void registerNumeric(int total, int decimal, Class<?> javaType) {
                 return Boolean.class;
```
Shredder121: Boolean is never returned, was that intended?

querydsl-sql/src/test/java/com/querydsl/sql/JDBCTypeMappingTest.java
```diff
@@ -69,8 +69,7 @@ public void NumericTypes() {
 //        2,0        -> Byte
 //        0-1,0      -> Boolean
```
Shredder121: So now we can update these comments, since Boolean isn't used in numeric types?

Shredder121:
In my opinion this is now also good to merge.

timowest:
Ok, good.

PR #1206, opened 2015-02-16
Fix String cast mapping

### LINKED ISSUE #1106: String cast generates invalid SQL using EclipseLink and MySQL

timowest:
Fixes #1106


Shredder121:
And I will merge this when the build finishes, any moment now.

PR #1207, opened 2015-02-16
Access JodaTime types directly

timowest:


Shredder121:
What is the motivation for this change?

timowest:
joda-time was once an optional dependency in querydsl-sql, but isn't anymore, so it's safe to access the classes directly.

Shredder121:
(I also went for that explanation)

Then I am all for this change, less reflection is always better.

PR #1209, opened 2015-02-17
Support for Date Time API types in SQL module

arturchyzy:
I created JSR 310 (Date Time API) support in sql module. Still the problem is in sql module build. 
```
[INFO] --- animal-sniffer-maven-plugin:1.13:check (ensure-java-1.6-class-library) @ querydsl-sql ---
```
All new api classes will not be seen so I'm unable to build module on JDK 8. 
Please continue from here or let me know what to fix.


Shredder121:
@timowest there are actually two solutions to this.
- move this out, so we can make only that build on jdk8
- add a [JSR310-compatible library](http://www.threeten.org/) to the classpath, so the packages are known and we can compile against that.

Shredder121:
@arturchyzy thanks at least for the effort, I will fix the animal sniffer 'problem'.

timowest:
I'd maybe at first try the compatibility library approach, because there are only a few classes involved.

arturchyzy:
You think more of a separate library or new module ?

timowest:
I meant that the JSR 310 types would be included in querydsl-sql and the dependencies for the JDK 7 build would be provided by a compatibility library.

Shredder121:
ThreeTen is however published under a different package name, so unfortunately it doesn't work that well..

Shredder121:
What we could do, is via a maven profile:

with jdk8, we have another [plugin](http://maven.apache.org/plugins/maven-patch-plugin/) that applies a [patch](http://maven.apache.org/plugins/maven-patch-plugin/examples/single.html) to the imports of these classes.
The source references the backport's package structure, and for compatibility's sake we use that as reference what (if running jdk8) is available at runtime.
This would mean an additional querydsl-sql artifact with jdk8 classifier, so I don't know how you feel about this?

timowest:
I'd feel uncomfortable if we touch the sources. If we don't have a proper compatibility layer we probably need to have a separate module for the JSR 310 types. querydsl-sql-jsr310?

Shredder121:
Indeed.
I found one last thing, but it is rather.. weird.
Apparently, someone took ThreeTen-bp and [forked it](https://github.com/seratch/java-time-backport), to publish it under the java.time package.
This is not really a solution, since we can't conditionally check if the java.time.* classes come from the standard API or from that artifact.

In essence, I can't find a viable alternative, so a seperate module would be best.

arturchyzy:
I pushed some fixes and modification of data conversion (with set tests for added types). Please carefully take a look at them as I see those conversion are little bit tricky. I hope I did it well. Hope this issue will be resolved soon.

timowest:
Alternatively we can make JDK 1.8 a build requirement for Querydsl. Excluding querydsl-sql-jsr310 via a profile trick might be a bit messy.

Shredder121:
Yes, that is actually fine by me.
Since we strive and check for runtime compatibility, we shouldn't run into problems.

Shredder121:
I am reviewing this currently, and want to apply some fixes, is that okay?

For example [this method will never work](http://docs.oracle.com/javase/8/docs/api/java/sql/Time.html#toInstant--).

arturchyzy:
Just wanted to let you know that your changes has been merged. What's left to merge it to master ?

Shredder121:
We still need to update our build matrix.
If we were to merge it to master, none of our CI builds would compile, since it requires JDK8.
So @timowest shall we update travis.yml in this PR or a separate one?

timowest:
@Shredder121 It can be updated in this PR.

Shredder121:
I will do so shortly.

Shredder121:
I think it is a good idea to add a working test case, to test whether or not my DateTimeFormatter fix is any good.

Shredder121:
@timowest @arturchyzy I have added a few tests, and will push them.

There is however one problem with the JSR310ZonedDateTimeType.
There is no way to know what time zone was originally used, we can by default set it to UTC, but that is about it, and it is feels counterintuitive to the syntax.
```java
ZonedDateTime zonedDT = ...;
ZonedDateTime result = query().singleResult(Expressions.constant(zonedDT));
assertEquals(zonedDT, result);//<- fails, since TimeZone information could not be reconstructed.
```

Furthermore, it feels as if they made some mistake, `Instant`s feel practically useless.
I get it that they can't really be queried (day of week, etc), but all code relies on the querying model.
If you were to create a LocalDateTime out of it, the LocalDate part cannot be queried.

```java
Instant instant = ...;
LocalDateTime test = LocalDateTime.from(instant);// <- fails

```

After all of this I can say it is tricky to get working.

arturchyzy:
@Shredder121 If you want to save timezone I think you should use Types.TIMESTAMP_WIITH_TIMEZIME (available from Java 8) instead of Types.TIMESTAMP. But I feel it's not a common case to use it with timezone or limit use of ZoneDateTime to specific database usage. Joda-Time's DateTime also contains timezone and conversion was made to UTC so solution is compatible with Joda solution. 
About Instant - that's true. Instant is not a format to convert from, it's more like one to convert to. There are also  .of* methods which seems more like one which should be use when creating new instances. 

Please let me know if you need anything about this issue (writing tests, checking). Do you plan to push it to main repo ? Do you also plan to merge this request or work on your personal branch ? Anyway please let me know when I should close it.

Shredder121:
I (only) just pushed a few new commits, I think that writing tests would be very good indeed.
Also, I need people from different timezones, so it would be helpful to have a few people.
And I am not *quite* sure with what you meant with TIMESTAMP_WITH_TIMEZONE.

I think we can open up a pull request on our end? We can take the conversation there, this makes it a little easier to fix up some commits, and make the PR merge-able.
@timowest what is your opinion? do you agree?

timowest:
@Shredder121 I am fine with that.

arturchyzy:
@Shredder121 Check [TIMESTAMP_WITH_TIMEZONE](http://docs.oracle.com/javase/8/docs/api/java/sql/Types.html#TIMESTAMP_WITH_TIMEZONE) - introduced with java 8, previous versions of java had only [TIMESTAMP](http://docs.oracle.com/javase/8/docs/api/java/sql/Types.html#TIMESTAMP). Please be aware that there is also [TIME_WITH_TIMEZONE](http://docs.oracle.com/javase/8/docs/api/java/sql/Types.html#TIME_WITH_TIMEZONE) starting from Java 8

Shredder121:
I indeed found that already, but I meant that I don't *quite* see the relation between that type code and getting the timezone information from the resultset.
I think we'll have to make a test case with a column with timezone information and work from that.

timowest:
Closing in favor of https://github.com/querydsl/querydsl/issues/1399

Issue #1155, opened 2015-02-05
Can't trunc date to current day in Oracle


robertandrewbain:
In Oracle `TRUNC(some_date)` will remove the time element from the date. `TRUNC(some_date, 'day')` however, will return the start of the week, which is what you get if you call `SQLExpressions.datetrunc(DatePart.day, someDateTimePath)`. IMO, this is poor naming on Oracle's part, however I would have expected the SQL to manifest to be `TRUNC(some_date, 'dd')`, or for `DatePart` to include further values so that `TRUNC(some_date, 'dd')` was attainable, or better yet that `TRUNC(some_date)` with no time component was supported.

PR #1210, opened 2015-02-17
Fix date trunc mappings

timowest:
Related to #1155


Shredder121:
Are all commits backported? At least the comments are missing?

timowest:
The comments are missing, but I can add them.

timowest:
I compared the PRs and they seem to be equivalent.

Shredder121:
Then I'll merge it when the build finishes.

Issue #1204, opened 2015-02-16
Improve SelectBase DB coverage


timowest:
The coverage of the following tests should be enhanced

* SelectBase.Dates
* SelectBase.Date_Diff
* SelectBase.Date_Diff2
* SelectBase.DateTime_To_Date
* SelectBase.Order_NullsFirst
* SelectBase.Order_NullsLast
* SelectBase.YearWeek

Issue #1162, opened 2015-02-06
java.lang.NoClassDefFoundError: java/util/Objects - 3.6.1 with JDK 6


mmccreary:
The company I work for is still using JDK 6. After upgrading from Querydsl 3.6.0 to 3.6.1 I began receiving the following error while calling a JPQLQuery: 

java.lang.NoClassDefFoundError: java/util/Objects

It appears that java.util.Objects is new to JDK 7 and is not available in JDK 6

Shredder121:
Thank you for the bug report, and sorry for the inconvenience.
I am preparing a fix, but could you maybe post your stack trace here?

mmccreary:
Thank you for the quick response. Here is the full stack trace:

java.lang.NoClassDefFoundError: java/util/Objects
	at com.mysema.query.support.ReplaceVisitor.visit(ReplaceVisitor.java:75)
	at com.mysema.query.jpa.JPAMapAccessVisitor.visit(JPAMapAccessVisitor.java:25)
	at com.mysema.query.jpa.JPAMapAccessVisitor.visit(JPAMapAccessVisitor.java:14)
	at com.mysema.query.types.PathImpl.accept(PathImpl.java:94)
	at com.mysema.query.support.ReplaceVisitor.visit(ReplaceVisitor.java:162)
	at com.mysema.query.support.ReplaceVisitor.visit(ReplaceVisitor.java:49)
	at com.mysema.query.support.ReplaceVisitor.visit(ReplaceVisitor.java:30)
	at com.mysema.query.types.OperationImpl.accept(OperationImpl.java:90)
	at com.mysema.query.support.ReplaceVisitor.visit(ReplaceVisitor.java:162)
	at com.mysema.query.support.ReplaceVisitor.visit(ReplaceVisitor.java:49)
	at com.mysema.query.support.ReplaceVisitor.visit(ReplaceVisitor.java:30)
	at com.mysema.query.types.OperationImpl.accept(OperationImpl.java:90)
	at com.mysema.query.support.ReplaceVisitor.visit(ReplaceVisitor.java:162)
	at com.mysema.query.support.ReplaceVisitor.visit(ReplaceVisitor.java:49)
	at com.mysema.query.support.ReplaceVisitor.visit(ReplaceVisitor.java:30)
	at com.mysema.query.types.OperationImpl.accept(OperationImpl.java:90)
	at com.mysema.query.jpa.JPAQueryMixin.normalize(JPAQueryMixin.java:196)
[...]

PR #1211, opened 2015-02-18
Improve SelectBase coverage

timowest:
Fixes https://github.com/querydsl/querydsl/issues/1204

Databases to go through

* [x] CUBRID 9.2.0
* [x] DB2 10.1.2
* [x] Derby 10.11.1.1
* [x] Firebird 2.5.2
* [x] H2 1.4.178
* [x] HSQLDB 2.3.2
* [x] MSSQL 2008
* [x] MySQL 5.5
* [x] Oracle 11.2.0
* [x] PostgreSQL 9.3
* [X] SQLite
* [x] Teradata 14.0.3

DB2, MSSQL, Oracle and Teradata tests are not run in travis

backport https://github.com/querydsl/querydsl/pull/1269

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -29,7 +29,7 @@
     <!-- deps -->
     <derby.version>10.11.1.1</derby.version>
     <hsqldb.version>2.3.2</hsqldb.version>
-    <h2.version>1.4.178</h2.version>
+    <h2.version>1.4.186</h2.version>
     <postgresql.version>9.4-1200-jdbc41</postgresql.version>
     <oracle.version>11.1.0.7.0</oracle.version>
     <mysql.version>5.1.30</mysql.version>
--- a/pom.xml
+++ b/pom.xml
@@ -27,7 +27,7 @@
     <project.checkout>scm:git:git@github.com:querydsl/querydsl.git</project.checkout>
     
     <!-- deps -->
-    <derby.version>10.10.1.1</derby.version>
+    <derby.version>10.11.1.1</derby.version>
     <hsqldb.version>2.3.1</hsqldb.version>
     <h2.version>1.4.178</h2.version>
     <postgresql.version>9.3-1101-jdbc41</postgresql.version>
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
     
     <!-- deps -->
     <derby.version>10.11.1.1</derby.version>
-    <hsqldb.version>2.3.1</hsqldb.version>
+    <hsqldb.version>2.3.2</hsqldb.version>
     <h2.version>1.4.178</h2.version>
     <postgresql.version>9.3-1101-jdbc41</postgresql.version>
     <oracle.version>11.1.0.7.0</oracle.version>
--- a/pom.xml
+++ b/pom.xml
@@ -34,7 +34,7 @@
     <oracle.version>11.1.0.7.0</oracle.version>
     <mysql.version>5.1.30</mysql.version>
     <jtds.version>1.3.1</jtds.version>
-    <cubrid.version>8.4.0</cubrid.version>
+    <cubrid.version>9.3.1.0005</cubrid.version>
     <sqlite.version>3.7.2</sqlite.version>
     <teradata.version>13.10.00.35</teradata.version>
     <firebird.version>2.2.5</firebird.version>
--- a/pom.xml
+++ b/pom.xml
@@ -30,7 +30,7 @@
     <derby.version>10.11.1.1</derby.version>
     <hsqldb.version>2.3.2</hsqldb.version>
     <h2.version>1.4.178</h2.version>
-    <postgresql.version>9.3-1101-jdbc41</postgresql.version>
+    <postgresql.version>9.4-1200-jdbc41</postgresql.version>
     <oracle.version>11.1.0.7.0</oracle.version>
     <mysql.version>5.1.30</mysql.version>
     <jtds.version>1.3.1</jtds.version>
--- a/querydsl-sql-spatial/src/main/java/com/querydsl/sql/spatial/SQLServerGeometryType.java
+++ b/querydsl-sql-spatial/src/main/java/com/querydsl/sql/spatial/SQLServerGeometryType.java
@@ -13,16 +13,18 @@
  */
 package com.querydsl.sql.spatial;
 
-import javax.annotation.Nullable;
 import java.io.IOException;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Types;
 
-import com.querydsl.sql.types.AbstractType;
+import javax.annotation.Nullable;
+
 import org.geolatte.geom.Geometry;
 import org.geolatte.geom.codec.Wkt;
+
+import com.querydsl.sql.types.AbstractType;
 
 /**
  * @author tiwe
@@ -72,7 +74,7 @@ public class SQLServerGeometryType extends AbstractType<Geometry> {
         if (geometry.getSRID() > -1) {
             return "geometry::STGeomFromText('" + str + "', " + geometry.getSRID() + ")";
         } else {
-            return "geometry::STGeomFromText('" + str + "')";
+            return "geometry::STGeomFromText('" + str + "', 4326)";
         }
     }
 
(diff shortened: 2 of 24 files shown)


timowest:
Databases to go through

* [ ] CUBRID
* [ ] DB2
* [ ] Derby
* [ ] Firebird
* [ ] H2
* [ ] HSQLDB
* [ ] MSSQL
* [ ] MySQL
* [ ] Oracle
* [ ] PostgreSQL
* [ ] SQLite
* [ ] Teradata

timowest:
I rebased this now.

timowest:
@johnktims @Shredder121 This is ready to review now.

timowest:
@johnktims @Shredder121 How does this look?

pom.xml
```diff
@@ -27,14 +27,14 @@
     <project.checkout>scm:git:git@github.com:querydsl/querydsl.git</project.checkout>
     
     <!-- deps -->
-    <derby.version>10.10.1.1</derby.version>
-    <hsqldb.version>2.3.1</hsqldb.version>
+    <derby.version>10.11.1.1</derby.version>
+    <hsqldb.version>2.3.2</hsqldb.version>
     <h2.version>1.4.178</h2.version>
```
johnktims: 1.4.186 is the latest version for H2.

There are some other dependencies that could be updated. Separate PR, perhaps?

querydsl-sql-spatial/src/main/java/com/querydsl/sql/spatial/SQLServerGeometryType.java
```diff
@@ -72,7 +74,7 @@ public String getLiteral(Geometry geometry) {
         if (geometry.getSRID() > -1) {
             return "geometry::STGeomFromText('" + str + "', " + geometry.getSRID() + ")";
         } else {
-            return "geometry::STGeomFromText('" + str + "')";
+            return "geometry::STGeomFromText('" + str + "', 4326)";
```
johnktims: What is "4326"?

Shredder121: SQLServer's default `geometry.getSRID()`?

pom.xml
timowest: I can update it here as well, but for others a separate PR could be created.

querydsl-sql-spatial/src/main/java/com/querydsl/sql/spatial/SQLServerGeometryType.java
timowest: The SRID 4326 stands for WGS 84 http://en.wikipedia.org/wiki/World_Geodetic_System#A_new_World_Geodetic_System:_WGS_84

It is used as the default SRID elsewhere https://msdn.microsoft.com/en-us/library/bb933824.aspx

querydsl-sql/src/test/java/com/querydsl/sql/types/TypeTest.java
```diff
@@ -45,7 +45,29 @@
 
     @Override
     public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
-        if (method.getName().startsWith("get")) {
+        if (method.getName().equals("wasNull")) {
```
johnktims: It doesn't matter here since it's just a test, but I like to call `equals` on the constant string (`"wasNull".equals(method.getName());`) to eliminate any chance of an NPE.

querydsl-sql-spatial/src/main/java/com/querydsl/sql/spatial/SQLServerGeometryType.java
johnktims: I would probably name that magic constant, but I think the rest of the PR looks good.

querydsl-sql/src/test/java/com/querydsl/sql/types/TypeTest.java
timowest: How does that protect of an NPE in this case? `method.getName()` is guaranteed to be non null, and the reverse order doesn't handle the `method == null` case.

Shredder121: I think he means it more as a [general rule](http://checkstyle.sourceforge.net/config_coding.html#EqualsAvoidNull).
Since there is a checkstyle check, we could incorporate this.

querydsl-sql/src/test/java/com/querydsl/sql/SelectBase.java
```diff
@@ -395,7 +395,7 @@ public void Custom_Projection() {
     }
 
     @Test
-    @IncludeIn({H2, SQLSERVER, MYSQL, ORACLE, TERADATA}) // TODO fix postgresql
+    @ExcludeIn({CUBRID, DB2, DERBY, HSQLDB, POSTGRESQL, SQLITE, TERADATA})
```
Shredder121: I suggest that we also apply this on the other branch, so we don't pull our hairs out every time we change this line.
We could also make it a separate PR to equalize this everywhere.

querydsl-sql/src/test/java/com/querydsl/sql/types/TypeTest.java
timowest: I'd consider the inverted order only for known nullable types and not as a general practice, since having the non-null visible in the access is also valuable and at least for me this order is more readable.

querydsl-sql/src/test/java/com/querydsl/sql/SelectBase.java
timowest: what do you mean? The ExcludeIn usage for this method is the same on the other branch.

Shredder121: Also in that order? If that is the case, then it's okay.

timowest: it's POSTGRES in querydsl3 and POSTGRESQL in querydsl4, otherwise the same

querydsl-sql/src/test/java/com/querydsl/sql/types/TypeTest.java
Shredder121: Personally I find the order as it is now much more readable.
If I want to do a null-safe equals, I always use Guava's `Objects.equal` instead.

timowest: good point with `Objects.equals`, which solves the more general case

Shredder121: *__Not__ `Objects.equals`, that is `java.util.Objects` and is [not available in java 6](https://github.com/querydsl/querydsl/issues/1162)*

timowest: fixed

querydsl-sql/src/test/java/com/querydsl/sql/SelectBase.java
Shredder121: Still impossible to cherry pick, but it's okay.
Maybe it's a good idea to always order it alphabetically? (I don't know if there is a checkstyle check for that?)

querydsl-sql/src/test/java/com/querydsl/sql/types/TypeTest.java
johnktims: I wasn't aware of `Objects.equal`. I should have known guava would have a helper method for this!

Shredder121: Yes, they have helper methods for a lot of things, it's knowing (and using) them that makes Guava so powerful.
However, the initial discoveries can be quite slow, since a lot of helper methods are very domain specific (`io` package is pretty common, but I don't use `net` for example)

querydsl-sql/src/test/java/com/querydsl/sql/SelectBase.java
timowest: Yes, we can maybe homogenize that later.

timowest:
@johnktims @Shredder121 Is this good to merge, once the tests pass?

johnktims:
:+1:

Shredder121:
Yes, for me it looks good as well. :+1: 

Just a question, is there a reason we explicitly box primitives?

timowest:
Not a good reason for boxing, but that could be maybe removed in another PR?

Shredder121:
I didn't really mean so we could 'fix' that,  more if there was something behind it.

PR #1203, opened 2015-02-15
Transform containsKey and containsValue via JPAMapAccessVisitor

timowest:
Fixes #1198

PR #1215, opened 2015-02-19
Fix typo in ParameterizedExpression interface

johnktims:


timowest:
@johnktims Thanks for the PR!

I am not completely sure if the rename of the `ParametrizedExpression` interface is necessary, since there are multiple variants of the parametrize verb and parametrize with parametrized as the simple past seems to be one of them http://en.wiktionary.org/wiki/parameterize, although probably not the dominant one.

@Shredder121 What do you think?

johnktims:
@timowest Google and m-w.com redirect to "parameterize". The wikipedia page you linked says "Dominant modern spelling in US" at the bottom so maybe that's why "parametrize" looks wrong to me. :)

johnktims:
For what it's worth: https://books.google.com/ngrams/graph?content=parameterize%2Cparametrize

Shredder121:
I don't really have a biased opinion on this, but thanks for taking note @johnktims.
The only opinion I do have is that it should be used consistently.

So if we decide upon changing it, it should also be reflected on the [ new class in #1203](https://github.com/querydsl/querydsl/pull/1203/files#diff-69bb543da66867f237b0028ab558bd43).

timowest:
Ok, I am then fine with the change and I agree that the same style should also be applied elsewhere. We have a review time of 2 days, but after that this PR can be merged.

@johnktims Thanks for the contribution!

PR #1217, opened 2015-02-23
Separate projection from query execution

### LINKED ISSUE #1120: Separate projection from query execution
e.g.

```java
query.from(customer).where(customer.firstName.eq("Tom")).list(customer)
```

into

```java
query.select(customer).from(customer).where(customer.firstName.eq("Tom")).list()
```

This has the following benefits

* queries can be written both with select definition in the beginning as well as the end
* subqueries can be defined via normal Query instances
* Projectable and SimpleProjectable can be merged into the same class

Disadvantages
* more verbose syntax in some cases
* generic parameter in Query interface/class

timowest:
Fixes #1120

doesn't have a backport, since it is a backwards incompatible change

Does not yet compile and also SQL part has not yet been refactored, major SQL related PRs should be merged before this is continued.


timowest:
This PR contains now the first steps of the refactoring to the new query model. Although some usages become now a bit more verbose, the new model feels much more flexible.

timowest:
I am not yet quite happy with ProjectableQuery, maybe Selectable or something might be better?

timowest:
@Shredder121 @johnktims  Finally building now. Please review ;)

Some other changes done 
* rename from Projectable to Fetchable
* QueryMetadata validation off by default
* experimental Scala query types removed

timowest:
Let's keep this in progress until we finalize the method names for `Fetchable`

querydsl-collections/src/main/java/com/querydsl/collections/CollDeleteClause.java
```diff
@@ -33,10 +33,10 @@
 
     private final Path<T> expr;
 
-    private final CollQuery query;
+    private final CollQuery<T> query;
 
     public CollDeleteClause(QueryEngine qe, Path<T> expr, Collection<? extends T> col) {
-        this.query = new CollQuery(qe).from(expr, col);
+        this.query = new CollQuery<Void>(qe).from(expr, col).select(expr);
```
Shredder121: Maybe in favor of `new XXXQuery<Void>` we can add static factory methods that return a `XXXQuery<?>`.
Essentially to hide the not yet known parameter.
I recall having discussed this, but I think it's good to recap.

querydsl-collections/src/main/java/com/querydsl/collections/CollQuery.java
```diff
@@ -74,16 +73,19 @@ public CollQuery(QueryMetadata metadata, QueryEngine queryEngine) {
...
 
-    /**
-     * @return
-     */
     @Override
-    public QueryMetadata getMetadata() {
-        return queryMixin.getMetadata();
+    public <E> CollQuery<E> select(Expression<E> expr) {
+        queryMixin.setProjection(expr);
+        return (CollQuery<E>)this;
```
Shredder121: @timowest have you by chance experimented with letting the querymixin hold the result type? Is that an option?

querydsl-collections/src/main/java/com/querydsl/collections/CollDeleteClause.java
timowest: Good point. I will do that.

querydsl-collections/src/main/java/com/querydsl/collections/CollQuery.java
timowest: I am not sure. QueryMixin has the query type as an argument, but the select methods change it, so I guess there will be anyway a cast.

timowest:
Concerning finishing the javadocs, I'd like to continue that on a per module basis, that's why there are some empty stubs.
So in order to continue with the javadoc fixes, it would be good to get this merged.

Shredder121:
I see, I do agree with that.

PR #1220, opened 2015-02-24
Improve operator precedence mappings

### LINKED ISSUE #1219: Wrong precedence rules for SQL serialization
I have an issue with QueryDSL SQL >= 3.5.0 and the placement of parenthesis in the generated SQL in order by statements. The DB is H2 v1.3.176

Given the following statement:

    .orderBy(MY_TABLE.X.multiply(a).divide(MY_TABLE.Y.multiply(b)).asc())

with a = 100 and b = 200 for example

I expect the following SQL:

    order by MY_TABLE.X * 100 / (MY_TABLE.Y * 200) asc

but instead I get:

    order by MY_TABLE.X * 100 / MY_TABLE.Y * 200 asc

which does not lead to the same result.

timowest:
Fixes #1219

### DIFF
--- a/querydsl-core/src/main/java/com/querydsl/core/support/Normalization.java
+++ b/querydsl-core/src/main/java/com/querydsl/core/support/Normalization.java
@@ -26,7 +26,7 @@ public final class Normalization {
 
     // TODO simplify
     private static final Pattern FULL_OPERATION = Pattern.compile(
-            "(?<![\\d\\*/\"' ])" + "(\\b|\\(|\\s+)"  +
+            "(?<![\\d\\*/\"\\?' ])" + "(\\b|\\(|\\s+)"  +
             "(" + NUMBER + WS + "[+\\-/*]" + WS + ")+" + NUMBER + WS +
             "(?![\\d\\*/\"' ])");
 
--- a/querydsl-core/src/main/java/com/querydsl/core/support/Normalization.java
+++ b/querydsl-core/src/main/java/com/querydsl/core/support/Normalization.java
@@ -31,10 +31,8 @@ public final class Normalization {
             "(?![\\d\\*/\"' ])");
 
     private static final Pattern[] OPERATIONS = {
-            Pattern.compile(NUMBER + WS + "\\*" + WS + NUMBER),
-            Pattern.compile(NUMBER + WS + "/" + WS + NUMBER),
-            Pattern.compile(NUMBER + WS + "\\+" + WS + NUMBER),
-            Pattern.compile(NUMBER + WS + "\\-" + WS + NUMBER)
+            Pattern.compile(NUMBER + WS + "(\\*|/)" + WS + NUMBER),
+            Pattern.compile(NUMBER + WS + "(\\+|\\-)" + WS + NUMBER),
     };
 
     private static String normalizeOperation(String queryString) {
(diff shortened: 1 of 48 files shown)


timowest:
Databases to go through

* [x] CUBRID 9.2.0
* [x] DB2 10.1.2
* [x] Derby 10.11.1.1
* [x] Firebird 2.5.2
* [x] H2 1.4.178
* [x] HSQLDB 2.3.2
* [x] MSSQL 2008
* [x] MySQL 5.5
* [x] Oracle 11.2.0
* [x] PostgreSQL 9.3
* [ ] SQLite
* [ ] Teradata 14.0.3

timowest:
@Shredder121 This is close to the finish line. Could you take a look?

I will make a separate PR to tune the handling of the template strings.

timowest:
@Shredder121 Could you take a look at this? After this has been merged, I will also finish also https://github.com/querydsl/querydsl/pull/1211

Shredder121:
Yes, I will.

timowest:
Shall we merge this first https://github.com/querydsl/querydsl/pull/1236

Shredder121:
Yes, you [said that already](https://github.com/querydsl/querydsl/pull/1236#issuecomment-77589064).
I am looking at it first.

timowest:
I rebased this now.

querydsl-core/src/main/java/com/querydsl/core/support/Normalization.java
```diff
@@ -26,15 +26,13 @@
...
             "(" + NUMBER + WS + "[+\\-/*]" + WS + ")+" + NUMBER + WS +
             "(?![\\d\\*/\"' ])");
 
     private static final Pattern[] OPERATIONS = {
-            Pattern.compile(NUMBER + WS + "\\*" + WS + NUMBER),
-            Pattern.compile(NUMBER + WS + "/" + WS + NUMBER),
-            Pattern.compile(NUMBER + WS + "\\+" + WS + NUMBER),
-            Pattern.compile(NUMBER + WS + "\\-" + WS + NUMBER)
+            Pattern.compile(NUMBER + WS + "(\\*|/)" + WS + NUMBER),
+            Pattern.compile(NUMBER + WS + "(\\+|\\-)" + WS + NUMBER),
```
Shredder121: Can you consider using character classes?
There are only 2 options, so it can easily fit in a `java.util.regex.Pattern$BitClass` [which is available from at least 1.6](http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b27/java/util/regex/Pattern.java#Pattern.BitClass), so the performance should be better than a [branch](http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b27/java/util/regex/Pattern.java#4126)('|').

See [match](http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b27/java/util/regex/Pattern.java#Pattern.BmpCharProperty.match%28java.util.regex.Matcher%2Cint%2Cjava.lang.CharSequence%29) and BitClass' [implementation of `isSatisfiedBy()`](http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b27/java/util/regex/Pattern.java#Pattern.BitClass.isSatisfiedBy%28int%29), it then uses a simple array lookup instead of a foreach.

Also, in a character class, meta characters lose their meaning, so escaping is not necessary. (- is not a metacharacter)

timowest: - is a metachar in char groups, at least in special positions

Shredder121: >in special positions

Yes, you are very right, but what I meant was that escaping isn't necessary in this case, since the '-' is not in a special position.
+1 for you.

timowest:
@Shredder121 Is this good to merge now?

Shredder121:
I believe it is now yes.

Issue #1223, opened 2015-02-25
com.querydsl.codegen.BeanSerializer.addToString doesn't generate @Override annotation on method


altarsofmadness:
com.querydsl.codegen.BeanSerializer.addToString doesn't generate @Override annotation on method which leads to compiler warnings.

timowest:
Thanks for the issue! I added two related PRs.

PR #1226, opened 2015-02-25
Add Override annotation

timowest:
Related to #1223


timowest:
I guess this can be now merged as well.

Issue #1112, opened 2015-01-09
Adjustment suggestions to GroupBy generics


TuomasKiviaho:
Hi,

I noticed a slight discomfort in using generics while grouping.

``GSum<T extends Number & Comparable<T>>``  could be just ``GSum<T extends Number>``
``GAvg<T extends Number & Comparable<T>>``  could be just ``GAvg<T extends Number>``
``GMin<T extends Comparable<T>>`` should be``GMin<T extends Comparable<? super T>>``
``GMax<T extends Comparable<T>>`` should be ``GMax<T extends Comparable<? super T>>``

The latter ones would actually be useful for me because I have to dumb down the generics in order to get things working (or typecast the date as ``java.util.Date``)
``GroupExpression<java.sql.Date, java.sql.Date> min = GroupBy.min((DatePath) date)``

timowest:
@TuomasKiviaho We are currently preparing to switch to a different package structure for the 4.0 release so it might be easier to do this change after that.

PR #1232, opened 2015-03-03
Add GroupBy.sortedSet and sortedMap

timowest:
Backports for #1116 and https://github.com/querydsl/querydsl/issues/1112

### DIFF
--- a/querydsl-core/src/main/java/com/mysema/query/group/GroupBy.java
+++ b/querydsl-core/src/main/java/com/mysema/query/group/GroupBy.java
@@ -13,9 +13,7 @@
  */
 package com.mysema.query.group;
 
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
 
 import com.infradna.tool.bridge_method_injector.WithBridgeMethods;
 import com.mysema.commons.lang.Pair;
@@ -57,7 +55,7 @@ public final class GroupBy {
      * @param expression
      * @return
      */
-    public static <E extends Comparable<E>> AbstractGroupExpression<E, E> min(Expression<E> expression) {
+    public static <E extends Comparable<? super E>> AbstractGroupExpression<E, E> min(Expression<E> expression) {
         return new GMin<E>(expression);
     }
 
@@ -67,7 +65,7 @@ public final class GroupBy {
      * @param expression
      * @return
      */
-    public static <E extends Number & Comparable<E>> AbstractGroupExpression<E, E> sum(Expression<E> expression) {
+    public static <E extends Number> AbstractGroupExpression<E, E> sum(Expression<E> expression) {
         return new GSum<E>(expression);
     }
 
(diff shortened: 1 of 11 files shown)


querydsl-core/src/main/java/com/mysema/query/group/GroupBy.java
```diff
@@ -106,43 +103,200 @@
...
+
+
+    /**
+     * Create a new aggregating map expression using a backing LinkedHashMap
      *
      * @param key
      * @param value
      * @return
      */
-    @WithBridgeMethods(value=Expression.class,castRequired=true)
```
Shredder121: Are you sure the bridge method is not necessary?

timowest: I will fix that

timowest:
@Shredder121 Could you take a look at this? Is it good to merge?

Shredder121:
Just wondering, did you actually look at the signatures?
You never answered my question.

timowest:
Sorry for forgetting to answer. No, I didn't look. I think you posted a signature comparison already some time ago and the change seemed to be backwards compatible.

Shredder121:
No I haven't posted a signature comparison, only a way of obtaining one.
I just [made one for this changeset](https://gist.github.com/Shredder121/a7882cda241a84ff0936), to the best of my ability (windows and all).

Anyway, most notable are [this one](https://gist.github.com/Shredder121/a7882cda241a84ff0936#file-output-diff-L61) and [this one](https://gist.github.com/Shredder121/a7882cda241a84ff0936#file-output-diff-L262), where the erased runtime method name stay the same.

One notable difference that I picked up: The numbering of inner classes is off. I am looking into it now.
If the existing inner classes come first, the numbering is fine again. (otherwise i.e. GMap$1 is not the GMap$1 that you think it is, and it is actually moved to GMap$4.

Look for example at [this](https://gist.github.com/Shredder121/a7882cda241a84ff0936#file-output-diff-L219), where the GroupCollector is moved to [GSet$4](https://gist.github.com/Shredder121/a7882cda241a84ff0936#file-output-diff-L252)

Shredder121:
GMap is now fixed, looking into GSet.

Shredder121:
I fixed it now, [see revisions](https://gist.github.com/Shredder121/a7882cda241a84ff0936/revisions).

Note that GSet is harder to read, since [abstract is added](https://gist.github.com/Shredder121/a7882cda241a84ff0936#file-output-diff-L275), but all new classes come [after GSet$1](https://gist.github.com/Shredder121/a7882cda241a84ff0936#file-output-diff-L243).

Shredder121:
It probably didn't matter, but the 'diff'(of the signatures) is much cleaner now in my opinion.

Shredder121:
Actually, I think in a serialisation sense it did matter.

timowest:
This looks good now in my opinion.

Shredder121:
I also think it now looks good.

Shredder121:
I don't quite know how to improve serialisability or compatibility in serialising of anonymous inner classes, but in a members sense, I think nothing has changed in the GroupCollectors.

timowest:
We also don't enforce backwards compatibility of the serialized form in general, so this is fine with me.

Shredder121:
Then I'll merge this.

PR #1238, opened 2015-03-05
Tune surefire settings

timowest:


Shredder121:
Once we build with JDK8, there is no permgen space, so that flag would do nothing.
Or do you only want to apply this in the meantime?

timowest:
I can remove it directly.

timowest:

Shredder121:
Yes, this is good to merge. I will do so.

PR #1239, opened 2015-03-05
Simplify OperatorImpl serialization

### LINKED ISSUE #1237: Concurrency Issue Loading Ops and OperatorImpl
Whenever `com.mysema.query.types.OperatorImpl` and `com.mysema.query.types.Ops` are loaded simultaneously loaded a deadlock is reached.

The following code could be used to reproduce the issue:

```java
import org.junit.Test;

import com.mysema.query.jpa.impl.JPAQuery;

public class CycleClassInitDependencyTest {

	@Test(timeout = 2000)
	public void test() throws InterruptedException {

		// If class is loaded before hand it will work for example:
		// System.out.println(Ops.DateTimeOps.DATE);

		// each thread wants
		Thread t1 = new Thread(new LoadClassRunnable("com.mysema.query.types.OperatorImpl"));
		Thread t2 = new Thread(new LoadClassRunnable("com.mysema.query.types.Ops"));
		t1.start();
		t2.start();

		// trying to create anything depending on Ops class for instance:
		// JPAQuery -> JPQLTemplates -> Ops
		new JPAQuery();

	}

	public class LoadClassRunnable implements Runnable {

		private String classToLoad;

		public LoadClassRunnable(String classToLoad) {
			this.classToLoad = classToLoad;
		}

		public void run() {
			try {
				Class.forName(classToLoad);
			} catch (ClassNotFoundException e) {
				e.printStackTrace();
			}
		}

	}

}
```

A jstack relevant part of the execution of code exposed above is:
```
"OpslLoading        " prio=10 tid=0x00007f47a4003000 nid=0x2dd3 in Object.wait() [0x00007f47b7cfb000]
   java.lang.Thread.State: RUNNABLE
	at com.mysema.query.types.Ops.<clinit>(Ops.java:30)
	at java.lang.Class.forName0(Native Method)
	at java.lang.Class.forName(Class.java:188)
	at CycleClassInitDependencyTest$LoadClassRunnable.run(CycleClassInitDependencyTest.java:37)
	at java.lang.Thread.run(Thread.java:722)

"OperatorImplLoading" prio=10 tid=0x00007f47a4001800 nid=0x2dd2 in Object.wait() [0x00007f47b7dfb000]
   java.lang.Thread.State: RUNNABLE
	at sun.misc.Unsafe.ensureClassInitialized(Native Method)
	at sun.reflect.UnsafeFieldAccessorFactory.newFieldAccessor(UnsafeFieldAccessorFactory.java:43)
	at sun.reflect.ReflectionFactory.newFieldAccessor(ReflectionFactory.java:140)
	at java.lang.reflect.Field.acquireFieldAccessor(Field.java:949)
	at java.lang.reflect.Field.getFieldAccessor(Field.java:930)
	at java.lang.reflect.Field.get(Field.java:372)
	at com.mysema.query.types.OperatorImpl.<clinit>(OperatorImpl.java:42)
	at java.lang.Class.forName0(Native Method)
	at java.lang.Class.forName(Class.java:188)
	at CycleClassInitDependencyTest$LoadClassRunnable.run(CycleClassInitDependencyTest.java:37)
	at java.lang.Thread.run(Thread.java:722)

"Thread-0" prio=10 tid=0x00007f47ec205000 nid=0x2dd1 in Object.wait() [0x00007f47b7efc000]
   java.lang.Thread.State: RUNNABLE
	at com.mysema.query.types.Templates.<init>(Templates.java:50)
	at com.mysema.query.types.Templates.<init>(Templates.java:38)
	at com.mysema.query.types.Templates.<clinit>(Templates.java:27)
	at com.mysema.query.jpa.impl.JPAQuery.<init>(JPAQuery.java:35)
	at CycleClassInitDependencyTest.test(CycleClassInitDependencyTest.java:23)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:601)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:47)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:44)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.FailOnTimeout$StatementThread.run(FailOnTimeout.java:74)
```

timowest:
Fixes #1237

### DIFF
--- a/querydsl-core/src/main/java/com/mysema/query/types/OperatorImpl.java
+++ b/querydsl-core/src/main/java/com/mysema/query/types/OperatorImpl.java
@@ -13,12 +13,8 @@
  */
 package com.mysema.query.types;
 
+import java.io.ObjectStreamException;
 import java.lang.reflect.Field;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
 
 import javax.annotation.concurrent.Immutable;
 
@@ -27,24 +23,6 @@ import javax.annotation.concurrent.Immutable;
  */
 @Immutable
 public final class OperatorImpl<T> implements Operator<T> {
-
-    static final Map<String, Operator<?>> OPS = new HashMap<String, Operator<?>>(150);
-
-    static {
-        try {
-            // initialize all fields of Ops
-            List<Field> fields = new ArrayList<Field>();
-            fields.addAll(Arrays.asList(Ops.class.getFields()));
-            for (Class<?> cl : Ops.class.getClasses()) {
-                fields.addAll(Arrays.asList(cl.getFields()));
-            }
-            for (Field field : fields) {
-                field.get(null);
-            }
-        } catch (Exception e) {
-            throw new RuntimeException(e.getMessage(), e);
-        }
-    }
 
     private static final long serialVersionUID = -2435035383548549877L;
 
@@ -59,7 +37,6 @@ public final class OperatorImpl<T> implements Operator<T> {
     private OperatorImpl(String id) {
         this.id = id;
         this.hashCode = id.hashCode();
-        OPS.put(id, this);
     }
 
     /**
@@ -80,5 +57,19 @@ public final class OperatorImpl<T> implements Operator<T> {
     @Override
     public int hashCode() {
         return hashCode;
+    }
+
+    private Object readResolve() throws ObjectStreamException {
+        try {
+            String[] names = id.split("#");
+            Field opField = Class.forName(names[0]).getDeclaredField(names[1]);
+            return opField.get(null);
+        } catch (NoSuchFieldException e) {
+            throw new RuntimeException(e);
+        } catch (ClassNotFoundException e) {
+            throw new RuntimeException(e);
+        } catch (IllegalAccessException e) {
+            throw new RuntimeException(e);
+        }
     }
 }
--- a/querydsl-core/src/main/java/com/mysema/query/types/OperatorImpl.java
+++ b/querydsl-core/src/main/java/com/mysema/query/types/OperatorImpl.java
@@ -62,7 +62,7 @@ public final class OperatorImpl<T> implements Operator<T> {
     private Object readResolve() throws ObjectStreamException {
         try {
             String[] names = id.split("#");
-            Field opField = Class.forName(names[0]).getDeclaredField(names[1]);
+            Field opField = Class.forName(names[0]).getField(names[1]);
             return opField.get(null);
         } catch (NoSuchFieldException e) {
             throw new RuntimeException(e);
--- a/querydsl-jpa/src/test/java/com/mysema/query/jpa/CycleClassInitDependencyTest.java
+++ b/querydsl-jpa/src/test/java/com/mysema/query/jpa/CycleClassInitDependencyTest.java
@@ -1,0 +1,45 @@
+package com.mysema.query.jpa;
+
+import org.junit.Test;
+
+import com.mysema.query.jpa.impl.JPAQuery;
+
+public class CycleClassInitDependencyTest {
+
+    @Test(timeout = 2000)
+    public void test() throws InterruptedException {
+
+        // If class is loaded before hand it will work for example:
+        // System.out.println(Ops.DateTimeOps.DATE);
+
+        // each thread wants
+        Thread t1 = new Thread(new LoadClassRunnable("com.mysema.query.types.OperatorImpl"));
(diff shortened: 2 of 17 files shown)


querydsl-jpa/src/test/java/com/mysema/query/jpa/CycleClassInitDependencyTest.java
```diff
@@ -0,0 +1,45 @@
...
+
+        // each thread wants
+        Thread t1 = new Thread(new LoadClassRunnable("com.mysema.query.types.OperatorImpl"));
+        Thread t2 = new Thread(new LoadClassRunnable("com.mysema.query.types.Ops"));
+        t1.start();
+        t2.start();
+
+        // trying to create anything depending on Ops class for instance:
+        // JPAQuery -> JPQLTemplates -> Ops
+        new JPAQuery();
```
Shredder121: I will move this to core, since it is also well reproducible there and more local to the source being tested, okay?
Also, you and your "no newline at end of file" ;)

Shredder121:
I also think it is a good idea to make a unit test that validates the naming convention of Ops fields.
This way the error detection is done using that test, and not at runtime, or (de)serialization tests.

timowest:
Which surefire improvements do you mean?

Shredder121:
I meant #1238

timowest:
I thought I'd make a separate PR for that, because the targets are also different.

Shredder121:
Also, the targets are indeed different.
Then it's okay and a separate pull request is in order.

querydsl-core/src/main/java/com/mysema/query/types/OperatorImpl.java
```diff
@@ -81,4 +58,18 @@ public String toString() {
     public int hashCode() {
         return hashCode;
     }
+
+    private Object readResolve() throws ObjectStreamException {
+        try {
+            String[] names = id.split("#");
+            Field opField = Class.forName(names[0]).getDeclaredField(names[1]);
```
Shredder121: Consider using `getField()` instead, it's implementation in OpenJDK 8 seems more performant.
Also the semantics of that method apply better in this case, since it is a public field.

Shredder121: *EDIT: I think there is no difference, it is your call.*

timowest: Maybe `getField` because of the semantics.

Shredder121:
Just a small note.
This can not be used as a regression test, as I am fairly sure that whenever surefire reaches this test class, Ops and OperatorImpl might already be loaded.
But for this case, we fixed it.

timowest:
Shall we add a test that uses a fresh classloader?

Shredder121:
Yes, I think that should do it.
Could you also put the erroneous loading back to see if it works?

timowest:
Yes, I will test it also with the old logic.

timowest:
Added the fresh classloader usage and verified that it works without the fix.

timowest:
I was sloppy in my testing, the fresh class loader requirement might need a customer runner.

Shredder121:
Why? didn't it work?
A Thread seems to inherit the parent's contextClassLoader.

timowest:
Overriding the contextClassLoader worked, but it didn't affect the classloading of the test class.

Shredder121:
I fixed it now. Class.forName can take a ClassLoader argument.
Furthermore, I found a better way to create the URLClassLoader, so I'll commit that.
Also, when creating **fresh** ClassLoaders, it must not have a parent ClassLoader.

Even though the *test* is not the same as was reported, it tests the same, I have verified that it now works as expected.

timowest:
@Shredder121 Shall we merge this now?

Shredder121:
Yes, I think this now works as expected. I will merge it.
Do you think it could be worthwhile to extract the fresh classloader part to a junit rule?

timowest:
If it affects only the context class loader then maybe not, if it guarantees a fully independent classloading of the test class, then yes, because otherwise the contract might not be very clear.

Shredder121:
There you have a point. I think it is better like this then.

PR #1242, opened 2015-03-06
Improve Scala usage

timowest:
Changes are
* remove unnecessary semicolons
* follow Scala conventions in method invocation syntax
* fix lambda usage
* use file compilation instead of interpreter based compilation

backport https://github.com/querydsl/querydsl/pull/1258


timowest:
This seems to have stabilized, I have now replaced all interpreter based compilation with another compliation variant.

timowest:
@Shredder121 @johnktims Could one of you take a look at this? The `CompilerTestUtils` changes are the most significant. After this has been merged, the querydsl-scala failures should be (temporarily?) a thing of the past.

Shredder121:
To me it looks good, there are a few `reset()` methods that could be called (can't quite recall where I saw them though) but they are merely for resetting, and not for explicitly 'closing' a resource.

So I think it looks good. The Scala syntax is quite new to me though, but it doesn't read all that bad. At times.

timowest:
Ok, I will prepare the backport tomorrow.

PR #1243, opened 2015-03-06
Optimize exists queries

### LINKED ISSUE #1241: Inefficient implementation (in my case) of method com.querydsl.jpa.JPAQueryBase.exists()
[boolean com.querydsl.jpa.JPAQueryBase.exists()](https://github.com/querydsl/querydsl/blob/master/querydsl-jpa/src/main/java/com/querydsl/jpa/JPAQueryBase.java#L66-L73)
In my case I'm trying to check if active User presents in DB. For this purposes I use exists method. I expected that QueryDSL will generate one simple SQL request. But it generates not only one SQL request for extracting User, but also a few requests for extracting whole bunch of EAGER fields of my User model (believe me, I need them))). After that I saw [this piece of code](https://github.com/querydsl/querydsl/blob/master/querydsl-jpa/src/main/java/com/querydsl/jpa/JPAQueryBase.java#L66-L73) and I undertand that in fact there are two subvariants of exists method. But both of them first of all extracts full User model with LIMIT 1.
But it looks like that it would be better to use simple SELECT with COUNT(*) and check if result: 
`if result>1 than return true, else -> return false`.
I understand that for giant tables and very simple models COUNT will be working slower than SELECT with LIMIT 1, but in my case (User table has ~1500 rows) COUNT is working much more faster than current implementation of exists() method.

timowest:
Fixes #1241

### DIFF
--- a/querydsl-jpa/src/main/java/com/querydsl/jpa/EclipseLinkTemplates.java
+++ b/querydsl-jpa/src/main/java/com/querydsl/jpa/EclipseLinkTemplates.java
@@ -95,10 +95,4 @@ public class EclipseLinkTemplates extends JPQLTemplates {
         return false;
     }
 
-    @Override
-    public boolean isSelect1Supported() {
-        return true;
-    }
-
-
 }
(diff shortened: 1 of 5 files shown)


timowest:
@Shredder121 Could you take a look at this? I can make a backport as well, if it looks ok.

Shredder121:
Yes it looks good.
I shall merge it.

PR #1244, opened 2015-03-07
Use deterministic path creation

### LINKED ISSUE #1231: usage of any() pollutes query cache
We are using Spring Data together with query dsl 3.5.1.

We have several queries that make use of the any() expression.
When looking at the JPA query String that is created it shows an exists query with a table that contains a generated table name. More specifically the name seems to ends with some random alphanumeric characters.

Unfortunately each time querydsl generates the JPA query this name is different.
This in turn causes polution in the hibernate query cache.

Would it be possible to implement a naming strategy that produces reproducible names for the any()-tables?

timowest:
Fixes #1231 

The change in this PR is backwards incompatible, so a backport variant can't be provided.

### DIFF
--- a/querydsl-core/src/main/java/com/querydsl/core/support/CollectionAnyVisitor.java
+++ b/querydsl-core/src/main/java/com/querydsl/core/support/CollectionAnyVisitor.java
@@ -31,7 +31,7 @@ import com.querydsl.core.types.template.BooleanTemplate;
 @SuppressWarnings("unchecked")
 public class CollectionAnyVisitor implements Visitor<Expression<?>,Context> {
 
-    public static final CollectionAnyVisitor DEFAULT = new CollectionAnyVisitor();
+    private int replacedCounter;
 
     @SuppressWarnings("rawtypes")
     private static <T> Path<T> replaceParent(Path<T> path, Path<?> parent) {
@@ -110,7 +110,7 @@ public class CollectionAnyVisitor implements Visitor<Expression<?>,Context> {
             Path<?> parent = (Path<?>) expr.getMetadata().getParent().accept(this, context);
             expr = new PathImpl<Object>(expr.getType(), PathMetadataFactory.forCollectionAny(parent));
             EntityPath<?> replacement = new EntityPathBase<Object>(expr.getType(),
-                    ExpressionUtils.createRootVariable(expr));
+                    ExpressionUtils.createRootVariable(expr, replacedCounter++));
             context.add(expr, replacement);
             return replacement;
 
@@ -134,7 +134,5 @@ public class CollectionAnyVisitor implements Visitor<Expression<?>,Context> {
     public Expression<?> visit(ParamExpression<?> expr, Context context) {
         return expr;
     }
-
-    protected CollectionAnyVisitor() {}
 
 }
--- a/querydsl-core/src/main/java/com/querydsl/core/types/ExpressionUtils.java
+++ b/querydsl-core/src/main/java/com/querydsl/core/types/ExpressionUtils.java
@@ -13,9 +13,9 @@
  */
 package com.querydsl.core.types;
 
-import javax.annotation.Nullable;
-
 import java.util.*;
+
+import javax.annotation.Nullable;
 
 import com.google.common.collect.ImmutableList;
 import com.querydsl.core.BooleanBuilder;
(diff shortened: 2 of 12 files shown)


querydsl-core/src/main/java/com/querydsl/core/types/ExpressionUtils.java
```diff
@@ -533,14 +533,14 @@ public static Predicate or(Predicate left, Predicate right) {
...
-     * Create a new root variable based on the given path
+     * Create a new root variable based on the given path and suffix
      *
      * @param path
+     * @param suffix
      * @return
      */
-    public static String createRootVariable(Path<?> path) {
+    public static String createRootVariable(Path<?> path, int suffix) {
         String variable = path.accept(ToStringVisitor.DEFAULT, TEMPLATES).replace('.', '_');
```
Shredder121: The templates used are already UnderscoreTemplates, so is it necessary to replace '.' with '_'?

querydsl-core/src/main/java/com/querydsl/core/support/CollectionAnyVisitor.java
```diff
@@ -110,8 +113,9 @@ protected Predicate exists(Context c, Predicate condition) {
             Path<?> parent = (Path<?>) expr.getMetadata().getParent().accept(this, context);
             expr = new PathImpl<Object>(expr.getType(), PathMetadataFactory.forCollectionAny(parent));
             EntityPath<?> replacement = new EntityPathBase<Object>(expr.getType(),
-                    ExpressionUtils.createRootVariable(expr));
+                    ExpressionUtils.createRootVariable(expr, replaced.size()));
             context.add(expr, replacement);
+            replaced.add(expr);
```
Shredder121: Doesn't `replaced` only grow?

querydsl-core/src/test/java/com/querydsl/core/support/CollectionAnyVisitorTest.java
```diff
@@ -78,8 +78,5 @@ protected Predicate exists(Context c, Predicate condition) {
         Expression<?> transformed = expression.accept(visitor, new Context());
```
Shredder121: Keep in mind, that collectionAnyVisitor is now stateful.
The tests should reflect that as well.

Shredder121: At least to some degree.

querydsl-core/src/main/java/com/querydsl/core/support/CollectionAnyVisitor.java
timowest: Yes, shall we replace it with a simple counter? Putting the state into the visitor was easier than handling it via the context, because the context is new for each invocation.

timowest:
@Shredder121 Could you look again? I made some changes.

Shredder121:
Yes, I will.

querydsl-core/src/main/java/com/querydsl/core/support/CollectionAnyVisitor.java
```diff
@@ -110,7 +110,7 @@ protected Predicate exists(Context c, Predicate condition) {
             Path<?> parent = (Path<?>) expr.getMetadata().getParent().accept(this, context);
             expr = new PathImpl<Object>(expr.getType(), PathMetadataFactory.forCollectionAny(parent));
             EntityPath<?> replacement = new EntityPathBase<Object>(expr.getType(),
-                    ExpressionUtils.createRootVariable(expr));
+                    ExpressionUtils.createRootVariable(expr, replacedCounter++));
```
Shredder121: @timowest just a coding style question.
Is this statement okay for you? Most coding styles suggest that one statement (one line) should do one thing.
I was wondering what your opinion is, so that we can incorporate this into checkstyle (I think it has rules for that).

timowest: Do you mean the post increment operator? I am quite used to it, but we can do the incrementation also in a separate statement.

Shredder121: Yes I mean the post increment operator.
I am also used to it, and figured that you were as well, but since it's a part of coding style, it's good to think it through.
I don't mind if we keep it, I just wanted to bring it up for discussion.
@johnktims you can join us as well.

johnktims: I do prefer to minimize the number of statements in a single line but I think that this particular instance is probably fine.

#1192 might be a good place to discuss coding style.

Shredder121: > I think that this particular instance is probably fine.

I don't think discriminating is a good feature in coding style.
But yes, we'll take the conversation there!

Shredder121:
I think it looks good.
So I'll merge it then.

johnktims:
It looks good to me.

PR #1245, opened 2015-03-08
Improve normalization

### LINKED ISSUE #1224: Some strings are cutted when use literal mode
I've got a querydsl-sql (3.6.0) query like this

```java
		SQLQuery queryCheck = queryDslJdbcTemplateViaHibernateSession
				.getQuery()
				.from(isee159WsInpsAttestazioni_QSQL)
				.where(isee159WsInpsAttestazioni_QSQL.identificatoreRichiestaInps
						// .eq(typeAttestazione.getIdentificatoreRichiestaINPS()));
						.eq("INPS-ISET-0000-00000000A"));
                queryCheck.exists();
```
the output on jdbc side is

```sql
SELECT 1
FROM isee159_ws_inps_attestazioni isee159_ws_inps_attestazioni
WHERE isee159_ws_inps_attestazioni.identificatore_richiesta_inps = 'INPS-ISET0A' limit 1
```

the field is a string, as you can see the string is cutted on where clause **'INPS-ISET0A'**.
This behavior is only in literal mode, but I need it.

To put it in a nutshell I've made some tests and it sounds like '0' are packed inside the string.
Something like this:

* INPS-ISET-0000-12345678A  ==>  INPS-ISET-12345678A
* INPS-ISET-0000X00000000A ==>  INPS-ISET-0000X00000000A
* INPS-ISET-0000-00000000A  ==>  INPS-ISET0A

Please note that going inside the QueryBase object metadata.where has the right string while
metadata.self has the error. It sounds like a serialization issue.
Issue is not related to tables/query used. It happens on every query using where condition like that.
Any ideas?

johnktims:
Fixes #1224

### DIFF
--- a/querydsl-core/src/main/java/com/mysema/query/support/Normalization.java
+++ b/querydsl-core/src/main/java/com/mysema/query/support/Normalization.java
@@ -22,7 +22,7 @@ public final class Normalization {
 
     private static final String WS = "\\s*";
 
-    private static final String NUMBER = "([\\+\\-]?\\d+\\.?\\d*)";
+    private static final String NUMBER = "([\\+\\-]?\\d+\\.?\\d*)(?!\\p{Alpha})";
 
     // TODO simplify
     private static final Pattern FULL_OPERATION = Pattern.compile(
--- a/querydsl-core/src/test/java/com/mysema/query/support/NormalizationTest.java
+++ b/querydsl-core/src/test/java/com/mysema/query/support/NormalizationTest.java
@@ -65,6 +65,8 @@ public class NormalizationTest {
         assertEquals("-2.5", Normalization.normalize("2.5 * -1"));
         assertEquals("hours * 2 + 3", Normalization.normalize("hours * 2 + 3"));
         assertEquals("2 + 3 * hours", Normalization.normalize("2 + 3 * hours"));
+        assertEquals("2 + 3 * 0hours", Normalization.normalize("2 + 3 * 0hours"));
+        assertEquals("a like '1 + 2 ' and b like '2 * 3'", Normalization.normalize("a like '1 + 2 ' and b like '2 * 3'"));
     }
 
     @Test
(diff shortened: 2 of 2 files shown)


timowest:
@johnktims Thanks for the PR. Could you also add a test for quoted literals that have arithmetic expressions?

timowest:
This PR contains also normalization related improvements https://github.com/querydsl/querydsl/pull/1220

johnktims:
@timowest like
```java
assertEquals("'2 + 0hours'", Normalization.normalize("'2 + 0hours'"));
```
?

johnktims:
Do you want to just close this PR and make the changes in #1220 ?

timowest:
Maybe a literal like `a like '1 + 2 ' and b like '2 * 3'`

timowest:
No, the other PR covers only precedence rules, not literal detection. That should be handled in another PR, so it's ok to do it here.

Shredder121:
>a literal like `a like '1 + 2 '`

@timowest did you also want that space in the expression?

timowest:
The space came accidently, but maybe both could be used.

johnktims:
Are these normalizations really worthwhile? I imagine most people do their own computations before issuing the query instead of constructing a string of expressions, and I'm sure the database's query optimizer performs these sorts of optimizations anyway.

timowest:
These optimizations are not performance optimizations but more two compress the result of serialization patterns which contain arithmetic expressions. It will make debugging the output easier if arithmetic expressions are handled.

johnktims:
@timowest Should I close this in favor of #1252 ?

Shredder121:
They have different targets.
One is master, the other QUERYDSL_3.

johnktims:
Whoops. You're right.
I see that @timowest added a few more tests. Should I copy those over into this PR as well?

timowest:
Yes, you can copy the tests. I have usually combined commits quite alot via
rebases or git commit --amend, so that might be also here a nice way.

Shredder121:
Yes, just wanted to suggest that.

PR #1256, opened 2015-03-13
Remove ResourceCheck based test filtering

timowest:
backport https://github.com/querydsl/querydsl/pull/1257


johnktims:
Doesn't `ExportMSSQLTest` now require `@Category(ExternalDB.class)` to be excluded?

timowest:
Good point, I fixed that.

johnktims:
This looks good to me now.

PR #1258, opened 2015-03-14
Improve Scala usage

timowest:
backport of https://github.com/querydsl/querydsl/pull/1242

### DIFF
--- a/querydsl-scala/pom.xml
+++ b/querydsl-scala/pom.xml
@@ -18,7 +18,7 @@
     <character.encoding>UTF-8</character.encoding>
 
     <scala.version>2.10</scala.version>
-    <scala.full.version>2.10.4</scala.full.version>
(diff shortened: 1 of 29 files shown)


timowest:
@Shredder121 @johnktims Is this good as well? The main PR is also merged. 

Or shall we see a positive review comment on the main PR also as the OK for the backport PR?

johnktims:
This looks fine to me, but I'll admit that I'm not very familiar with Scala.

I think one review should suffice so long as the backport isn't significantly different.

Shredder121:
Yes, that is why backports generally can be merged within the 2 day mark, since the first PR is already merged.
Note that a review of backwards compatibility is necessary sometimes.

timowest:
Ok, let's do so. I am also for it.

PR #1213, opened 2015-02-18
Sync surefire dependencies

Shredder121:
The surefire artifact's versions were mismatched.
For some reason surefire 2.18.1 still caused issues, so I went for 2.18 instead.

PR #1259, opened 2015-03-14
Optimize surefire config in travis profile

timowest:
backports of https://github.com/querydsl/querydsl/pull/1186 and https://github.com/querydsl/querydsl/pull/1213


timowest:
Sorry, here the 2 days mark had not yet passed.

PR #1261, opened 2015-03-14
Changes to Expression instance creation

timowest:
Related to https://github.com/querydsl/querydsl/issues/706

Tasks
* [x] Rename com.querydsl.core.types.expr to com.querydsl.core.types.dsl
* [x] Move com.querydsl.support.Expressions to com.querydsl.core.types.dsl
* [x] Make constructors of SimpleExpression subclasses in those packages protected
* [x] Remove static factory methods from the same classes
* [x] Change Querydsl Spatial expression types to use same creation pattern
* [x] Combine contents of com.querydsl.core.types.expr, path, query and template
* [x] Add simple* methods to Expressions
* [x] Consistent generic signature

This PR doesn't have a backport variant, since it is a backwards incompatible change.


timowest:
I changed it now to have no static imports to have a consistent baseline. 

Unfortunately there seem to be some querydsl-scala stability issues left.

Shredder121:
You can call `interpreter.reset` in the finally block of `assertCompileSuccess()`.
I don't quite know if it helps, but it does more than `interpreter.close()`.
The rest should be handled by the GC.

timowest:
The Scala compilation is not anymore interpreter based.

Shredder121:
And what about `Global.cleanup()`?

timowest:
`cleanup()` seems to be some kind of accessor.

timowest:
@johnktims @Shredder121 This is ready to review now.

Shredder121:
Then I'll do so.

Shredder121:
I'll update this branch.

Shredder121:
Since you relaxed the generics a bit there are a few redundant casts that the compiler can figure out on its own.
(I'll only mention a few)

Shredder121:
@timowest do you want me to update your commits or do you want to do it yourself?
Or do you think that one additional commit is cleaner?

timowest:
Thanks a lot for the review. I can do the updates.

Shredder121:
Keep in mind that I haven't named them all, but it should get you on the right track ;)

timowest:
I removed now all redundant Expression related casts.

querydsl-core/src/test/java/com/querydsl/core/types/dsl/PathBuilderValidatorTest.java
```diff
@@ -7,6 +7,8 @@
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNull;
 
+import com.querydsl.core.types.dsl.PathBuilderValidator;
```
Shredder121: Could you run through the imports to remove the imports from the same package?

querydsl-core/src/main/java/com/querydsl/core/types/dsl/CaseBuilder.java
```diff
@@ -186,7 +186,7 @@ public Initial(Predicate b) {
             return new Cases<A, SimpleExpression<A>>((Class)expr.getType()) {
```
Shredder121: I think relaxing the type parameter in `Cases`' `type`,`createResult` and constructor to `Class<? extends A>` helps a lot too to eliminate these casts **and** `@SuppressWarnings`.

Shredder121:
There are quite a few more occurrences in `SQLExpressions`

querydsl-core/src/main/java/com/querydsl/core/types/dsl/DateTimeExpression.java
```diff
@@ -55,7 +55,7 @@
      * @return
      */
     public static <T extends Comparable> DateTimeExpression<T> currentDate(Class<T> cl) {
-        return DateTimeOperation.<T>create(cl, Ops.DateTimeOps.CURRENT_DATE);
+        return Expressions.<T>dateTimeOperation(cl, Ops.DateTimeOps.CURRENT_DATE);
```
Shredder121: This explicit type argument is not necessary.

querydsl-core/src/main/java/com/querydsl/core/types/dsl/Expressions.java
```diff
@@ -0,0 +1,1296 @@
...
+
+    public static final NumberExpression<Integer> ZERO = numberTemplate(Integer.class, "0");
+
+    public static final BooleanExpression TRUE = booleanTemplate("true");
+
+    public static final BooleanExpression FALSE = booleanTemplate("false");
+
+    private Expressions() {}
+
+    @SuppressWarnings("unchecked")
```
Shredder121: This `@SuppressWarnings` is now not necessary.

querydsl-core/src/main/java/com/querydsl/core/types/dsl/SimpleExpression.java
```diff
@@ -59,7 +59,7 @@ public SimpleExpression(Expression<T> mixin) {
     @Override
     @SuppressWarnings("unchecked")
```
Shredder121: This `@SuppressWarnings` is now not necessary anymore.

querydsl-core/src/main/java/com/querydsl/core/types/dsl/SimpleExpression.java
```diff
@@ -289,7 +289,7 @@ public final BooleanExpression notIn(CollectionExpression<?,? extends T> right)
      */
     @SuppressWarnings("unchecked")
```
Shredder121: This `@SuppressWarnings` is now not necessary anymore.

querydsl-core/src/main/java/com/querydsl/core/types/dsl/TimeExpression.java
```diff
@@ -11,7 +11,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package com.querydsl.core.types.expr;
+package com.querydsl.core.types.dsl;
 
 import java.sql.Time;
```
Shredder121: There are also instances of `@SuppressWarnings` [on the type level, like right under this comment](https://github.com/querydsl/querydsl/pull/1261/files#diff-0ef6d19b79529e691ba93255126408f7R32).
These are now not necessary anymore, and not desired.

timowest:
@Shredder121 Ready to merge or still something to improve?

Shredder121:
We could merge `RelationalFunctionCall`'s creation methods and move it to `Expressions`, what do you think?

timowest:
It's in `querydsl-sql`, so maybe `SQLExpressions`?

Shredder121:
Ah indeed.
Yes that would work.

Shredder121:
> Is there something more to do here?

The fixes are getting rare. We could revise and relax all creation methods or handle it optionally in another PR?
Because I think we have fixed it all.

timowest:
Maybe another PR. I am also going to move the ExpressionUtils usage into a similar direction so have the creation of the simple Expression types follow the same pattern.

PR #1265, opened 2015-03-17
Normalize non public classes to visible classes

### LINKED ISSUE #1262: querydsl-collections: Evaluator compilation failure
As discussed on groups here: https://groups.google.com/forum/#!topic/querydsl/NPrkWsoq-iM

The generated code compilation fails when we use guava's Optional, or more generally when using some non public class (like com.google.common.base.Present implicitely used through Optional).

Reproduction steps: just run the given gist https://gist.github.com/jotak/54e46b022f6b700a5f91

        Error: 
        Exception in thread "main" com.mysema.codegen.CodegenException: Compilation of public class Q_0390602011_1275614662_1275614662_566403833 {
    
        public static Iterable<cat.Cat> eval(Iterable<cat.Cat> cat_, com.google.common.base.Present a1) {
            java.util.List<cat.Cat> rv = new java.util.ArrayList<cat.Cat>();
            for (cat.Cat cat : cat_) {
                try {
                    if (com.mysema.query.collections.CollQueryFunctions.equals(cat.getPedigree(), a1)) {
                        rv.add(cat);
                    }
                } catch (NullPointerException npe) { }
            }
            return rv;    }
    
    }
    
    failed.
            /Q_0390602011_1275614662_1275614662_566403833.java:3: error: Present is not public in com.google.common.base; cannot be accessed from outside package
    public static Iterable<cat.Cat> eval(Iterable<cat.Cat> cat_, com.google.common.base.Present a1) {
            ^
            1 error
    
            at com.mysema.codegen.JDKEvaluatorFactory.compile(JDKEvaluatorFactory.java:74)
            at com.mysema.codegen.AbstractEvaluatorFactory.createEvaluator(AbstractEvaluatorFactory.java:128)

As said in groups post, it looks like the signature of the generated "eval" method is wrong, as it uses a non accessible class "com.google.common.base.Present" instead of its base class Optional. Maybe the signature generation should look for the first public class in ancestors?

By the way, do you see any workaround for me, so I can use Optional fields?

I'm using querydsl 3.3.3, I tried with the latest (3.6.2) but the issue is still there.

Thanks

JT

Shredder121:
Will fix this issue #1262 together with #1290 

We can discuss the changes needed here.

### DIFF
--- a/querydsl-collections/src/test/java/com/querydsl/collections/GuavaOptionalTest.java
+++ b/querydsl-collections/src/test/java/com/querydsl/collections/GuavaOptionalTest.java
@@ -1,0 +1,49 @@
+package com.querydsl.collections;
+
+import static com.querydsl.collections.CollQueryFactory.from;
+import static com.querydsl.core.alias.Alias.$;
+import static com.querydsl.core.alias.Alias.alias;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+import java.util.Collection;
+
+import org.junit.Test;
+
+import com.google.common.base.Optional;
+import com.google.common.collect.Lists;
+
+public class GuavaOptionalTest {
+
+    public static class Cat {
+
+        private Optional<String> pedigree = Optional.absent();
+
+        public Cat() {
+        }
+
+        public Cat(Optional<String> pedigree) {
+            this.pedigree = pedigree;
+        }
+
+        public Optional<String> getPedigree() {
+            return pedigree;
+        }
+    }
+
+    @Test
+    public void Test() {
+        Collection<Cat> cats = Lists.newArrayList();
+        cats.add(new Cat(Optional.<String>absent()));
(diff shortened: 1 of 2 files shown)


querydsl-collections/src/test/java/com/querydsl/collections/GuavaOptionalTest.java
```diff
@@ -0,0 +1,49 @@
...
+        public Optional<String> getPedigree() {
+            return pedigree;
+        }
+    }
+
+    @Test
+    public void Test() {
+        Collection<Cat> cats = Lists.newArrayList();
+        cats.add(new Cat(Optional.<String>absent()));
+        cats.add(new Cat(Optional.of("persan")));
```
johnktims: *persian

Shredder121:
@johnktims does it look good?
I will backport this to the QUERYDSL_3 branch, so it can make the 1 day review window for the release.

johnktims:
Yes. It looks good to me.

PR #1268, opened 2015-03-18
Fix isNotEmpty bug in mongodb serialization

### LINKED ISSUE #1264: MongodbSerializer does not serialize isNotEmpty() properly
Original link: http://stackoverflow.com/questions/29104204/querydsl-isnotempty-expression-failing

There appears to be a serialization error in `MongodbSerializer` for
```java
BooleanExpression expr = qa.matches.isNotEmpty();
Iterable<A> result = aRepo.findAll(expr);

assertThat(result, is(not((emptyIterable()))));
```

` MongodbSerializer` is spitting out
```json
{ "$or" : { "$not" : [ { "matches" : [ ]} , { "matches" : { "$exists" : false}}]}}
```
which fails when I paste it into mongo.

```
> db.a.find({ "$or" : { "$not" : [ { "matches" : [ ]} , { "matches" : { "$exists" : false}}]}})
Error: error: {
	"$err" : "Can't canonicalize query: BadValue $or needs an array",
	"code" : 17287
}
> 
```

I rewrote the query as
```
> db.a.find({ "$or" : [ { "matches" : {$ne: [ ]}}, { "matches" : { "$exists" : false}}]}).pretty()
{
	"_id" : "a1",
	"_class" : "querydsldemo.A",
	"matches" : [
		{
			"_id" : ObjectId("55088ec4e4b0b6e56e76b1d6"),
			"name" : "b1"
		},
		{
			"_id" : ObjectId("55088ec4e4b0b6e56e76b1d7"),
			"name" : "b2"
		}
	]
}
{ "_id" : "a3", "_class" : "querydsldemo.A" }
```
and then further simplified it as
```
> db.a.find({ "matches" : {$ne: [ ]}}).pretty()
{
	"_id" : "a1",
	"_class" : "querydsldemo.A",
	"matches" : [
		{
			"_id" : ObjectId("55088ec4e4b0b6e56e76b1d6"),
			"name" : "b1"
		},
		{
			"_id" : ObjectId("55088ec4e4b0b6e56e76b1d7"),
			"name" : "b2"
		}
	]
}
{ "_id" : "a3", "_class" : "querydsldemo.A" }
> 
```
but I don't know if that's robust enough.

johnktims:
Fixes #1264 

I backported the fix and verified that the tests in the original SO issue pass.

### DIFF
--- a/querydsl-mongodb/src/main/java/com/querydsl/mongodb/MongodbSerializer.java
+++ b/querydsl-mongodb/src/main/java/com/querydsl/mongodb/MongodbSerializer.java
@@ -15,6 +15,7 @@ package com.querydsl.mongodb;
 
 import java.util.Collection;
 import java.util.List;
+import java.util.Map;
 import java.util.regex.Pattern;
 
 import org.bson.BSONObject;
@@ -117,23 +118,14 @@ public abstract class MongodbSerializer implements Visitor<Object, Void> {
 
         } else if (op == Ops.NOT) {
             //Handle the not's child
-            BasicDBObject arg = (BasicDBObject) handle(expr.getArg(0));
-
-            //Only support the first key, let's see if there
-            //is cases where this will get broken
-            String key = arg.keySet().iterator().next();
             Operation<?> subOperation = (Operation<?>) expr.getArg(0);
             Operator subOp = subOperation.getOperator();
             if (subOp == Ops.IN) {
                 return visit(OperationImpl.create(Boolean.class, Ops.NOT_IN, subOperation.getArg(0),
                         subOperation.getArg(1)), context);
-            } else if ("$or".equals(key)) {
-                return asDBObject("$nor", arg.get(key));
-
-            } else if (subOp != Ops.EQ && subOp != Ops.STRING_IS_EMPTY) {
-                return asDBObject(key, asDBObject("$not", arg.get(key)));
             } else {
-                return asDBObject(key, asDBObject("$ne", arg.get(key)));
+                BasicDBObject arg = (BasicDBObject) handle(expr.getArg(0));
+                return negate(arg);
             }
 
         } else if (op == Ops.OR) {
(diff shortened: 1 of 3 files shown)


querydsl-mongodb/src/main/java/com/querydsl/mongodb/MongodbSerializer.java
```diff
@@ -127,6 +127,11 @@ public Object visit(Operation<?> expr, Void context) {
             if (subOp == Ops.IN) {
                 return visit(OperationImpl.create(Boolean.class, Ops.NOT_IN, subOperation.getArg(0),
                         subOperation.getArg(1)), context);
+            } else if (subOp == Ops.COL_IS_EMPTY) {
```
timowest: alternatively this could be also a general inversion of `$or` to `$nor`

johnktims: Can you give me a little more detail? I don't think I understand.

timowest: I meant to detect $or expressions by looking at the nested result and
replacing $or with $nor.

Shredder121: Do you mean instead of relying on the implementation of `COL_IS_EMPTY` (which uses $or) detecting $or and change that to $nor?
How deep do you want to detect?
If there is an AND involved, with two ORs which have two leaves (`x.isEmpty().and(y.isEmpty()).not()` it could cause problems.
Does Inverting the $or then work?

timowest: Yes, that's what I mean, and only on the first level. Replacing the top level $or with $nor should work.

timowest:
@johnktims Could you also add a test in `MongodbQueryTest`?

timowest:
I suggest to test for all predicates in `MongodbQueryTest` also the negation via `predicate.not()`. That way we might catch also other similar issues.

Shredder121:
I agree.
I have also looked at other [logical operators](http://docs.mongodb.org/manual/reference/operator/query-logical/), but there are not any special ones we need to consider.

timowest:
I'll push soon some further improvements.

Shredder121:
Can you also update this branch while you are at it?
There seem to have been simultaneous changes to the import order of MongodbQueryTest.

querydsl-mongodb/src/main/java/com/querydsl/mongodb/MongodbSerializer.java
```diff
@@ -118,19 +119,30 @@ public Object visit(Operation<?> expr, Void context) {
...
                 return visit(OperationImpl.create(Boolean.class, Ops.NOT_IN, subOperation.getArg(0),
                         subOperation.getArg(1)), context);
-            } else if (subOp != Ops.EQ && subOp != Ops.STRING_IS_EMPTY) {
-                return asDBObject(key, asDBObject("$not", arg.get(key)));
+            } else if (arg.size() == 1) {
+                Map.Entry<String, Object> entry = arg.entrySet().iterator().next();
+                if (entry.getKey().equals("$or")) {
+                    return asDBObject("$nor", entry.getValue());
+                } else if (entry.getKey().equals("$and")) {
+                    return asDBObject("$nand", entry.getValue());
```
Shredder121: Where did you find this operator?

timowest:
There are still failing test cases, for example `{ "age" : { "$not" : { "$gte" : 20 , "$lte" : 30}}}`.

It seems to be interpreted as `(not (age >= 20)) and (not (age <= 30))` instead of `not (age >= 20 and age <= 30))`.

querydsl-mongodb/src/main/java/com/querydsl/mongodb/MongodbSerializer.java
Shredder121: Relevant SO post:
http://stackoverflow.com/questions/20810698/mongodb-how-to-query-with-nand-operator.

timowest: Good point, I'll fix it.

timowest:

johnktims:
I only visually inspected it but it looks good to me.

Issue #1132, opened 2015-01-24
querydsl-root POM changes


Shredder121:
As querydsl-root's parent is 'mysema-source', it inherits its properties.

This means that the `<organisation>` of Querydsl is still 'Mysema Ltd'.

It should be a better idea to replace (or eliminate) 'mysema-source' as parent.

timowest:
Maybe com.querydsl:querydsl-parent as a separate github project could work.

Shredder121:
Just a few more ideas for the common parent:

* unify the test configuration
 * probably look at including `TravisFoldListener` in some way?
* code style checks

anything else?

johnktims:
What do you think about moving the contents of querydsl-root into the root directory of the project?

That seems to be the generally-accepted way of configuring a multi-module project but it does result in a more cluttered root directory.

timowest:
@johnktims Moving the root to the root of the project is fine with me. But preferably only on the master side, not backport. @Shredder121 What do you think?

Shredder121:
@timowest Just your opinion, why not on the backport?

In my opinion this makes it better to support the development platforms that silently support only the common practice. (various cloud IDEs I tried don't like it when there is 'no pom.xml' and generate one for you..)
So I think this is better. It also means we don't have an (almost) empty GitHub repository only for the parent root.

timowest:
Yes, technically backwards compatibility is preserved, so the backport should be ok as well then

Shredder121:
If you have an opinion, you can share it of course.

timowest:
I'd like to keep the changes to the QUERYDSL_3 branch under the surface, on the other hand we agreed that backwards compatible changes should be applied there. So I am trying to be consistent and agree with you ;)

Shredder121:
So to move this along, we can backport it now?

johnktims:
I'll backport #1270 and my documentation fixes as soon as I get home.

Shredder121:
That is fine.

PR #1270, opened 2015-03-18
Move querydsl-root to root directory

johnktims:
Part of #1132 

I'll backport it if this looks good.


dist.sh
```diff
@@ -14,11 +14,11 @@ done
 
 echo "Creating reference documentation"
 cd ../querydsl-docs
```
Shredder121: `dist.sh` is already in the root. So this must be `cd querydsl-docs`.

dist.sh
```diff
@@ -14,11 +14,11 @@ done
...
-mkdir -p ../querydsl-root/target/dist/reference
+mkdir -p ../target/dist/reference
 mvn -Dxslthl.config=http://docbook.sourceforge.net/release/xsl/current/highlighting/xslthl-config.xml clean package
-cp -R target/docbook/publish/en-US/* ../querydsl-root/target/dist/reference/
+cp -R target/docbook/publish/en-US/* ../target/dist/reference/
 mvn -Dxslthl.config=http://docbook.sourceforge.net/release/xsl/current/highlighting/xslthl-config.xml -Dtranslation=ko-KR clean package
-cp -R target/docbook/publish/ko-KR ../querydsl-root/target/dist/reference/
-cd ../querydsl-root
+cp -R target/docbook/publish/ko-KR ../target/dist/reference/
+cd ../
```
Shredder121: The same here, `dist.sh` is already in the root.

dist.sh
Shredder121: Also, this is dependent on line 11.
I think t's better to store `pwd` beforehand, and append the modules to that. Since that won't work anymore.

timowest:
It looks good now in my opinion.

Shredder121:
I am still looking. One moment.

Shredder121:
It looks good in my opinion as wel.

PR #1276, opened 2015-03-25
Fixed issue with order by expressions for non-root entity paths

spelchri:
The same join is added multiple times when using multiple order-by expressions for non-root entity paths. This is a possible fix.


timowest:
@spelchri Thanks a lot for the PR. I replaced the logic change with something simpler in a new PR. I will also backport it to `QUERYDSL_3`.

PR #1277, opened 2015-03-25
Fix issue with order by expressions for non root paths

timowest:
based on https://github.com/querydsl/querydsl/pull/1276
backport https://github.com/querydsl/querydsl/pull/1278

### DIFF
--- a/querydsl-jpa/src/main/java/com/querydsl/jpa/JPAQueryMixin.java
+++ b/querydsl-jpa/src/main/java/com/querydsl/jpa/JPAQueryMixin.java
@@ -146,6 +146,7 @@ public class JPAQueryMixin<T> extends QueryMixin<T> {
             Path<T> oldPath = new PathImpl<T>(path.getType(),
                     new PathMetadata(parent, metadata.getElement(), metadata.getPathType()));
             Path<T> newPath = new PathImpl<T>(type, oldPath.toString().replace('.', '_'));
+            aliases.put(path, newPath);
             leftJoin(oldPath, newPath);
             return newPath;
         }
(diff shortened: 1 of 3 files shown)


Shredder121:
Well, in my opinion it looks straightforward, I finished reviewing.
The only thing we *could* enhance is using UnderscoreTemplates, instead of [`oldPath.toString().replace('.', '_')`](https://github.com/querydsl/querydsl/pull/1277/files#diff-5ea4f95c4cb3ea30a76656d43e421599R148).

Shredder121:
This master PR just needs to be rebased, but in my opinion they can be merged.

querydsl-jpa/src/main/java/com/querydsl/jpa/JPAQueryMixin.java
```diff
@@ -145,7 +145,8 @@ private boolean isEntityPath(Path<?> path) {
             Path<?> parent = shorten(metadata.getParent(), paths);
             Path<T> oldPath = new PathImpl<T>(path.getType(),
                     new PathMetadata(parent, metadata.getElement(), metadata.getPathType()));
-            Path<T> newPath = new PathImpl<T>(type, oldPath.toString().replace('.', '_'));
+            Path<T> newPath = new PathImpl<T>(type, ExpressionUtils.createRootVariable(oldPath));
```
Shredder121: We could use `ExpressionsUtils.path` for the master PR.

timowest: that's not available, so maybe do the change here https://github.com/querydsl/querydsl/pull/1285 ?

PR #1278, opened 2015-03-25
Fix issue with order by expressions for non root paths

timowest:
based on https://github.com/querydsl/querydsl/pull/1276
backport of https://github.com/querydsl/querydsl/pull/1277


spelchri:
Thanks for including the PR. Will it make it into 3.6.3?

Shredder121:
The PRs look good to me, so I think it should in my opinion.

timowest:
@spelchri Yes, it will make it into 3.6.3.

PR #1282, opened 2015-03-26
Remove unnecessary reflection code from JavaTemplates

Shredder121:


querydsl-core/src/test/java/com/querydsl/core/serialization/JavaTemplatesTest.java
```diff
@@ -16,27 +16,21 @@
...
+        Templates templates = new JavaTemplates();
         int matched = 0;
-        for (Field field : Ops.class.getFields()) {
-            if (Operator.class.isAssignableFrom(field.getType())) {
-                matched++;
-                Operator operator = (Operator) field.get(null);
-                assertNotNull(field.getName() + " missing", templates.getTemplate(operator));
-            }
+        for (Operator operator : Ops.values()) {
+            ++matched;
```
Shredder121: This doesn't do anything anymore.
you could assert that the `Ops.values()` is not empty :smile:

johnktims:
I contemplated removing it and just decided to leave it. Ops is never going to be empty so I think I can safely remove that final assertion altogether.

querydsl-core/src/test/java/com/querydsl/core/serialization/JavaTemplatesTest.java
johnktims: I contemplated removing it and just decided to leave it. Ops is never going to be empty so I think I can safely remove that final assertion altogether.

Shredder121:
I am rebasing this, fixup for [imports of 871b998](https://github.com/querydsl/querydsl/blob/871b998ed4c1fa08f7a7b2f14533a72191e92878/querydsl-core/src/main/java/com/querydsl/core/types/JavaTemplates.java#L16).

PR #1288, opened 2015-03-28
Add fix for extending external types

Shredder121:


Shredder121:
Using the creation methods for `// custom` and `// inherited` properties might be safer, if you for example have no control of a custom entity (e.g. external jar).

querydsl-apt/src/test/apt/com/querydsl/ExampleEntity2.java
```diff
@@ -0,0 +1,10 @@
+package com.querydsl;
+
+import com.querydsl.core.annotations.QueryEntity;
+
+@QueryEntity
+public class ExampleEntity2 {
```
Shredder121: I forgot one detail of this test case, it should extend an external `@QueryEntity` so that it inherits a `NumberPath` that is normally not assignable to a QInteger.
```java
private final QInteger integerProp = _super.integerProp; // doesn't work.
```

Shredder121:
There is only a soft `progress` label, we can discuss about this PR.

timowest:
This looks good in my opinion, the change is well scoped.

Shredder121:
That is what I think as well, but what happens when you want to extend a custom type?
So if in a jar QInteger (or any QEntity) is present, and you want to add a `@QueryDelegate` for `Integer`?

timowest:
For that case maybe another issue could be opened? I believe with the current logic it would create duplicate classes.

Shredder121:
Yes, and maybe it won't occur that much.
I think that this fix is at least ready to merge.
It is up to you if it is good to merge.

timowest:
I think it's fine, but could you also provide the backport? Then they could be merged as a pair.

Shredder121:
I am doing that now, after I said that I think it's ready.
So it is on its way.

PR #1296, opened 2015-04-01
Improve case operation serialization

### LINKED ISSUE #1286: Unable to use custom UserType in when
I'm not able to use custom UserType in when statement in **3.6.0+**. The following exception is always thrown.

```
Exception in thread "main" java.lang.IllegalArgumentException: Unsupported constant Approved
```

It works fine with **3.5.1**.

sample query:
```java
QRequest r = QRequest.request;
JPAQuery q = new JPAQuery(em);
q.from(r);
q.orderBy(r.status.when(new Approved()).then(0).otherwise(1).asc());
q.list(r);
```
status field is defined as:
```java
...
@Type(type = "com.xyz.status.StatusType")
private RequestStatus status;
...
```

StatusType is custom UserType which stores java object to database as string 

```java
...
public void nullSafeSet(PreparedStatement st, Object value, int index, SessionImplementor session) throws HibernateException, SQLException {
		if (value == null) {
			st.setNull(index, StringType.INSTANCE.sqlType());
		} else {
			st.setString(index, value.getClass().getSimpleName());
		}
	}
...
```

timowest:
Fixes #1286
backport https://github.com/querydsl/querydsl/pull/1299

### DIFF
--- a/querydsl-jpa/src/main/java/com/querydsl/jpa/JPQLSerializer.java
+++ b/querydsl-jpa/src/main/java/com/querydsl/jpa/JPQLSerializer.java
@@ -42,6 +42,9 @@ public class JPQLSerializer extends SerializerBase<JPQLSerializer> {
             Ops.ADD, Ops.SUB, Ops.MULT, Ops.DIV,
             Ops.LT, Ops.LOE, Ops.GT, Ops.GOE, Ops.BETWEEN);
 
+    private static final Set<Operator> CASE_OPS = ImmutableSet.<Operator>of(
+            Ops.CASE_EQ_ELSE, Ops.CASE_ELSE);
+
     private static final String COMMA = ", ";
 
     private static final String DELETE = "delete from ";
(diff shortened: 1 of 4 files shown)


querydsl-jpa/src/main/java/com/querydsl/jpa/JPQLSerializer.java
```diff
@@ -337,7 +340,7 @@ public void visitLiteral(Object constant) {
         } else if (constant instanceof Enum) {
```
Shredder121: For consistency the `instanceof` can also be `Enum<?>`.

querydsl-jpa/src/test/java/com/querydsl/jpa/AbstractJPATest.java
```diff
@@ -319,6 +319,11 @@ public void Case3() {
     }
 
     @Test
+    public void Case4() {
+        query().from(cat).list(cat.mate.when(savedCats.get(0)).then(0).otherwise(1));
```
johnktims: You might want to assert something here.

Also, you added a test named `Case4` to #1305 .

timowest:
I added some assertions.

Shredder121:
EclipseLink seems to return a `Vector` of `Long`s.

timowest:
Yes, I am working on it right now.

timowest:
I put the fix here https://github.com/querydsl/querydsl/pull/1305

Shredder121:
@johnktims This one looks good to me.

PR #1297, opened 2015-04-01
Fix literal escaping

### LINKED ISSUE #1275: Line breaks not created/escaped correctly in WHERE clauses
As described here: https://groups.google.com/forum/#!topic/querydsl/z08cNws-WSI

Using querydsl-sql 3.6.2 with MS SQL and MySQL for literal query generation. Simple example:
```java
private String createContainsQuery(final SQLTemplates template, final String filter) {
        final Configuration configuration = new Configuration(template);
        configuration.setUseLiterals(true);
        final SimplePath<Object> table = Expressions.path(Object.class, "Tablename");
        final StringPath column = Expressions.stringPath(table, "aColumn");
        final SQLQuery sqlQuery = new SQLQuery(configuration).from(table).where(column.contains(filter));
        return sqlQuery.getSQL(column).getSQL();
    }
```

Creating contains and equals statements with a line break ("foo\nbar") or a literal "\n" ("foo\\nbar") leads to the following reults:

### SQL Server 

* Contains with line break:
```sql
select Tablename.aColumn
from Tablename
where Tablename.aColumn like '%foo\nbar%' escape '\'
```
This does not work since SQL server (at least 2014) does not seem to accept a literal "\n" as a line break, only **'%foo' + CHAR(13)  + CHAR(10) + 'bar%'** or **'%foo' + CHAR(10) + 'bar%'** (depending on the OS) seem to work.
 
* Contains with literal \n:
```sql
select Tablename.aColumn
from Tablename
where Tablename.aColumn like '%foo\\nbar%' escape '\'
```
This does work although **'%foo\nbar%'** (without escape) would be sufficient
 
* Equals with line break:
```sql
select Tablename.aColumn
from Tablename
where Tablename.aColumn = 'foo\nbar'
```
* Equals with literal \n:
```sql
select Tablename.aColumn
from Tablename
where Tablename.aColumn = 'foo\nbar'
```
One query should be looking for a line break and the other should be looking for a literal "\n" but both produce the same output. (with the latter beeing correct)
 
### MySQL

* Contains with line break:
```sql
select Tablename.aColumn
from Tablename
where Tablename.aColumn like '%foo\\nbar%'
```
The "\n" is escaped wrongly and thus the query is looking for the literal "\n" and not for a line break.
 
* Contains with literal \n:
```sql
select Tablename.aColumn
from Tablename
where Tablename.aColumn like '%foo\\\\nbar%'
```
This query searches for the literal "\\\\n" (two backslashes) rather than the literal "\n"
 
* Equals with line break:
```sql
select Tablename.aColumn
from Tablename
where Tablename.aColumn = 'foo\\nbar'
```
* Equals with literal \n:
```sql
select Tablename.aColumn
from Tablename
where Tablename.aColumn = 'foo\\nbar'
 ```
Again both queries produce the same output although they should search for different results.

timowest:
Fixes #1275

### DIFF
--- a/querydsl-sql/src/main/java/com/querydsl/sql/SQLTemplates.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/SQLTemplates.java
@@ -503,13 +503,7 @@ public class SQLTemplates extends Templates {
     public String escapeLiteral(String str) {
         StringBuilder builder = new StringBuilder();
         for (char ch : str.toCharArray()) {
-            if (ch == '\n') {
-                builder.append("\\n");
-                continue;
-            } else if (ch == '\r') {
-                builder.append("\\r");
-                continue;
-            } else if (ch == '\'') {
+            if (ch == '\'') {
                 builder.append("''");
                 continue;
             }
--- a/querydsl-sql/src/test/java/com/querydsl/sql/SelectBase.java
+++ b/querydsl-sql/src/test/java/com/querydsl/sql/SelectBase.java
@@ -879,12 +879,14 @@ public class SelectBase extends AbstractBaseTest {
     @Test
     @ExcludeIn(FIREBIRD)
     public void Like_Escape() {
-        List<String> strs = ImmutableList.of("%a", "a%", "%a%", "_a", "a_", "_a_", "[C-P]arsen");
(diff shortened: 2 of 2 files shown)


Shredder121:
To me these changes look good, as long as the `''` cannot be prematurely closed (e.g. SQL injection)

timowest:
Yes, that's guaranteed.

PR #1305, opened 2015-04-05
Fix case operation precedence

### LINKED ISSUE #1303: CaseBuilder() is throwing error
I am trying to create Case Expression 

```java
NumberExpression<Float> numExpression = (summary.field1.floatValue().divide(summary.field2.floatValue()).multiply(100)); 		
NumberExpression<Float> numberExpression2 = summary.impressions.when(0L).then(0.0F).otherwise(numExpression);
```

So the formed SQL Query will contain : 

```sql
when summary.impression=0 then 0.0
else cast summay.field as flaot/ cast summary.field2 as float) * 100 end
```

it throwing error at * ...unexpected token end i guess not allowing any operators we can expect same error for / also

timowest:
Fixes #1303 
backport https://github.com/querydsl/querydsl/pull/1307

### DIFF
--- a/querydsl-jpa/src/test/java/com/querydsl/jpa/AbstractJPATest.java
+++ b/querydsl-jpa/src/test/java/com/querydsl/jpa/AbstractJPATest.java
@@ -319,6 +319,14 @@ public abstract class AbstractJPATest {
     }
 
     @Test
+    @ExcludeIn(MYSQL) // doesn't work in Eclipselink
+    public void Case4() {
+        NumberExpression<Float> numExpression = cat.bodyWeight.floatValue().divide(otherCat.bodyWeight.floatValue()).multiply(100);
+        NumberExpression<Float> numExpression2 = cat.id.when(0).then(0.0F).otherwise(numExpression);
+        query().from(cat, otherCat).list(numExpression2);
+    }
+
+    @Test
     public void CaseBuilder() {
         QCat cat2 = new QCat("cat2");
         NumberExpression<Integer> casex = new CaseBuilder()
--- a/querydsl-jpa/src/test/java/com/querydsl/jpa/AbstractJPATest.java
+++ b/querydsl-jpa/src/test/java/com/querydsl/jpa/AbstractJPATest.java
@@ -305,6 +305,14 @@ public abstract class AbstractJPATest {
     }
 
     @Test
+    @NoBatooJPA
+    public void Case_Long() {
+        assertEquals(ImmutableList.of(1L, 2L, 2L, 2L, 2L, 2L),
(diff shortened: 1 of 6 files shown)


querydsl-jpa/src/test/java/com/querydsl/jpa/AbstractJPATest.java
```diff
@@ -319,6 +319,14 @@ public void Case3() {
     }
 
     @Test
+    @ExcludeIn(MYSQL) // doesn't work in Eclipselink
+    public void Case4() {
+        NumberExpression<Float> numExpression = cat.bodyWeight.floatValue().divide(otherCat.bodyWeight.floatValue()).multiply(100);
+        NumberExpression<Float> numExpression2 = cat.id.when(0).then(0.0F).otherwise(numExpression);
+        query().from(cat, otherCat).list(numExpression2);
```
johnktims: Shouldn't these tests make assertions or are you just assuming that if an exception isn't thrown it's working?

querydsl-sql/src/test/java/com/querydsl/sql/SelectBase.java
```diff
@@ -232,6 +232,13 @@ public void Boolean_Any() {
     }
 
     @Test
+    public void Case() {
```
johnktims: This test passes even if I revert the fix.

johnktims: I see you only updated `JPQLTemplates` so I'm guessing the issue only existed in JPA and you're just adding a test for the SQL suite.

timowest: the SQL side works, but I added a test to assert that also.

querydsl-jpa/src/test/java/com/querydsl/jpa/AbstractJPATest.java
timowest: Yes, some assertions are missing, I will add it here and more in a future PR.

timowest:
I added now assertions.

querydsl-sql/src/test/java/com/querydsl/sql/SelectBase.java
johnktims: @timowest I tried to add an assertion here but H2, Derby, and Hsql seem to order the result differently. Can you tell what I'm missing here? I'll look at it tomorrow if you haven't fixed it.

timowest:
One way to get consistent results is to use `numExpression2.floor().intValue()`, looks like the rounding behaviour in the casts is not consistent.

PR #1306, opened 2015-04-05
Introduce LiteralExpression

timowest:
LiteralExpression extracts some cast related operations in ComparableExpressionBase to a separate class, since not all Comparable expressions are also literals.

Since most literal types are also comparable, LiteralExpression extends ComparableExpression.


timowest:
@Shredder121 @johnktims Is this ok to merge? There is no backport, since this one is not backwards compatible.

johnktims:
I wish the comment on `LiteralExpression` had a few more words, but other than that I think it's good to go.

Shredder121:
You are free to change it if you so desire.

timowest:
Yes, feel free to update it.

PR #1310, opened 2015-04-08
Fix multiplication normalization

### LINKED ISSUE #1221: Numbers not multiplied correctly in order by clause
Using QueryDSL 3.6.1 for SQL with H2 v1.3.176, I have an issue when multiplying BigDecimals (or Doubles) in an order by clause.

Given the following statement:

    .orderBy(MY_TABLE.X.multiply(a).multiply(x).asc())

with a = 1.000E5 and b = 500, 'a' being either a BigDecimal or a Double, 'b' an integer

I expect the following SQL:

    order by MY_TABLE.X * 5.000E6 asc

but instead I get:

    order by MY_TABLE.X * 1.000E2500 asc

which obviously leads to a numeric value out of range error.

johnktims:
Fixes #1221

### DIFF
--- a/querydsl-core/src/main/java/com/querydsl/core/support/Normalization.java
+++ b/querydsl-core/src/main/java/com/querydsl/core/support/Normalization.java
@@ -26,7 +26,7 @@ final class Normalization {
 
     // TODO simplify
     private static final Pattern FULL_OPERATION = Pattern.compile(
-            "(?<![\\d*/\"?' ])" + "(\\b|\\(|\\s+)"  +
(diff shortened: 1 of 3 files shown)


timowest:
maybe also take this into account https://groups.google.com/forum/#!topic/querydsl/usM9Zorm4mw

timowest:
The problem here is that we shouldn't match operations that are part of operations with higher precedence, e.g.

hours * 2 + 3 != hours * 5

but

hours * 2 * 3 = hours * 6

johnktims:
This will be handled in #1611.

Issue #1235, opened 2015-03-04
Add support for zip protocol to ClassPathUtils


ukchucktown:
When calling classLoader.getResources(packagePath) WebLogic returns "zip" as the protocol and the current code in scanPackage only handles "jar" as the protocol and will not call the scanJar method resulting in an illegal argument exception.

timowest:
Maybe this library should be considered as a replacement to the ClassPathUtils usage http://code.google.com/p/reflections/

Shredder121:
We already have it in our *test* classpath, and I personally use it in a few of our projects.
I am very fond of the way it is implemented, and the things you can do with it, but the question is 'do we want to impose this to the users of Querydsl'.

timowest:
`ClassPathUtils` is mostly used from the codegen modules and since those are not used at runtime, the extra dependency there shouldn't hurt that much.

Shredder121:
It isn't exactly the dependency that is the problem. I think the imported classpath is not quite right.
I think that at least [javassist](https://github.com/ronmamo/reflections/blob/master/pom.xml#L72) should be excluded, since it is probably 'included' [by mistake](https://github.com/ronmamo/reflections/issues/41)?
Apart from that, indeed, the extra dependency shouldn't hurt too much.

bcubk:
I'm currently working on this bug/task for contribution and I excluded the javassist dependency in the pom file. But I get following Exception:
```
java.lang.NoClassDefFoundError: javassist/bytecode/ClassFile
	at org.reflections.adapters.JavassistAdapter.getOfCreateClassObject(JavassistAdapter.java:100)
	at org.reflections.adapters.JavassistAdapter.getOfCreateClassObject(JavassistAdapter.java:24)
	at org.reflections.scanners.AbstractScanner.scan(AbstractScanner.java:30)
	at org.reflections.Reflections.scan(Reflections.java:243)
	at org.reflections.Reflections.scan(Reflections.java:203)
	at org.reflections.Reflections.<init>(Reflections.java:128)
	at com.querydsl.core.util.ClassPathUtils.scanPackage(ClassPathUtils.java:49)
	at com.querydsl.core.util.ClassPathUtils.scanPackage(ClassPathUtils.java:40)
```
It can't be a mistake.

timowest:
What do the reflections docs say on the javassist dependency? For which
functionality is it needed?

Shredder121:
[...]

PR #1312, opened 2015-04-09
Changed "scanPackage" method implementation. Used Reflections API.

bcubk:
Fixed implementation of ClassPathUtils to support 'zip' files. I used the method 'safeClassForName' to initialize the retrieved classes manually. The lib doesn't initialize the classes. 
Issue: #1235

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -85,7 +85,13 @@
       <groupId>org.reflections</groupId>
       <artifactId>reflections</artifactId>
       <version>0.9.9</version>
-      <scope>test</scope>
+      <scope>compile</scope>
+        <exclusions>
+            <exclusion>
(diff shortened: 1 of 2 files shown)


bcubk:
Sorry, I did a mistake

Shredder121:
Just wondering, if you don't mind me asking, what IDE do you use?

bcubk:
First, I was confused about the results of the Travis build. After I checked my code and found a tremendous bug. That's the reason why I closed this PR. I'm using IntelliJ IDEA.

Shredder121:
That is fine, I just noticed the change in the import order (I thought I devised a pretty 'universal'-ish one).

PR #1316, opened 2015-04-14
Simplify serializerbase

johnktims:


johnktims:
What do you think about reducing the scope of some of the helper methods like `append`, `prepend`, etc to `protected` or `private` in `SerializerBase`?

timowest:
Sounds good.

johnktims:
I couldn't reduce the visibility of `prepend`/`append`/`insert` since they're called in some of the `Templates` classes.

querydsl-core/src/test/java/com/querydsl/core/serialization/SerializerBaseTest.java
```diff
@@ -43,3 +42,26 @@ public void test() {
     }
 
 }
+
+class DummySerializer extends SerializerBase<DummySerializer> {
```
timowest: I'd prefer to have this as a top level class with package private visibility. I try to follow the one top level class per file rule.

johnktims: When I make a simple concrete class to test an abstract class, I tend to bundle it with the test so that its purpose is clear, but you have a good point so I'll move it back when I get home.

PR #1319, opened 2015-04-18
Replace deprecated class usage

timowest:
backport https://github.com/querydsl/querydsl/issues/1320


johnktims:
The only other instances of deprecated API usage that I can find outside of the tests is all the calls to `SqlTemplates#SqlTemplates(String quoteStr, char escape, boolean useQuotes)`. Do you want to correct that here?

timowest:
Not here, since that is not external API. Also it will require quite a lot of work.

johnktims:
Okay. This PR looks good to me.

Shredder121:
@johnktims yes, that is more of a reminder that the keyword quoting support is only the 'default SQL' keywords.
It is quite a lot of work to get it right for all the Templates.

Issue #1190, opened 2015-02-08
Improve javadocs of public API


timowest:
* [x] apt
* [x] codegen
* [x] collections
* [x] core
* [x] hibernate-search
* [x] jdo
* [x] jpa
* [x] jpa-codegen
* [x] lucene3
* [x] lucene4
* [x] maven-plugin
* [x] mongodb
* [x] spatial
* [x] sql
* [x] sql-codegen
* [x] sql-spatial
* [x] sql-spring

PR #1323, opened 2015-04-19
Improve javadocs for querydsl-apt

timowest:


querydsl-apt/src/main/java/com/querydsl/apt/VisitorConfig.java
```diff
@@ -14,7 +14,7 @@
 package com.querydsl.apt;
 
 /**
- * VisitorConfig defines the entity type specific visiting configuration
+ * {@link VisitorConfig} defines the entity type specific visiting configuration
```
johnktims: This sentence seems a little weird but I couldn't think of a better way to rewrite it.

querydsl-apt/src/main/java/com/querydsl/apt/SimpleTypeVisitorAdapter.java
```diff
@@ -13,19 +13,21 @@
...
 import java.lang.reflect.Method;
 import java.util.List;
 
+import javax.lang.model.type.IntersectionType;
+import javax.lang.model.type.TypeMirror;
+import javax.lang.model.util.SimpleTypeVisitor6;
+
 /**
- * Converts Java 8 IntersectionType instances into their first bound when visiting
+ * Converts Java 8 {@link IntersectionType} instances into their first bound when visiting
```
johnktims: This sentence also seems a little weird but I don't completely understand what it's describing. I'll review it again tomorrow.

Shredder121: While visiting the AST, it [flattens the IntersectionType model to just the first bound](https://github.com/querydsl/querydsl/blob/apt-javadocs/querydsl-apt/src/main/java/com/querydsl/apt/SimpleTypeVisitorAdapter.java#L46-L47).
So `Number & Comparable` would just be `Number`.

querydsl-apt/src/main/java/com/querydsl/apt/VisitorConfig.java
Shredder121:
```java
/**
* {@link VisitorConfig} defines visiting configuration specific to entity types
*/

```

Maybe throw in an `{@link }` for `EntityType`?

querydsl-apt/src/main/java/com/querydsl/apt/VisitorConfig.java
```diff
@@ -13,8 +13,10 @@
  */
 package com.querydsl.apt;
 
+import com.querydsl.codegen.EntityType;
+
 /**
- * VisitorConfig defines the entity type specific visiting configuration
+ * {@code VisitorConfig} defines the {@link EntityType}-specific visiting configuration
```
Shredder121: `{@linkplain }` reads better if you want it to directly flow in the sentence(like it does here).

timowest: But according to the docs it is mostly useful when the link text is not code. At least the JDK javadocs use `@code` quite a lot for self references.

PR #1325, opened 2015-04-20
Improve javadocs for querydsl-spatial

timowest:


querydsl-spatial/src/main/java/com/querydsl/spatial/GeometryPaths.java
```diff
@@ -25,6 +25,8 @@
 import org.geolatte.geom.Polygon;
 
 /**
+ * {@code GeometryPaths} provides factory methods for {@link GeometryExpression} creation
```
Shredder121: Shouldn't `{@link GeometryExpression}` be `{@link GeometryPath}`?
I only see `Path` subtypes.

timowest: The return types extend GeometryExpression & Path.

timowest:
Needs still some work to be fully complete.

PR #1330, opened 2015-04-26
Make cached properties transient

timowest:


timowest:
TODO the same for the spatial module

johnktims:
I've rebased onto master now.

PR #1332, opened 2015-04-27
Improve count serialization

### LINKED ISSUE #1329: Failed JPASubQuery with count() in projection (Causes parse error in OpenJPA)
I tried to get the following (stripped down) query to work:

```java
JPAQuery query = new JPAQuery(logEntityManager);
QMessageLog messageLog = QMessageLog.messageLog;
QMessageLog otherMessageLog = new QMessageLog("otherMessageLog");

List<MessageLogProjection> messageLogResult = query.
  from(messageLog).
  where(filter).
  list(
    new QMessageLogProjection(
      messageLog, 
      new JPASubQuery().from(otherMessageLog).where(otherMessageLog.originalMessageNumber.eq(messageLog.messageNumber)).count())
    );
```
```java
@Immutable
public class MessageLogProjection {
  ...
  @QueryProjection
  public MessageLogProjection(MessageLog messageLog, Long newCount) {
  ...
  }
}
```
but it causes an error in OpenJPA:
```
org.apache.openjpa.persistence.ArgumentException: An error occurred while parsing the query filter "select messageLog_, (select count(otherMessageLog) from MessageLog otherMessageLog_ where otherMessageLog_.originalMessageNumber = messageLog_.messageNumber) from MessageLog messageLog_". Error message: Attempt to query field "otherMessageLog" from non-entity variable "otherMessageLog". Perhaps you forgot to prefix the path in question with an identification variable from your FROM clause?
```

To me it seems that the entity name in **count(otherMessageLog)** is missing an underscore and should have been **count(otherMessageLog_)** instead or am I missing something?

If I replace count() in the sub query with eg unique(otherMessageLog.id) (which also results in a Long) the query works.

I'm new to QueryDSL and unsure about which query constructions that is supported... 
Is it a bug or am I doing something wrong? Can it be worked around?

Versions used was: OpenJPA 2.3.0 and QueryDSL 3.6.3

timowest:
Fixes #1329

### DIFF
--- a/querydsl-jpa/src/main/java/com/mysema/query/jpa/AbstractJPASubQuery.java
+++ b/querydsl-jpa/src/main/java/com/mysema/query/jpa/AbstractJPASubQuery.java
@@ -13,6 +13,9 @@
  */
 package com.mysema.query.jpa;
 
+import java.util.List;
+
+import com.google.common.collect.Lists;
 import com.mysema.query.DefaultQueryMetadata;
 import com.mysema.query.JoinExpression;
 import com.mysema.query.JoinType;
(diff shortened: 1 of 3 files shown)


timowest:
Is this good to merge?

johnktims:
Looks good to me.

PR #1333, opened 2015-04-28
Improve javadocs for querydsl-collections

timowest:


querydsl-collections/src/main/java/com/querydsl/collections/AbstractCollQuery.java
```diff
@@ -25,8 +25,7 @@
 import com.querydsl.core.types.*;
 
 /**
- * AbstractCollQuery provides a base class for Collection query implementations.
- *
+ * {@ode AbstractCollQuery} provides a base class for Collection query implementations. *
```
johnktims: I think you meant to remove that `*` at the end of the line. Did you also want to add `{@link Collection}`?

PR #1334, opened 2015-04-28
Improve javadocs for querydsl-core

timowest:

### DIFF
--- a/querydsl-core/src/main/java/com/querydsl/core/FetchableQuery.java
+++ b/querydsl-core/src/main/java/com/querydsl/core/FetchableQuery.java
@@ -16,34 +16,39 @@ package com.querydsl.core;
 import com.querydsl.core.types.Expression;
 
 /**
+ * {@code FetchableQuery} extends {@link Fetchable} and {@link SimpleQuery} with projection changing
+ * methods and result aggregation functionality using {@link ResultTransformer} instances.
  *
- * @param <T>
- * @param <Q>
+ * @param <T> element type
+ * @param <Q> concrete subtype
  */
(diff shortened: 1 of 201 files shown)


querydsl-core/src/main/java/com/querydsl/core/FetchableQuery.java
```diff
@@ -41,8 +45,8 @@
      * Apply the given transformer to this Projectable instance and return the results
```
Shredder121: Search+replace `Projectable` => `Fetchable` right?

timowest: right ;)

querydsl-core/src/main/java/com/querydsl/core/FetchableQuery.java
```diff
@@ -23,26 +25,28 @@
...
-     * @param exprs
-     * @return
+     * @param exprs new projection
+     * @return this
      */
     FetchableQuery<Tuple, ?> select(Expression<?>... exprs);
 
     /**
-     * Apply the given transformer to this Projectable instance and return the results
+     * Apply the given transformer to this Fetchable instance and return the results
```
Shredder121: I think this should actually be `{@code FetchableQuery}`, my bad

querydsl-core/src/main/java/com/querydsl/core/support/FetchableQueryBase.java
```diff
@@ -26,7 +26,7 @@
 import com.querydsl.core.types.SubQueryExpression;
 
 /**
- * ProjectableQuery extends the {@link QueryBase} class to provide default
+ * {@code FetchableQueryBase} extends the {@link FetchableQueryBase} class to provide default
```
Shredder121: It still extends QueryBase.

PR #1345, opened 2015-05-05
Update reference docs

timowest:


timowest:
@johnktims @Shredder121 This also needs to be merged for the 4.0 release. Sorry that it took me so long to update it.

Shredder121:
Do you feel confident to also update the Korean snippets?

timowest:
That's a little tricky, since I updated also the texts.

Shredder121:
Okay, it's your call.
I haven't really looked at *all* the snippets (specifically if you missed any), but to me it looks good.

timowest:
Well, we can't really publish the 4.0 code with the 3.* docs. I am a bit unhappy with the native query examples, but otherwise I think it's good.

PR #1351, opened 2015-05-06
Add additional MorphiaQuery constructor

### LINKED ISSUE #1347: Construct MorphiaQueries with entity type (Class<K>) instead of EntityPath<K>
Hi,

Regarding com.mysema.query.mongodb.morphia.MorphiaQuery:

I would like to suggest using a Class<K> instead of a EntityPath<K> for the construction of a MorphiaQuery. The current implementation pulls unnecessary dependencies on EntityPath and in the case I had, forces me to create an EntityPath just so the MorphiaQuery can get the type of my entity out of it.

timowest:
Fixes #1347


timowest:

Shredder121:
Yes looks good to me.
@johnktims do you want to merge when you agree?

gaspardpetit:
Much appreciated, thanks you!

PR #1353, opened 2015-05-06
Disallow order by on subqueries

### LINKED ISSUE #1346: orderBy(...) in JPASubQuery goes against JPQL specs
Both orderBy functions in JPASubQuery should be disallowed, since JPQL doesn't allow "order by" clauses in subqueries (see https://openjpa.apache.org/builds/1.2.0/apache-openjpa-1.2.0/docs/manual/jpa_langref.html#jpa_langref_subqueries ).

Posted this to (https://groups.google.com/d/msg/querydsl/Vy5KNTEj2c8/OaHFvJ33sH0J) and it was suggested I post here. Hope I posted it correctly :)

timowest:
Fixes #1346


timowest:
This is a little trickier for Querydsl 4, since there we have only a single query type.

Shredder121:
To me this variant looks good, or do you want to have a master counterpart first?

timowest:
I can provide the master counterpart later.

PR #1354, opened 2015-05-06
Improve quoting

timowest:
Fixes https://github.com/querydsl/querydsl/issues/1355


timowest:

johnktims:
Do we also need to quote `top` as mentioned in the ticket?

timowest:
I think that would be a different PR. That would be defining the reserved keywords, which is different from the general escaping logic.

johnktims:
Then I think this is ready to be merged.

PR #1357, opened 2015-05-09
Extend Checkstyle checks

timowest:


timowest:
@johnktims @Shredder121 Some of the changes are backwards incompatible, so it would be good to get it merged before 4.0.

The `ImportOrder` required for some reason also grouping of the static imports, that's why I commented it out. Also rules should be enabled again.

Shredder121:
Grouping how?

As for merging, it is indeed necessary.
I think that it's fine for this occasion to merge it before the 2 day mark, since it is only checkstyle pointing out what should change instead of us making something.

timowest:
> Grouping how?

Splitting the static imports with new line separator the same way like the other imports. It looks like a bug in checkstyle, but maybe I also misunderstood the documentation: http://checkstyle.sourceforge.net/config_imports.html

PR #1359, opened 2015-05-10
Fix APT injection issues

timowest:


timowest:
@Shredder121 @johnktims Is this good to merge? It's also a signature change.

Shredder121:
Yes, to me it looks good.

PR #1361, opened 2015-05-10
Hide constructors

timowest:


timowest:
@Shredder121 @johnktims Since this is also a signature change it would be good to merge it before 4.0. It makes ConstantImpl usage more consistent.

Shredder121:
We could for consistency's sake use `Expressions.constant()`, other than that it's good to merge.

timowest:
Yes, `ConstantImpl.create()` is more for internal use. `Expressions.constant()` is for external use, but doesn't have the optimized primitive variants though.

PR #1363, opened 2015-05-10
Fix recursion

timowest:
This should be also merged before the 4.0 release, since it changes the signature.


Shredder121:
Could you include the test case in c6acbec858284c9722dc82dbfad21783bb051501?

PR #1366, opened 2015-05-11
Extend Checkstyle checks

timowest:


johnktims:
Would it be better to move the querydsl_checks*.xml files to src/main/resources?

Also, I'd also prefer to be a little more specific with the name of the querydsl_checks.xml file. Maybe querydsl_checks_base.xml or common? Otherwise it looks like there are two main files.

timowest:
Yes, these changes make sense.

timowest:
Is this good to merge? This is so massive that I'd prefer to merge it ASAP.

Shredder121:
Yes, since it is essentially only reports from checkstyle that are fixed I think this is good to merge.

PR #1370, opened 2015-05-13
Add in/notIn for subqueries

timowest:


Shredder121:
I think adding at least a compile snippet is good.

timowest:
This doesn't need to be backported, since in Querydsl 3 subqueries are collection expressions.

querydsl-core/src/test/java/com/querydsl/core/types/ExpressionUtilsTest.java
```diff
@@ -148,6 +150,12 @@ public void In() {
     }
 
     @Test
+    public void In_SubQuery() {
+        String s = ExpressionUtils.in(str, new SubQueryExpressionImpl<String>(String.class, new DefaultQueryMetadata())).toString();
+        assertTrue(s.startsWith("str in com.querydsl.core.DefaultQueryMetadata@c"));
```
Shredder121: This is a strange toString, did you intend to make it that way?

timowest: Alternatively we can use some SQL-like syntax to serialize module independent subquery expressions.

Now it just calls just toString() on the metadata.

timowest: But maybe that change could be done in a different PR.

Shredder121:
Well, to me it looks good so far.
@johnktims what's your view?

johnktims:
Looks good to me. I'll merge it.

PR #1371, opened 2015-05-14
Enable customization of serializer config

### LINKED ISSUE #1368: GenericExporter's private final SerializerConfig serializerConfig
There is no way to customize SerializerConfig in com.mysema.query.codegen.GenericExporter.
For example I need set useEntityAccessors()

timowest:
Fixes #1368

### DIFF
--- a/querydsl-codegen/src/test/java/com/querydsl/codegen/GenericExporterTest.java
+++ b/querydsl-codegen/src/test/java/com/querydsl/codegen/GenericExporterTest.java
@@ -117,5 +117,17 @@ public class GenericExporterTest {
         exporter.export(Cat.class.getPackage());
     }
 
+    @Test
+    public void Export_SerializerConfig() {
+        exporter.setTargetFolder(new File("target/gen7"));
+        exporter.setSerializerConfig(new SimpleSerializerConfig(true, true, true, true, ""));
+        exporter.export(getClass().getPackage());
+        assertTrue(new File("target/gen7/com/querydsl/codegen/QExampleEmbeddable.java").exists());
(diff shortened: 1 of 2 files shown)


Shredder121:
To me these looks good.

querydsl-codegen/src/test/java/com/querydsl/codegen/GenericExporterTest.java
```diff
@@ -117,5 +117,11 @@ public void Export_Domain_Package() {
         exporter.export(Cat.class.getPackage());
     }
 
+    @Test
+    public void Export_SerializerConfig() {
+        exporter.setTargetFolder(new File("target/gen7"));
+        exporter.setSerializerConfig(new SimpleSerializerConfig(true, true, true, true, ""));
+        exporter.export(Cat.class.getPackage());
```
johnktims: Shouldn't there be an assertion here?

Shredder121: Do you want to introduce helper methods for that?
Getting contents of a file in Java requires a lot of boilerplate code, if you want to introduce content assertions it's good to get it right the first time.
It's then also useful for other tests to benefit from the scaffolding.

johnktims: If we don't assert anything then how would the test fail? I don't see any exceptions that would be triggered. I can work on easing the process if you think that would help.

Shredder121: At least, since it's `codegen`, if the `code` is `gen`-ed it's a step in the right direction.
But assertions on its content is also good.

timowest:
I added now some assertions.

PR #1374, opened 2015-05-14
Add connection closing functionality via listener

### LINKED ISSUE #1254: Connection pooling does not appear to be correctly supported
Using database connection pooling with `org.apache.commons.dbcp2.BasicDataSource`, upon executing a `SQLQuery` the connection is never returned to the pool, unless you put it inside a transaction, which is not always convenient. The following workaround manually closes the connection, returning it to the pool.

    protected Connection getSQLQueryConnection(@NonNull SQLQuery query) {
        Field declaredField;
        try {
            declaredField = SQLQuery.class.getSuperclass().getDeclaredField("conn");
            declaredField.setAccessible(true);
            return (Connection) declaredField.get(query);
        } catch (NoSuchFieldException | SecurityException | IllegalArgumentException | IllegalAccessException e) {
            log.error("Could not retrieve connection object from Object {}", query);
        }

        return null;
    }

    protected void closeSQLQuery(@NonNull SQLQuery query) {
        try {
            if (!TransactionSynchronizationManager.isActualTransactionActive()) {
                getSQLQueryConnection(query).close();
            } else {
                log.warn("Not closing connection on transaction wrapped connection {}", query);
            }
        } catch (SQLException e) {
            log.error("Could not close connection {}", query, e);
        }
    }

The most notable drawback with this mechanism is the fact that it relies on the programmer knowing that this *must* be called, otherwise the connection pool will soon run out of available connections.

It would be far better if querydsl could provide support for connection pooling, ensuring that connections are returned to the pool upon execution finishing.

timowest:
Fixes #1254

### DIFF
--- a/querydsl-sql/src/main/java/com/querydsl/sql/AbstractSQLQuery.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/AbstractSQLQuery.java
@@ -268,21 +268,21 @@ public abstract class AbstractSQLQuery<T, Q extends AbstractSQLQuery<T, Q>> exte
             listeners.executed(context);
 
             if (expr == null) {
-                return new SQLResultIterator<T>(configuration, stmt, rs) {
+                return new SQLResultIterator<T>(configuration, stmt, rs, listeners, context) {
                     @Override
                     public T produceNext(ResultSet rs) throws Exception {
                         return (T) rs.getObject(1);
                     }
                 };
             } else if (expr instanceof FactoryExpression) {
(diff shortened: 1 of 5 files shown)


timowest:
I fixed the remaining issue by introducing a parent child relationship between the contexts. I made the key constant protected since I wasn't sure where to put the key for now. What do you think?

Shredder121:
I think this now looks good, @johnktims what do you think?

johnktims:
I think this looks okay. I'll merge it.

PR #1378, opened 2015-05-15
Doc update: Remove querydsl-apt from regular classpath

SvenBunge:
When a dependency is set to 'provided' the classes are available on classpath during development. Even worse: The compiler is compiling the code with success and the implementation will fail due runtime.

The solution for that issue is (I think): Add the dependency to the maven plugin itself. So it's only available due runtime of the plugin and won't be added to the classpath of the programm itself.

### DIFF
--- a/querydsl-jpa/README.md
+++ b/querydsl-jpa/README.md
@@ -9,13 +9,6 @@ The JPA module provides integration with the JPA 2 persistence API.
 ```XML
 <dependency>
   <groupId>com.querydsl</groupId>
-  <artifactId>querydsl-apt</artifactId>
-  <version>${querydsl.version}</version>
-  <scope>provided</scope>
-</dependency>    
-    
-<dependency>
-  <groupId>com.querydsl</groupId>
   <artifactId>querydsl-jpa</artifactId>
   <version>${querydsl.version}</version>
 </dependency>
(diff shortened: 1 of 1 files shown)


timowest:
Thanks for the contribution. The APT processor uses the project compile classpath for compilation. Plugin dependencies won't be part of that classpath. Did you try that configuration?

I agree that there is the risk that APT dependencies might be referenced also elsewhere. Did that happen in your case?

SvenBunge:
For sure: I've tried it in an example project and it works very well. The Q-Files are generated and they work in all my test cases.
As far as I know the dependency added by a plugin is only on the classpath when the lifecycle runs the plugin goal. See Sonatype documentation 7.1.2: http://books.sonatype.com/mvnref-book/reference/_configuring_maven_plugins.html

I removed slf4j-log4j as well because it clashes with my logback on the CP:
```
SLF4J: Found binding in [jar:file:/<private>/.m2/repository/ch/qos/logback/logback-classic/1.1.3/logback-classic-1.1.3.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/<private>/.m2/repository/org/slf4j/slf4j-log4j12/1.7.11/slf4j-log4j12-1.7.11.jar!/org/slf4j/impl/StaticLoggerBinder.class]
```
Are you sure this dependency is really needed? I'm new to QueryDSL but if it uses slf4j for all logstatements, it is ok without slf4j-log4j. If you use log4j as well, you should replace the entry with log4j-over-slf4j (http://www.slf4j.org/legacy.html).

johnktims:
I've been using this configuration without issues: https://github.com/johnktims/spring-boot-querydsl-liquibase-starter/blob/master/pom.xml#L81

timowest:

johnktims:
:+1:

Shredder121:
Yes, it's good, @johnktims do you want to merge?

PR #1382, opened 2015-05-19
Improve QBean alias handling

### LINKED ISSUE #1379: Projections.bean discards aliases required for sorting
```java
final List<Expression<?>> list = new ArrayList<>();
final NumberPath<Double> avgAmountPath = Expressions.numberPath(Double.class, "avgAmount");
list.add(QEntity.entity.amount.avg().as(avgAmountPath));
jpaQuery.groupBy(QEntity.entity.something);
jpaQuery.orderBy(avgAmountPath.asc());

// this throws exception: Caused by: com.mysql.jdbc.exceptions.jdbc4.MySQLSyntaxErrorException: Unknown column 'avgAmount' in 'order clause'
jpaQuery.list(Projections.bean(MyResult.class, list.toArray(new Expression[0])));

// this is ok
jpaQuery.list(list.toArray(new Expression[0]));
```

timowest:
Fixes #1379

### DIFF
--- a/querydsl-core/src/main/java/com/querydsl/core/types/QBean.java
+++ b/querydsl-core/src/main/java/com/querydsl/core/types/QBean.java
@@ -27,6 +27,7 @@ import java.util.Map;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.primitives.Primitives;
+import com.querydsl.core.group.GroupExpression;
 
 /**
  * {@code QBean} is a JavaBean populating projection type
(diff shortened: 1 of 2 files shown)


querydsl-core/src/main/java/com/querydsl/core/types/QBean.java
```diff
@@ -73,6 +78,10 @@
         return rv.build();
     }
 
+    private static boolean isAliasSplitted(Expression<?> expr) {
```
johnktims: I don't understand why an expression that is an instance of `FactoryExpression` or  `GroupExpression` is considered a split alias. I think the notion of splitting is confusing me.

timowest: FactoryExpressions and GroupExpressions are compound expressions which are serialized as individual items, so you can't really alias them on the SQL/JPQL level.

The change introduced in this PR is to serialize the alias into the query, if the left hand side of the alias operation is not a FactoryExpression or GroupExpression.

timowest: Shall I rename the method?

timowest:
@johnktims @Shredder121 Is this ok to merge?

querydsl-core/src/main/java/com/querydsl/core/types/QBean.java
johnktims: Sorry. I forgot to respond to your question. I would rename the method. `isCompoundExpression`?

timowest: Sounds good, I'll do that.

timowest: Changed the name.

johnktims:
This PR looks good to me. @Shredder121 ?

Shredder121:
Yes, it looks good now.
For the sake of completeness I'd say let the build finish, and then I'll merge it.
Or you can do it as well.

PR #1386, opened 2015-05-20
Fix serialization issues

### LINKED ISSUE #1383: Serialisation
Hi,

Since few versions, it's not possible to serialize some objects like Predicate (and maybe OrderSpecicier).

I want to use the 3.6.3.

I'm using it to directly transfer restrictions between my FRONT and my BACK offices.

The class with problem is at least : com.mysema.query.types.path.PathBuilderValidator.
The 3 final anonymous classes inside aren't Serializable.

Thanks

timowest:
Fixes #1383

### DIFF
--- a/querydsl-core/src/main/java/com/querydsl/core/types/FactoryExpressionBase.java
+++ b/querydsl-core/src/main/java/com/querydsl/core/types/FactoryExpressionBase.java
@@ -82,4 +82,16 @@ public abstract class FactoryExpressionBase<T> extends ExpressionBase<T> impleme
         return new FactoryExpressionWrapper<T>(this);
     }
 
+    @Override
+    public boolean equals(Object o) {
+        if (o == this) {
+            return true;
+        } else if (o instanceof FactoryExpression) {
+            return getClass().equals(o.getClass())
+                    && getArgs().equals(((FactoryExpression) o).getArgs());
+        } else {
+            return false;
+        }
+    }
+
 }
--- a/querydsl-core/pom.xml
+++ b/querydsl-core/pom.xml
@@ -72,7 +72,7 @@
       <artifactId>jdepend</artifactId>
       <version>2.9.1</version>
       <scope>test</scope>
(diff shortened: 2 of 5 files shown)


timowest:
I fixed the duplicate reflections dependency issue

johnktims:
This looks fine to me. What do you think, @Shredder121 ?

querydsl-core/src/main/java/com/querydsl/core/types/FactoryExpressionBase.java
```diff
@@ -82,4 +82,15 @@ public FactoryExpressionBase(Class<? extends T> type) {
         return new FactoryExpressionWrapper<T>(this);
     }
 
+    @Override
+    public boolean equals(Object o) {
+        if (o == this) {
+            return true;
+        } else if (o instanceof FactoryExpression) {
+            return getClass().equals(o.getClass()) && getArgs().equals(((FactoryExpression) o).getArgs());
```
Shredder121: Does it sound reasonable to wrap this line before the `&&`?

Shredder121:
To me it looks good as well, I will wait for answer to [this comment](https://github.com/querydsl/querydsl/pull/1386#discussion_r31389468) and I'll merge it.

querydsl-core/src/main/java/com/querydsl/core/types/FactoryExpressionBase.java
timowest: Yes, that could be done.

Shredder121: Shall I do that then?

timowest: Ok. I'll continue working in the evening.

Shredder121:
I updated both PRs.
@johnktims I think when the build finishes you can merge it if you agree.

PR #1387, opened 2015-05-21
Fix serialization issues

### LINKED ISSUE #1383: Serialisation

timowest:
Fixes #1383 
backport of https://github.com/querydsl/querydsl/pull/1386

### DIFF
--- a/querydsl-core/pom.xml
+++ b/querydsl-core/pom.xml
@@ -72,7 +72,7 @@
       <artifactId>jdepend</artifactId>
       <version>2.9.1</version>
       <scope>test</scope>
-    </dependency>    
+    </dependency>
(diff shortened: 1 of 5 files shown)


Shredder121:
@johnktims this is the backport variant.

johnktims:
I was waiting for Travis to finish and then I forgot about it.

PR #1390, opened 2015-05-21
Add more Expressions.arrayPath variants

### LINKED ISSUE #1384: Add more Expressions.arrayPath variants

timowest:
Fixes #1384

### DIFF
--- a/querydsl-core/src/test/java/com/querydsl/core/types/dsl/ExpressionsTest.java
+++ b/querydsl-core/src/test/java/com/querydsl/core/types/dsl/ExpressionsTest.java
@@ -57,6 +57,11 @@ public class ExpressionsTest {
                 assertReturnType(Expressions.class.getMethod(type + "Template", Class.class, Template.class, ImmutableList.class));
             }
         }
+
+        // arrays
+        assertReturnType(Expressions.class.getMethod("arrayPath", Class.class, String.class));
+        assertReturnType(Expressions.class.getMethod("arrayPath", Class.class, Path.class, String.class));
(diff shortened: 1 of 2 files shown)


johnktims:
This looks okay. I worry that `Expressions` is getting too large but we can discuss that elsewhere.

querydsl-core/src/test/java/com/querydsl/core/types/dsl/ExpressionsTest.java
```diff
@@ -57,6 +57,11 @@ public void  Signature() throws NoSuchMethodException {
                 assertReturnType(Expressions.class.getMethod(type + "Template", Class.class, Template.class, ImmutableList.class));
             }
         }
+
+        // arrays
+        assertReturnType(Expressions.class.getMethod("arrayPath", Class.class, String.class));
+        assertReturnType(Expressions.class.getMethod("arrayPath", Class.class, Path.class, String.class));
+        assertReturnType(Expressions.class.getMethod("arrayPath", Class.class, PathMetadata.class));
```
johnktims: Is there a reason to make sure the return type is the same as the name of the method? It just seems like an odd thing to test.

timowest: It's a test for the naming convention, the return type simple name is the capitalized method name.

timowest:
@johnktims @Shredder121 Is this ok to merge?

johnktims:
:+1:

Shredder121:
I don't know whether bounded wildcards would be more relaxing, but it looks good.
Maybe we could add a few methods that use them later?
But this PR is good to merge.

PR #1392, opened 2015-05-24
Add more queryFactory constructors

timowest:


querydsl-jpa/src/main/java/com/querydsl/jpa/hibernate/HibernateQueryFactory.java
```diff
@@ -37,6 +37,20 @@
...
+    }
+
+    public HibernateQueryFactory(JPQLTemplates templates, final Session session) {
+        this.session = new Provider<Session>() {
+            @Override
+            public Session get() {
+                return session;
+            }
+        };
+        this.templates = templates;
```
johnktims: You could also create the `Provider<Session>` and pass it to the existing constructor `this(templates, new Provider<Session>...);`.

timowest: Syntactically it feels cleaner this way.

johnktims: That's fine with me. I think this PR is ready to be merged then. @Shredder121 ?

Shredder121: Yes, syntactically it is much better this way.
I think it's good to merge, I'll do so.

PR #1394, opened 2015-05-29
Fix Scala case class generation dropping parameters

mdmoss:
Mapping on a set produced Parameters with hashCode collisions, which were then dropped from the resulting set. Converting to a Vector before mapping prevents this.


timowest:
I made a PR for the original issue https://github.com/querydsl/codegen/pull/26

johnktims:
@mdmoss Thank you for the contribution but I'm going to close this PR in favor of querydsl/codegen#26 .

mdmoss:
Awesome, that works for me. Thanks all.

PR #1399, opened 2015-05-31
JSR310 Date Time support

Shredder121:
I have edited some commits from #1209, but there are still some support issues regarding time zones.


Shredder121:
The finishing touches of Timo look good.

johnktims:
Agreed.

PR #1404, opened 2015-06-01
Bump codegen version

timowest:


johnktims:
I verified that codegen's version was incremented but I can't tell if anything else was changed in the file since it was reformatted. If that's all that changed, this looks fine to me.

timowest:
I'll make a cleaner commit instead.

johnktims:
That's better. Thanks.

It looks good to me.

Shredder121:
What's the deal with changing the last lines everytime, what is intellij set to?
Other than that, it looks good to me as well.

timowest:
I am not sure which setting causes it.

PR #1408, opened 2015-06-09
Make BeanPath.casts threadsafe

### LINKED ISSUE #1403: BeanPath.casts should be made safe for concurrent usage
BeanPath.casts should be made safe for concurrent usage

timowest:
Fixes #1403


Shredder121:
I think these both look good.

johnktims:
Looks good to me.

PR #1412, opened 2015-06-19
Add more assertions

timowest:


Shredder121:
To me it looks good.
@johnktims what about you?

johnktims:
If you search for "assertions" there are several comments that say "// TODO : assertions". Also, searching for "System.out.print" reveals a few more places where assertions could be made.

timowest:
I'll try to add more assertions for the compilation cases.

johnktims:
There are some more assertions that could be added but I think this is good enough for now.

Shredder121:
Agreed.

PR #1415, opened 2015-06-23
Improve Union interface

### LINKED ISSUE #1414: Rename Union.list() to fetch
For backwards compatibility list() should be deprecated. 

Union should probably also extend Fetchable.

timowest:
Fixes #1414

### DIFF
--- a/querydsl-sql/src/main/java/com/querydsl/sql/Union.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/Union.java
@@ -16,6 +16,7 @@ package com.querydsl.sql;
 import java.util.List;
 
 import com.mysema.commons.lang.CloseableIterator;
+import com.querydsl.core.Fetchable;
 import com.querydsl.core.types.*;
 
 /**
@@ -25,13 +26,14 @@ import com.querydsl.core.types.*;
  *
  * @param <RT> return type of projection
  */
-public interface Union<RT> extends SubQueryExpression<RT> {
+public interface Union<RT> extends SubQueryExpression<RT>, Fetchable<RT> {
 
     /**
      * Get the projection as a typed List
      *
-     * @return results
+     * @deprecated Use {@link Union#fetch()}
      */
+    @Deprecated
     List<RT> list();
 
     /**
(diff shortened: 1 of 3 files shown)


querydsl-sql/src/main/java/com/querydsl/sql/Union.java
```diff
@@ -25,23 +25,15 @@
...
  * @param <RT> return type of projection
  */
-public interface Union<RT> extends SubQueryExpression<RT> {
+public interface Union<RT> extends SubQueryExpression<RT>, Fetchable<RT> {
 
     /**
-     * Get the projection as a typed List
-     *
-     * @return results
+     * Use fetch instead
```
johnktims: You might want to change this to `@deprecated Use {@link Union#fetch()}` so people can easily navigate to the new method.

querydsl-sql/src/test/java/com/querydsl/sql/UnionBase.java
```diff
@@ -40,7 +40,7 @@ public void In_Union() {
     public void Union() throws SQLException {
         SubQueryExpression<Integer> sq1 = query().from(employee).select(employee.id.max());
         SubQueryExpression<Integer> sq2 = query().from(employee).select(employee.id.min());
-        List<Integer> list = query().union(sq1, sq2).list();
+        List<Integer> list = query().union(sq1, sq2).fetch();
```
johnktims: Since `list()` is still available do we need to keep the tests for it?

querydsl-sql/src/main/java/com/querydsl/sql/Union.java
```diff
@@ -25,23 +25,17 @@
...
      */
+    @Deprecated
     List<RT> list();
 
     /**
-     * Get the projection as a typed Iterator
-     *
-     * @return result iterator
-     */
-    CloseableIterator<RT> iterate();
```
Shredder121: I would keep this method here, because I am not sure if it is binary compatible.
It also doesn't hurt that much.

Shredder121: Is there a reason that the whitespace changed?

Shredder121:
This looks now good to me.

PR #1421, opened 2015-06-26
Replace tabs with spaces

timowest:


Shredder121:
Did you check other files for potential tabs?

timowest:
I extended this now.

querydsl-apt/src/main/hibernate.xml
```diff
@@ -7,13 +7,12 @@
   </formats>
   <includeBaseDirectory>false</includeBaseDirectory>
   <fileSets>
+    <fileSet>      <directory>src/apt/hibernate</directory>
```
Shredder121: What happened here?

Shredder121:
As for the contents of this PR, it's not a feature that has to be reviewed so merging it early is fine with me.

johnktims:
It looks like you found all the tabs so I'll merge this.

PR #1422, opened 2015-06-28
Consolidate examples

johnktims:
This PR moves querydsl-customer-dao, querydsl-sql-example, and querydsl-jpa-example under querydsl/querydsl/querydsl-examples.


johnktims:
I need to upgrade querydsl-sql-example and querydsl-jpa-example to use querydsl 4.0.1 and add the examples module to the travis profile.

timowest:
This looks good to me.

querydsl-examples/querydsl-example-jpa-guice/pom.xml
```diff
@@ -0,0 +1,210 @@
...
+  <packaging>jar</packaging>
+  <name>Querydsl example - JPA Guice</name>
+
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <failIfNoTests>false</failIfNoTests>
+
+    <hibernate.version>4.1.7.Final</hibernate.version>
+    <hibernate.validator.version>4.3.0.Final</hibernate.validator.version>
+    <querydsl.version>4.0.1</querydsl.version>
```
Shredder121: I would just remove all mentionings of `querydsl.version`.

Shredder121:
So does this mean that the original repositories can be removed?
Should we add a new label for `example` or something?

johnktims:
I think we should remove the original repositories.

I'm not sure a new label is necessary but I guess it wouldn't hurt to create one.

timowest:
I agree, we can remove the original repositories.

Shredder121:
Then I'll merge this.

PR #1423, opened 2015-06-30
Release 4.0.2

timowest:


pom.xml
```diff
@@ -239,6 +239,7 @@
                 <requireBackwardCompatibility implementation="org.semver.enforcer.RequireBackwardCompatibility">
                   <excludes>
                     <exclude>com/querydsl/core/types/dsl/PathBuilderValidator</exclude>
+                    <exclude>com/querydsl/sql/Union</exclude>
```
Shredder121: Wasn't it backwards compatible? What did the enforcer say?

timowest: The enforcer complained about the Union interface change. A superinterface was added.

Shredder121: http://docs.oracle.com/javase/specs/jls/se7/html/jls-13.html Chapter 13, last item of the bulleted list suggests it _should_ be a compatible change.

PR #1428, opened 2015-07-06
Fix alias serialization for MongoDB

### LINKED ISSUE #1427: Mongodb: cast with "as" in query produces unexpected result
I'm trying to cast a class in a query on mongodb, but it doesn't work as I would expect.

I wrote some unit tests to illustrate, you can see it here: https://github.com/jotak/querydsl/commit/28756b91c5a3494130f684e19c4266f1bb5300aa

2 of my tests fail:
- "countCatsWithInstanceOf": quite unrelated, but worth mentioning: "instanceOf" doesn't seem to be supported
- "countFishFromNameAndBreedWithCast" : this illustrates the bug. When running
      **.and(QFood.food.as(QFish.class).breed.eq("unknown"))**

the json produced is

    { "name" : "f1" , "food.breed" : "unknown"}

instead of expected:

    { "name" : "f1" , "breed" : "unknown"}

So it returns 0 result

For information, this ticket follows what I wrote on discussion board: https://groups.google.com/forum/#!topic/querydsl/KaUWIUwov6w

jotak:
Fixes #1427 + add junit tests:

When querying for **QFood.food.as(QFish.class).breed.eq("unknown")**

I expect to see **{ "name" : "f1" , "breed" : "unknown"}**

rather than **{ "name" : "f1" , "food.breed" : "unknown"}**

Note this is my first commit on querydsl, so I hope I have a correct understanding of what "PathType.DELEGATE" is and how it should be interpreted.

### DIFF
--- a/querydsl-mongodb/src/main/java/com/querydsl/mongodb/MongodbSerializer.java
+++ b/querydsl-mongodb/src/main/java/com/querydsl/mongodb/MongodbSerializer.java
@@ -326,7 +326,8 @@ public abstract class MongodbSerializer implements Visitor<Object, Void> {
         if (metadata.getParent() != null) {
             if (metadata.getPathType() == PathType.COLLECTION_ANY) {
                 return visit(metadata.getParent(), context);
-            } else if (metadata.getParent().getMetadata().getPathType() != PathType.VARIABLE) {
+            } else if (metadata.getParent().getMetadata().getPathType() != PathType.VARIABLE
+                    && metadata.getParent().getMetadata().getPathType() != PathType.DELEGATE) {
                 String rv = getKeyForPath(expr, metadata);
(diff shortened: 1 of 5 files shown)


timowest:
Thanks a lot for the PR! In pull request commit messages it's not necessary to refer to the issue. 

But the PR description should contain `Fixes #1427`, and the PR title could be something like `Fix alias serialization for MongoDB`.

jotak:
Ok I renamed the PR

timowest:
@johnktims @Shredder121 This looks ok to me. How about you?

Shredder121:
Yes, looks good to me as well.

johnktims:
Looks fine to me. Thanks for the PR, @jotak !

jotak:
Thanks for merging! sure i'll open another PR

Issue #1431, opened 2015-07-10
Weird Behavior for byte[] array in Hibernate Objects


voor:
Attempting to utilize a class that has a field mapping byte[] to Blob.  Getting the following error:

```
[B cannot be cast to [Ljava.lang.Object;
! java.lang.ClassCastException: [B cannot be cast to [Ljava.lang.Object;
! at com.querydsl.jpa.hibernate.HibernateUtil.setValue(HibernateUtil.java:71) ~[querydsl-jpa-4.0.2.jar:na]
! at com.querydsl.jpa.hibernate.HibernateUtil.setConstants(HibernateUtil.java:63) ~[querydsl-jpa-4.0.2.jar:na]
! at com.querydsl.jpa.hibernate.AbstractHibernateQuery.createQuery(AbstractHibernateQuery.java:105) ~[querydsl-jpa-4.0.2.jar:na]
! at com.querydsl.jpa.hibernate.AbstractHibernateQuery.createQuery(AbstractHibernateQuery.java:97) ~[querydsl-jpa-4.0.2.jar:na]
! at com.querydsl.jpa.hibernate.AbstractHibernateQuery.fetch(AbstractHibernateQuery.java:174) ~[querydsl-jpa-4.0.2.jar:na]
```

voor:
Looks like it's because a byte array is considered an array.

It's attempting to cast the array to an Object array, which you can't do with a byte array.

```java
    private static void setValue(Query query, String key, Object val) {
        if (val instanceof Collection<?>) {
            query.setParameterList(key, (Collection<?>) val);
        } else if (val.getClass().isArray()) {
            query.setParameterList(key, (Object[]) val);
        } else if (TYPES.containsKey(val.getClass())) {
            query.setParameter(key, val, TYPES.get(val.getClass()));
        } else {
            query.setParameter(key, val);
        }
    }
```

Would this possibly fix it?  (Or something similar)

```java
    private static void setValue(Query query, String key, Object val) {
        if (val instanceof Collection<?>) {
            query.setParameterList(key, (Collection<?>) val);
        } else if (val.getClass().equals(byte[].class)) {
            ...
        } else if (val.getClass().isArray()) {
            query.setParameterList(key, (Object[]) val);
        } else if (TYPES.containsKey(val.getClass())) {
            query.setParameter(key, val, TYPES.get(val.getClass()));
        } else {
            query.setParameter(key, val);
        }
    }
```

Shredder121:
[...]

PR #1432, opened 2015-07-11
Simple fix to resolve byte arrays being slammed into array groupings.

voor:
Very basic pull request regarding #1431


timowest:
Thanks for PR! I submitted this one instead https://github.com/querydsl/querydsl/pull/1440 which should solve the problem in a more general fashion.

PR #1442, opened 2015-07-18
Add a strategy to allow customize the default variable name case

marcellor:
Proposed feature: Add a strategy to allow the user customize the default variable case generation.
1. Added an interface and a default implementation with implements the current uncapSimpleName  behavior.
2. Injected a new String parameter to the EntitySerializer constructor, which defines the strategy implementation class to use.
3. Bounded the new String parameter to the Configuration.

### DIFF
--- a/querydsl-codegen/src/main/java/com/querydsl/codegen/EntitySerializer.java
+++ b/querydsl-codegen/src/main/java/com/querydsl/codegen/EntitySerializer.java
@@ -52,6 +52,8 @@ public class EntitySerializer implements Serializer {
 
     protected final Collection<String> keywords;
 
+    private CaseTransformer caseTransformer;
+
     /**
      * Create a new {@code EntitySerializer} instance
      *
@@ -59,9 +61,14 @@ public class EntitySerializer implements Serializer {
      * @param keywords keywords to be used
      */
     @Inject
-    public EntitySerializer(TypeMappings mappings, @Named("keywords") Collection<String> keywords) {
+    public EntitySerializer(TypeMappings mappings, @Named("keywords") Collection<String> keywords, @Named(CodegenModule.CASE_TRANSFORMER_CLASS) String caseTransfomerClass) {
         this.typeMappings = mappings;
         this.keywords = keywords;
+        try {
+           this.caseTransformer = (CaseTransformer) Class.forName(caseTransfomerClass).newInstance();
+        } catch (Exception e) {
+          this.caseTransformer = new UncapitalizedCaseTransformer();
+        }
     }
 
     protected void constructors(EntityType model, SerializerConfig config,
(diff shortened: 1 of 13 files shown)


querydsl-codegen/src/main/java/com/querydsl/codegen/UncapitalizedCaseTransformer.java
```diff
@@ -0,0 +1,19 @@
...
+ * Uncapitalizes the first letter of a name.
+ *
+ */
+public class UncapitalizedCaseTransformer implements CaseTransformer {
+
+    /**
+     * @see com.querydsl.codegen.CaseTransformer#transform(java.lang.String)
+     */
+    @Override
+    public String transform(String name) {
```
Shredder121: I think it's better to name this class 'DefaultVariableNameFunction' and make it a Guava Function<EntityType, String>.
This makes the call from EntitySerializer a bit more natural and doesn't limit the Function to just String manipulation.
What do you think?

marcellor: Yes, that looks good.
What about the instatiation of the variableNameFunction? Would you inject the Function in the constructor?

marcellor: Hi Ruben,

As per my last comment on the pull request, do you think the rest of the
code looks ok?
Should we inject the strategy class name in order to allow the user plug in
its own implementation?

Thanks,

Marcello

querydsl-codegen/src/main/java/com/querydsl/codegen/EntitySerializer.java
```diff
@@ -59,9 +61,14 @@
      * @param keywords keywords to be used
      */
     @Inject
-    public EntitySerializer(TypeMappings mappings, @Named("keywords") Collection<String> keywords) {
+    public EntitySerializer(TypeMappings mappings, @Named("keywords") Collection<String> keywords, @Named(CodegenModule.CASE_TRANSFORMER_CLASS) String caseTransfomerClass) {
         this.typeMappings = mappings;
         this.keywords = keywords;
+        try {
+           this.caseTransformer = (CaseTransformer) Class.forName(caseTransfomerClass).newInstance();
+        } catch (Exception e) {
```
Shredder121: What do you think about this broad catch clause?
Do you maybe think it can be split up?
I think at least a catch clause for RuntimeException (and rethrowing it) would be better.

marcellor: Yes, that would be safer.

timowest:
@marcellor Could you take a look at the build failures? There seems to be a javadoc checkstyle rule violation.

marcellor:
The uncapSimpleName of Public domain is generated as "public", which is not allowed as identifier:

[ERROR] COMPILATION ERROR : 
[ERROR] /home/travis/build/querydsl/querydsl/querydsl-apt/target/generated-test-sources/java/com/querydsl/apt/domain/QPublic.java:[20,32] <identifier> expected
...

marcellor:
The checkstyle rule violation is solved, now a compilation problem in https://github.com/querydsl/querydsl/pull/1458

timowest:
@marcellor Still the checkstyle issue

```
src/main/java/com/querydsl/codegen/UncapitalizedCaseTransformer.java[11] (javadoc) JavadocStyle: Javadoc has empty description section.
```

marcellor:
In my last comment I was referring to the new build errors in pull request https://github.com/querydsl/querydsl/pull/1458:
https://travis-ci.org/querydsl/querydsl/builds/72634623
Can we close the previous request (1442) ?
Thanks

PR #1444, opened 2015-07-20
Skip the enforcer plugin in the examples

Shredder121:


johnktims:
Was the enforcer causing problems?

Shredder121:
Well, the enforcer is set to backwards compatibility from the root project.
This means that the public API (public methods) cannot just change.
This is quite unnecessary in a example projects, so I figured from `querydsl-examples` and down the enforcer can be skipped.

So yes, it was causing problems.

johnktims:
Sounds good. @timowest can merge it if he agrees.

PR #1447, opened 2015-07-21
Fix PostgreSQL numerics handling

### LINKED ISSUE #1425: postgresql: numeric mapped to BigInteger
Hi!

"without any precision or scale creates a column in which numeric values of any precision and scale can be stored, up to the implementation limit on precision."
http://www.postgresql.org/docs/9.0/static/datatype-numeric.html

However querydsl (4.0.2) maps it to BigInteger instead of BigDecimal (probably the <= 0 in JDBCTypeMapping.getNumericClass()).

timowest:
Fixes #1425


robertandrewbain:
Looks good to me.

PR #1448, opened 2015-07-21
Add SELECT FOR UPDATE support

### LINKED ISSUE #1446: sqlQuery.forUpdate() does not work on SQLServer
In case we have an SQLQuery and call query.forUpdate(), that does not work on SQLServer.

The reason is that while all other databases use the syntax:

    SELECT... FROM...WHERE... FOR UPDATE

SQLServer uses the following:

   SELECT ... FROM ... WITH (UPDLOCK) ... WHERE...

That means that pessimistic lock can be done with a Table Hint in SQLServer.

There is a class called SQLServerQuery and the hint can be added easily there. However, by using that class, the code will be database dependent.

It would be nice if sqlQuery.forUpdate() added a table hint instead of QueryFlag in case of SQLServer so the client code could be database independent.

timowest:
Fixes #1446

### DIFF
--- a/querydsl-sql/src/main/java/com/querydsl/sql/AbstractSQLQuery.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/AbstractSQLQuery.java
@@ -128,7 +128,9 @@ public abstract class AbstractSQLQuery<T, Q extends AbstractSQLQuery<T, Q>> exte
      * @return the current object
      */
     public Q forUpdate() {
-        return addFlag(SQLOps.FOR_UPDATE_FLAG);
+        QueryFlag forUpdateFlag = configuration.getTemplates().getForUpdateFlag();
+        return addFlag(forUpdateFlag);
+
     }
 
     @Override
--- a/querydsl-sql/src/main/java/com/querydsl/sql/SQLOps.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/SQLOps.java
@@ -88,12 +88,13 @@ public enum SQLOps implements Operator {
         return type;
     }
 
+    @Deprecated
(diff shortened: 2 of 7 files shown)


balazs-zsoldos:
That was fast, thank you!

The problem is this modification only works if there is no WHERE clause in the SQL statement. WITH must come in SQLServer after FROM or the last JOIN.

E.g.:

    SELECT a.x, b.x  FROM a INNER JOIN b ON a.x = b.x __WITH (UPDLOCK)__ WHERE a.y = ?

The generated SQL based on the modification throws a syntax error:

    SELECT a.x, b.x  FROM a INNER JOIN b ON a.x = b.x  WHERE a.y = ? __WITH (UPDLOCK)__

It is very common that there is at least one predicate in WHERE clause to filter the results when someone wants to use pessimistic locking. I suggest that the Unit tests should contain a where part as well. E.g.:

    query().from(survey).forUpdate().select(survey.id).where(suervey.gt(-1)).fetchOne();

In my code, I used the following line in case of SQLServer instead of _.forUpdate()_ (after analyzing SQLQuery class):

    query.addJoinFlag(" with (" + SQLServerTableHints.UPDLOCK.name() + ")", Position.END);

I cannot see how this could be added by only changing the SQLTemplates class yet.

timowest:
The position of the flag is now also customizable and is fixed for MSSQL.

balazs-zsoldos:
That was fast again!!! I guess line 86 and 87 in SQLServerTemplates is not necessary anymore:

     add(SQLOps.FOR_UPDATE, "\nwith (updlock)");

timowest:
@balazs-zsoldos We need both the customization of the position and the template of the operation.

johnktims:
This looks good to me.

balazs-zsoldos:
For me, too. Waiting for the new release ;-)

Shredder121:
Yes looks good, I'll merge this.

PR #1449, opened 2015-07-21
Fix tests

timowest:


johnktims:
You ran the oracle tests locally and saw this one was broken?

timowest:
yes

johnktims:
I was having issues with my oracle setup and wasn't able to run the JPA suite for some reason so I'll have to take your word for those tests.

LGTM.

PR #1451, opened 2015-07-22
Fix alias serialization for MongoDB

### LINKED ISSUE #1427: Mongodb: cast with "as" in query produces unexpected result

jotak:
Forward PR #1428 to branch QUERYDSL_3

----

Fixes #1427 + add junit tests:

When querying for QFood.food.as(QFish.class).breed.eq("unknown")

I expect to see { "name" : "f1" , "breed" : "unknown"}

rather than { "name" : "f1" , "food.breed" : "unknown"}

Note this is my first commit on querydsl, so I hope I have a correct understanding of what "PathType.DELEGATE" is and how it should be interpreted.

### DIFF
--- a/querydsl-mongodb/src/main/java/com/mysema/query/mongodb/MongodbSerializer.java
+++ b/querydsl-mongodb/src/main/java/com/mysema/query/mongodb/MongodbSerializer.java
@@ -330,7 +330,8 @@ public abstract class MongodbSerializer implements Visitor<Object, Void> {
         if (metadata.getParent() != null) {
             if (metadata.getPathType() == PathType.COLLECTION_ANY) {
                 return visit(metadata.getParent(), context);
-            } else if (metadata.getParent().getMetadata().getPathType() != PathType.VARIABLE) {
+            } else if (metadata.getParent().getMetadata().getPathType() != PathType.VARIABLE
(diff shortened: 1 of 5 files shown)


Shredder121:
Thank you, since the main PR was already accepted we'll only wait for the build to pass.

johnktims:
Thanks for backporting this fix, @jotak.

jotak:
you're welcome

PR #1456, opened 2015-07-24
Fix typos

johnktims:


querydsl-jpa/src/main/java/com/querydsl/jpa/Conversions.java
```diff
@@ -63,16 +63,16 @@ private static boolean isEntityPathAndNeedsWrapping(Expression<?> expr) {
         return false;
     }
 
-    private static <RT> FactoryExpression<RT> createEntityPathConversions(FactoryExpression<RT> factorye) {
+    private static <RT> FactoryExpression<RT> createEntityPathConversions(FactoryExpression<RT> factory) {
         List<Expression<?>> conversions = Lists.newArrayList();
-        for (Expression<?> e : factorye.getArgs()) {
+        for (Expression<?> e : factory.getArgs()) {
```
Shredder121: Line 38 uses `factoryExpr` (I can recall it being `factorye` at first as well).
Do you think that is better to use here?

johnktims: Indeed. I'll update it.

timowest:
This looks good to me.

Shredder121:
To me as well.

PR #1458, opened 2015-07-25
Default variable case customization

marcellor:
Replaced case transformer interface with variable name guava function.
Added RuntimeException catch clause.
Solved check style rule violations.
Merging branch.


marcellor:
The uncapSimpleName of Public domain is generated as "public", which is not allowed as identifier:

[ERROR] COMPILATION ERROR : 
[ERROR] /home/travis/build/querydsl/querydsl/querydsl-apt/target/generated-test-sources/java/com/querydsl/apt/domain/QPublic.java:[20,32] expected

querydsl-codegen/src/main/java/com/querydsl/codegen/CodegenModule.java
```diff
@@ -56,6 +61,7 @@ protected void configure() {
         bind(EmbeddableSerializer.class);
         bind(ProjectionSerializer.class);
         bind(SupertypeSerializer.class);
+        bind(String.class);
```
timowest: You should use named injection instead, like with the keywords, String is not a service interface

querydsl-codegen/src/main/java/com/querydsl/codegen/EntitySerializer.java
```diff
@@ -315,7 +326,7 @@ protected void introClassHeader(CodeWriter writer, EntityType model) throws IOEx
     }
 
     protected void introDefaultInstance(CodeWriter writer, EntityType model, String defaultName) throws IOException {
-        String simpleName = !defaultName.isEmpty() ? defaultName : model.getUncapSimpleName();
+        String simpleName = !defaultName.isEmpty() ? defaultName : variableNameFunction.apply(model);
```
timowest: You need to preserve the escaping functionality of getUncapSimpleName(), that's the reason for the QPublic and QPrivate issues (public and private can't be used as variable names)

marcellor:
New commit on branch default_variable_case_customization:
- Preserved the escaping and the property renaming functionalities by moving the "variable naming" guava Function from the EntitySerializer to the EntityType constructor.
- Renamed getUncapedSimpleName to getModifiedSimpleName
- Changed calls to EntityType constructor.

querydsl-codegen/src/main/java/com/querydsl/codegen/CodegenModule.java
```diff
@@ -56,6 +58,7 @@ protected void configure() {
         bind(EmbeddableSerializer.class);
         bind(ProjectionSerializer.class);
         bind(SupertypeSerializer.class);
+        bind(Function.class, DefaultVariableNameFunction.class);
```
timowest: Function is also too general, a named injection should be used.

querydsl-codegen/src/main/java/com/querydsl/codegen/EntityType.java
```diff
@@ -172,8 +180,8 @@ public Supertype getSuperType() {
         return superTypes;
     }
 
-    public String getUncapSimpleName() {
-        return uncapSimpleName;
+    public String getModifiedSimpleName() {
```
timowest: this is not a backwards compatible change, consider deprecating the original one instead

querydsl-codegen/src/main/java/com/querydsl/codegen/EntityType.java
```diff
@@ -253,4 +261,9 @@ private Property validateField(Property field) {
     public Type getInnerType() {
         return type;
     }
+
+    public Function<EntityType, String> getVariableNameFunction() {
```
timowest: why is the function exposed?

querydsl-codegen/src/main/java/com/querydsl/codegen/EntityType.java
```diff
@@ -66,14 +75,13 @@ public EntityType(Type type) {
      *
      * @param type
      * @param superTypes
+     * @param variableNameFunction
      */
-    public EntityType(Type type, Set<Supertype> superTypes) {
+    public EntityType(Type type, Set<Supertype> superTypes, Function<EntityType, String> variableNameFunction) {
```
timowest: Maybe getUncapSimpleName could be deprecated and the transformation done outside of EntityType?

```java
String variableName = variableNameFunction(type.getSimpleName());
if (JavaSyntaxUtils.isReserved(name)) {
    variableName = variableName + "$";
}
```

querydsl-codegen/src/main/java/com/querydsl/codegen/CodegenModule.java
marcellor: Actually we do not even need to inject it. The variableNameFunction is read from the Configuration object and passed in the EntityTypeFactory:
```
this.typeFactory = new ExtendedTypeFactory(processingEnv, entityAnnotations, typeMappings, queryTypeFactory, conf.getVariableNameFunction());
```

querydsl-codegen/src/main/java/com/querydsl/codegen/EntityType.java
marcellor: It's true, there is no need to expose it.

querydsl-codegen/src/main/java/com/querydsl/codegen/EntityType.java
marcellor: I have tried to do the transformation in `EntitySerializer.introDefaultInstance`
```
String simpleName = !defaultName.isEmpty() ? defaultName : model.getModifiedSimpleName();
```
 but this way we were missing the logic in `EntityType` which mutates the `uncapSimpleName` before adding a `Property`
```
private Property validateField(Property field) {
        if (field.getName().equals(modifiedSimpleName) || field.getEscapedName().equals(modifiedSimpleName)) {
            do {
                modifiedSimpleName = StringUtils.uncapitalize(getType().getSimpleName()) + (escapeSuffix++);
            } while (propertyNames.contains(modifiedSimpleName));
        }
        return field;
    }
```
I found that the `EntityType` constructor was the most sensible place to make the transformation. The only problem being that `EntityType` is not a dependency injection managed class, so we have to change the code that instantiates it.

marcellor:
Hi @timowest, I made a new commit to the branch:
- Switched to `variableNameFunction` named injection.
- Reverted `getUncapSimpleName` in `EntityType` and deprecate it.
- No need to expose `variableNameFunction` in `EntityType`.
- Added a default constructor to `EntityType` which instantiates a default `variableNameFunction`.
- Reverted `TypeResolver` and other modified classes.

Thanks

marcellor:
`querydsl-codegen` module is failing to build in Travis
```
[ERROR] Exit code: 1 - /home/travis/build/querydsl/querydsl/querydsl-codegen/src/main/java/com/querydsl/codegen/Serializer.java:34: warning: no description for @throws
```
but builds successfully in my local machine.

marcellor:
Hi @timowest, does the approach to capitalize the default variable name sound sensible?

querydsl-apt/src/main/java/com/querydsl/apt/DefaultConfiguration.java
```diff
@@ -177,6 +181,19 @@ public DefaultConfiguration(
             }
         }
 
+        if (options.containsKey(QUERYDSL_VARIABLE_NAME_FUNCTION_CLASS)) {
+            try {
+                variableNameFunction = (Function<EntityType, String>) Class.forName(options.get(QUERYDSL_VARIABLE_NAME_FUNCTION_CLASS)).newInstance();
```
Shredder121: If you split this line up to something like this:
```java
Class<Function<EntityType, String>> variableNameFunctionClass
        = (Class<Function<EntityType, String>>) Class.forName(options.get(QUERYDSL_VARIABLE_NAME_FUNCTION_CLASS));
variableNameFunction = variableNameFunctionClass.newInstance();
```

Then that means not the entire constructor carries the `@SuppressWarnings` anymore.

querydsl-apt/src/main/java/com/querydsl/apt/DefaultConfiguration.java
```diff
@@ -177,6 +181,19 @@ public DefaultConfiguration(
...
+            } catch (RuntimeException e) {
+                throw e;
+            } catch (Exception e) {
+               variableNameFunction = new DefaultVariableNameFunction();
```
Shredder121: I would still log these, just for the sake of debugging.

querydsl-codegen/src/main/java/com/querydsl/codegen/CapitalizedVariableNameFunction.java
```diff
@@ -0,0 +1,31 @@
...
+
+import com.google.common.base.Function;
+import com.mysema.codegen.StringUtils;
+import com.querydsl.core.util.JavaSyntaxUtils;
+
+/**
+ * Variable name generation strategy which capitalizes the first letter of the variable name.
+ *
+ */
+public class CapitalizedVariableNameFunction implements Function<EntityType, String> {
```
Shredder121: If you yourself have your `com.your.package.name.VariableNameFunction`, I wouldn't really see a reason to include this file.
We'll then also have to provide support for it.

querydsl-apt/src/main/java/com/querydsl/apt/DefaultConfiguration.java
marcellor: Makes sense.

querydsl-apt/src/main/java/com/querydsl/apt/DefaultConfiguration.java
marcellor: I couldn't find many example of logging, is this what you meant?
```
catch (RuntimeException e) {
   System.err.println(e.getMessage());
   processingEnv.getMessager().printMessage(Kind.ERROR, e.getMessage());
   throw e;
} catch (Exception e) {
   System.err.println(e.getMessage());
   processingEnv.getMessager().printMessage(Kind.WARNING, String.format("An exception prevented instantiating tha variable name function of type %s", options.get(QUERYDSL_VARIABLE_NAME_FUNCTION_CLASS)));
   variableNameFunction = new DefaultVariableNameFunction();
}
```
But this requires to adding `ProcessingEnvironment` to the `DefaultConfiguration` constructor.

querydsl-codegen/src/main/java/com/querydsl/codegen/CapitalizedVariableNameFunction.java
marcellor: I agree, let's remove that class.

querydsl-apt/src/main/java/com/querydsl/apt/DefaultConfiguration.java
Shredder121: @timowest What is your take on this? I think a syserr statement isn't that bad (it's only on exceptional behavior), or should we add a logger to that class?

timowest:
Like expressed before the serialization logic should be moved out of EntityType.

querydsl-apt/src/main/java/com/querydsl/apt/DefaultConfiguration.java
timowest: Preferably using a logger.

Shredder121:
Also, since `getUncapSimpleName()` is now deprecated, all occurences of `model.getUncapSimpleName` in the `querydsl-scala` module can be changed to `model.getModifiedSimpleName`.

marcellor:
> Like expressed before the serialization logic should be moved out of EntityType.

@timowest , could you please be more specific on which logic should be moves outside of `EntityType`?
I can see that the core version of `EntityType` already contains serialization logic, as in:
```
public EntityType(Type type, Set<Supertype> superTypes) {
        super(type);
        this.modifiedSimpleName = StringUtils.uncapitalize(type.getSimpleName());
        if (JavaSyntaxUtils.isReserved(modifiedSimpleName)) {
            this.modifiedSimpleName = modifiedSimpleName + "$";
        }
        this.superTypes = superTypes;
    }
```
and
```
private Property validateField(Property field) {
        if (field.getName().equals(modifiedSimpleName) || field.getEscapedName().equals(modifiedSimpleName)) {
            do {
                modifiedSimpleName = StringUtils.uncapitalize(getType().getSimpleName()) + (escapeSuffix++);
            } while (propertyNames.contains(modifiedSimpleName));
        }
        return field;
    }
```
Thanks

querydsl-codegen/src/main/java/com/querydsl/codegen/CodegenModule.java
```diff
@@ -48,6 +48,11 @@
      */
     public static final String IMPORTS = "imports";
 
+    /**
+     * key for the custom imports set
```
Shredder121: I just noticed, this javadoc is not yet updated.

timowest:
@marcellor You are right that there is already serialization logic inside EntityType. This looks good to me then.

marcellor:
Hi @timowest, I've committed the following:
1. Added a logger to DefaultConfiguration.
2. Added slf4j-log4j12 dependency to querydsl-apt.
3. Removed unused variable name function.
4. Renamed deprecated method in querydsl-scala.
Can you please verify why is travis failing on executing the maven-javadoc-plugin?
What are the next steps?
Thanks.

querydsl-codegen/src/main/java/com/querydsl/codegen/EntityType.java
```diff
@@ -58,7 +65,18 @@
      * @param type
      */
     public EntityType(Type type) {
+        this(type, new DefaultVariableNameFunction());
+    }
+
+    /**
+     * Create a new {@code EntityType} instance for the given type
+     *
+     * @param type
```
Shredder121: This should have a description, the Java 8 Javadoc linter is strict like that.

querydsl-codegen/src/main/java/com/querydsl/codegen/EntityType.java
```diff
@@ -69,11 +87,21 @@ public EntityType(Type type) {
...
-        }
         this.superTypes = superTypes;
+        this.modifiedSimpleName = new DefaultVariableNameFunction().apply(this);
+    }
+
+    /**
+     * Create a new {@code EntityType} instance for the given type and superTypes
+     *
+     * @param type
+     * @param superTypes
```
Shredder121: These should also have a description

querydsl-codegen/src/main/java/com/querydsl/codegen/EntityType.java
```diff
@@ -172,8 +200,16 @@ public Supertype getSuperType() {
         return superTypes;
     }
 
+    /**
+     * Use {@link #getModifiedSimpleName()}
```
Shredder121: This still needs a `@return` part.

querydsl-apt/pom.xml
```diff
@@ -79,6 +79,15 @@
       <artifactId>morphia</artifactId>
       <version>${morphia.version}</version>
       <scope>provided</scope>
+    </dependency>
+    <dependency>
```
timowest: why is this dependency needed here?

marcellor: We need the `slf4j-log4j12` to use `org.slf4j.Logger` in `DefaultConfiguration`

timowest: Consider using a different means of logging, the querydsl-apt transitive dependencies should be quite minimal.

marcellor: I suggest we remove the logging from this class, I cannot see any other logging in apt.

marcellor:
Hi @timowest this looks now ready to merge.

querydsl-codegen/src/main/java/com/querydsl/codegen/EntityType.java
```diff
@@ -50,30 +57,51 @@
...
+     * @param superTypes the super types to be used
      */
     public EntityType(Type type, Set<Supertype> superTypes) {
         super(type);
-        this.uncapSimpleName = StringUtils.uncapitalize(type.getSimpleName());
-        if (JavaSyntaxUtils.isReserved(uncapSimpleName)) {
-            this.uncapSimpleName = uncapSimpleName + "$";
-        }
         this.superTypes = superTypes;
+        this.modifiedSimpleName = new DefaultVariableNameFunction().apply(this);
```
Shredder121: Could you maybe look if this could be removed?

marcellor: I've removed the unused constructor with three parameters.
We cannot remove the `new DefaultVariableNameFunction().apply(this)` from the `EntityType` constructor, unless we end up with merging the two constructors and changing all the calls in the code base. I would't see any benefit in that.

marcellor: Probably the second approach looks cleaner, we could make the three parameters constructor `private`, what do you think?

Shredder121: I think cascading like that is perfect.
If the constructor is not necessary I'd keep it private yes.
We could always change things a bit.
After that it looks good to me as well.

marcellor:
Great, I've committed the changes to the constructors.

marcellor:
HI @timowest , @Shredder121 , as reported above the branch is in conflict with the trunk.
Is this pull request going to be merged? I do not have write access to do that.
Please let me know. Thanks.

timowest:
Yes, we can merge it, but the conflicts are for you to resolve. Consider rebasing.

marcellor:
Where do I see the conflicts? Can you send me the logs please?

querydsl-codegen/src/main/java/com/querydsl/codegen/EntitySerializer.java
```diff
@@ -172,7 +172,7 @@ protected void constructorsForVariables(CodeWriter writer, EntityType model) thr
         if (stringOrBoolean) {
             writer.line(thisOrSuper,"(forVariable(variable)",additionalParams,");");
         } else {
-            writer.line(thisOrSuper,"(", localName.equals(genericName) ? EMPTY : "(Class)",
+            writer.line(thisOrSuper,"(", localName.equals(genericName) ? EMPTY : "(Class) ",
```
Shredder121: The conflicts are on this line and line 77, they are already resolved.
Consider removing them, since they are not part of name resolving.

marcellor:
I am not very familiar with git, and while trying to merge I cannot see my changes anymore.
Can you please provide some pointers?
Thanks

marcellor:
I am using Tortois git, and there seems not to be an option to checkout a particular commit tag, you can only checkout branches. Also by checking the logs I cannot see the last commits.

Shredder121:
I think you now combined your original pull request and the new one, I don't think you intended this?
If you want I can open up a new PR with the commits for you. Then we can merge that.

marcellor:
Yes please, can you open a new PR with the my previous commits?

Shredder121:
I now added the PR with your commits.

marcellor:
Thanks Ruben, please let me know if I need to make any other changes in order to merge.

timowest:
Closing in favor of https://github.com/querydsl/querydsl/pull/1523

Issue #1459, opened 2015-07-28
HibernateDomainExporter and inheritance with composite-id


vveloso:
With QueryDSL 3.6.5, the following XML mappings generate public final fields that are not initialized in leaf types.

```xml
<hibernate-mapping default-access="field" default-lazy="false">
 <class abstract="true"
  name="ConnectionRouteDb" table="DCN_NE_ROUTES">
  <composite-id access="field" name="routeKey" class="ConnectionRouteDbKey">
   <key-property name="type" type="string" column="ROUTE_TYPE"/>
   <key-property name="key" type="string" column="ROUTE_KEY"/>
  </composite-id>
  <discriminator force="false" insert="false" not-null="true">
   <column name="ROUTE_TYPE"/>
  </discriminator>
  <version access="field" column="VERSION" generated="never"
   name="version" type="int"/>
  <property access="field" generated="never" lazy="false" name="cost" type="int">
   <column name="COST"/>
  </property>
  <property access="field" generated="never" lazy="false" name="neId" type="int">
   <column name="NE_ID"/>
  </property>
 </class>
</hibernate-mapping>
```

```xml
<hibernate-mapping>
 <subclass discriminator-value="GATEWAY"
  extends="ConnectionRouteDb" lazy="false"
  name="GatewayRouteDb" select-before-update="false">
 </subclass>
</hibernate-mapping>
```
Generates:
```java
@Generated("com.mysema.query.codegen.EntitySerializer")
public class QNeConnectionRouteDb extends EntityPathBase<NeConnectionRouteDb<?, ?>> {

    private static final long serialVersionUID = -2050193206L;

    private static final PathInits INITS = PathInits.DIRECT2;

    public static final QNeConnectionRouteDb neConnectionRouteDb = new QNeConnectionRouteDb("neConnectionRouteDb");

    public final com.ossnms.bicnet.emne.core.entities.ne.data.QNeConnectionRouteData_NeConnectionRoutePrototype _super = new com.ossnms.bicnet.emne.core.entities.ne.data.QNeConnectionRouteData_NeConnectionRoutePrototype(this);

    public final NumberPath<Integer> cost = createNumber("cost", Integer.class);

    public final NumberPath<Integer> neId = createNumber("neId", Integer.class);

    public final QNeConnectionRouteDbKey routeKey;

    public final NumberPath<Integer> version = createNumber("version", Integer.class);

    @SuppressWarnings("all")
    public QNeConnectionRouteDb(String variable) {
[...]

PR #1460, opened 2015-07-30
HibernateDomainExporter and inheritance with composite-id

vveloso:
Modifications required to produce compilable code as described in https://github.com/querydsl/querydsl/issues/1459 .

The change did not break existing unit tests.

### DIFF
--- a/querydsl-codegen/src/main/java/com/mysema/query/codegen/EntitySerializer.java
+++ b/querydsl-codegen/src/main/java/com/mysema/query/codegen/EntitySerializer.java
@@ -64,10 +64,11 @@ public class EntitySerializer implements Serializer {
         this.keywords = keywords;
     }
 
-	private boolean superTypeHasEntityFields(EntityType model) {
-		Supertype superType = model.getSuperType();
-		return null != superType && null != superType.getEntityType() && superType.getEntityType().hasEntityFields();
-	}
+    private boolean superTypeHasEntityFields(EntityType model) {
+        Supertype superType = model.getSuperType();
+        return null != superType && null != superType.getEntityType()
+                && superType.getEntityType().hasEntityFields();
+    }
 	
     protected void constructors(EntityType model, SerializerConfig config,
             CodeWriter writer) throws IOException {
--- a/querydsl-codegen/src/main/java/com/mysema/query/codegen/EntitySerializer.java
+++ b/querydsl-codegen/src/main/java/com/mysema/query/codegen/EntitySerializer.java
@@ -64,18 +64,25 @@ public class EntitySerializer implements Serializer {
         this.keywords = keywords;
     }
 
+	private boolean superTypeHasEntityFields(EntityType model) {
+		Supertype superType = model.getSuperType();
+		return null != superType && null != superType.getEntityType() && superType.getEntityType().hasEntityFields();
+	}
+	
     protected void constructors(EntityType model, SerializerConfig config,
             CodeWriter writer) throws IOException {
+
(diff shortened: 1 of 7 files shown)


timowest:
@vveloso Thanks for the PR! Could you also add a test case that fails without your fix?

timowest:
Also could you use a more descriptive commit message?

vveloso:
I'm not sure what would be the best way to test this. I don't like to assert on text output, but it may well be what's required in this case. I'll take another look at the existing test cases to figure out how to go about building a new test case.

Shredder121:
If you add your source that didn't compile at first (or at least the relevant parts) to the test sources, then it will be processed, if there is a regression it will not compile as before.
Does that sound like something that you can do?

vveloso:
@Shredder121, I hope that this is what you meant. If not, it still seems to work: if the change I made to EntitySerializer is removed the Java code generated by the unit test will not compile.

Shredder121:
> ... if the change I made to EntitySerializer is removed the Java code generated by the unit test will not compile

That is exactly what I meant, then if (for some reason) the logic is changed in such a way that the code doesn't compile anymore we'll know about it.

querydsl-jpa-codegen/src/test/java/com/mysema/query/jpa/domain18/Route1.java
```diff
@@ -0,0 +1,28 @@
+package com.mysema.query.jpa.domain18;
+
+public class Route1 {
+
+	private RouteDbKey routeKey;
```
timowest: no tabs, only spaces

vveloso: How many spaces?

timowest: 4 spaces

vveloso: Done.

querydsl-codegen/src/main/java/com/mysema/query/codegen/EntitySerializer.java
```diff
@@ -64,18 +64,25 @@ public EntitySerializer(TypeMappings mappings, @Named("keywords") Collection<Str
         this.keywords = keywords;
     }
 
+	private boolean superTypeHasEntityFields(EntityType model) {
```
timowest: indentation

querydsl-jpa-codegen/src/test/java/com/mysema/query/jpa/domain18/RouteDbKey.java
```diff
@@ -0,0 +1,21 @@
+package com.mysema.query.jpa.domain18;
+
+public class RouteDbKey {
+
+	private String type;
```
timowest: tabs to spaces

vveloso:
All spaces in the right places (hopefully).

vveloso:
Sorry about the accidental close&reopen.

@timowest I hope that the latest commits address your concerns.

timowest:
Looks good. Could you also provide a PR for the master branch?

Shredder121:
To me it also looks good, so if the master counterpart could be created it is fine to merge.

vveloso:
@timowest I just created the PR for the master branch (#1509).

johnktims:
Looks good to me. Thanks for the PR @vveloso!

PR #1461, opened 2015-07-31
Add StatementOptions

### LINKED ISSUE #1419: SQLQuery option for Statement.setFetchSize(int)
Currently setting fetchSize hint for Statement requires custom implementation using SQLDetailedListener. As Statement.setFetchSize(int) is part of standard JDBC API, it would make sense to implement this in core API.

johnktims:
Fixes #1419

### DIFF
--- a/querydsl-sql/src/main/java/com/querydsl/sql/AbstractSQLQuery.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/AbstractSQLQuery.java
@@ -67,6 +67,7 @@ public abstract class AbstractSQLQuery<T, Q extends AbstractSQLQuery<T, Q>> exte
     private Object lastCell;
 
     private SQLListenerContext parentContext;
+    private StatementOptions statementOptions = new StatementOptions.Builder().build();
 
     public AbstractSQLQuery(@Nullable Connection conn, Configuration configuration) {
         this(conn, configuration, new DefaultQueryMetadata());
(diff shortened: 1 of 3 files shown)


johnktims:
This still needs some tests.

querydsl-sql/src/main/java/com/querydsl/sql/AbstractSQLQuery.java
```diff
@@ -237,6 +238,17 @@ public void close() throws SQLException {
         }
     }
 
+    private PreparedStatement getPreparedStatement(String queryString) throws SQLException {
+        PreparedStatement statement = conn.prepareStatement(queryString);
+
+        statement.setFetchSize(statementOptions.getFetchSize());
```
timowest: call the setter only when the argument is non null

querydsl-sql/src/main/java/com/querydsl/sql/AbstractSQLQuery.java
```diff
@@ -67,6 +67,7 @@
     private Object lastCell;
 
     private SQLListenerContext parentContext;
+    private StatementOptions statementOptions = new StatementOptions.Builder().build();
```
timowest: maybe via StatementOptions.builder()...build()

builder() as a factory method for the builder

johnktims:
Thanks for cleaning this up and adding a test. This looks good to me now.

Shredder121:
Yes, this looks good to me as well.

PR #1469, opened 2015-08-05
Extend CaseForEqBuilder signature

### LINKED ISSUE #1253: CaseForEqBuilder generates SimpleOperation instead of concrete type
Hello,

It seems somehow the same issue as:
https://github.com/querydsl/querydsl/issues/702
but applied to CaseForEqBuilder instead of CaseBuilder.

The expression returned by CaseForEqBuilder.Cases.otherwise() comes from a call to abstract method createResult.
There are three implementations of createResult, one that creates a SimpleOperation, another that creates a NumberOperation and the last one that create a StringOperation.
The implementation used depends on which CaseForEqBuilder.then method is used.

But for example, there is no way to obtain a DateExpression, a DateTimeExpression...

Perhaps a solution would be to do somehow like in CaseBuilder.then(Expression<A> expr).
Something like:

```java
	public <T> CaseForEqBuilder<D>.Cases<T, Expression<T>> then(final Expression<T> then)
	{
		type = then.getType();
		return this.inner.new Cases<T, Expression<T>>()
		{
			@Override
			protected Expression<T> createResult(Class<T> type, Expression<T> last)
			{
				if (then instanceof StringExpression)
				{
					return (Expression<T>) StringOperation.create(Ops.CASE_EQ, base, last);
				}
				else if (then instanceof NumberExpression)
				{
					return (Expression<T>) NumberOperation.create(type, base, last);
				}
				else if (then instanceof DateExpression)
				{
					return (Expression<T>) DateOperation.create(type, Ops.CASE_EQ, base, last);
				}
				else if (then instanceof DateTimeExpression)
				{
					return (Expression<T>) DateTimeOperation.create(type, Ops.CASE_EQ, base, last);
				}
				else
				{
					return SimpleOperation.create(type, Ops.CASE_EQ, base, last);
				}
			}
		}.when(this.other).then(then);
	}
```
What do you think about that?

timowest:
Fixes #1253

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -251,6 +251,7 @@
                   <compatibilityType>BACKWARD_COMPATIBLE_USER</compatibilityType>
                   <dumpDetails>true</dumpDetails>
                   <publicOnly>true</publicOnly>
+                  <previousVersion>4.0.0</previousVersion>
                 </requireBackwardCompatibility>
               </rules>
             </configuration>
--- a/querydsl-core/src/main/java/com/querydsl/core/types/dsl/CaseForEqBuilder.java
+++ b/querydsl-core/src/main/java/com/querydsl/core/types/dsl/CaseForEqBuilder.java
@@ -13,15 +13,14 @@
  */
 package com.querydsl.core.types.dsl;
 
+import java.sql.Timestamp;
 import java.util.ArrayList;
+import java.util.Date;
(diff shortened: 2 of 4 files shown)


Shredder121:
This one looks good

johnktims:
Looks good to me.

PR #1480, opened 2015-08-13
Fix listener execution

### LINKED ISSUE #1479: Insert operation doesn't fire listener executed() for non-batch operations
SQLInsertClause.executeWithKeys()

 334:             listeners.preExecute(context);
 335:             stmt.executeUpdate();
 336:             listeners.preExecute(context);   <---- Should this not be executed() ?

timowest:
Fixes #1479


Shredder121:
To me this looks good.

johnktims:
The fix looks correct but I think there should be a test.

Shredder121:
Hmm, yes you're right.
I could make something if you want.

timowest:
Ok

Shredder121:
I haven't yet had time to finish the tests, but is that a blocker for this issue? @johnktims

johnktims:
The fix looks correct and I'd like to see it in the release so I'll just merge it and we can add some tests later.

PR #1481, opened 2015-08-14
Added an option to switch between field and getter return types.

DaSilva2010:
We use Optional (guava) in our getters for fields that are nullable. But we don't want Optional for the generated classes. So this is our suggestion to make the generated return type more user-defined.

### DIFF
--- a/querydsl-maven-plugin/src/main/java/com/querydsl/maven/AbstractExporterMojo.java
+++ b/querydsl-maven-plugin/src/main/java/com/querydsl/maven/AbstractExporterMojo.java
@@ -77,6 +77,13 @@ public abstract class AbstractExporterMojo extends AbstractMojo {
     private boolean handleMethods = true;
 
     /**
+     * switch for usage of field types instead of getter types
+     *
+     * @parameter default-value-false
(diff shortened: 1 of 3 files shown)


Shredder121:
Thanks for the PR, may I ask if you could maybe remove the whitespace changes?
I'll also add some more comments on a few lines.

querydsl-maven-plugin/src/main/java/com/querydsl/maven/AbstractExporterMojo.java
```diff
@@ -77,6 +77,13 @@
     private boolean handleMethods = true;
 
     /**
+     * switch for usage of field types instead of getter types
+     *
+     * @parameter default-value-false
```
Shredder121: This should be `default-value=false`.

DaSilva2010:
I don't understand why the CI build fails since there were only whitespace changes between this and the last successful build.

timowest:
Looks good.

Shredder121:
Yes, looks good.
Thanks again for the PR!

PR #1485, opened 2015-08-17
Fix boolean defaults for AntMetaDataExporter

### LINKED ISSUE #1483: AntMetaDataExporter Defaults Incorrect
The documentation for the ant metadata exporter says that exportTables, exportViews and exportForeignKeys all default to true. In fact, they all default to false (possibly exportPrimaryKeys as well). This was changed between 3.6.5 and 4.0.0.

This caused my build script to unexpectedly stop generating code with no error when I updated to 4.0.0. I solved it by explicitly setting exportTables and exportViews back to true.

timowest:
Fixes #1483


johnktims:
This looks good to me.

Shredder121:
Yes, it's fine to me as well.

PR #1488, opened 2015-08-18
Document rename functionality

### LINKED ISSUE #1473: MetadataExporter should provide option to override output schema
We are using Oracle in this case. We create our `Q*` classes via `Gradle + AntMetadataExplorer` without JPA model. When running the code generator, the schema rendered into the `Q*` classes in the end is the username.

In our use case that is a bit problematic. Let´s assume we have a semantical schema `APPDATA`. The codegen runs against a local dev vm, where we have multiple schemas, e.g. `APPDATA_STAGE` and `APPDATA_INTEGRATION_TEST`.

For sure I can take one of these and remap it at runtime with `Configuration.registerSchemaOverride(…)`, but then the mapping would be:

`APPDATA_STAGE` => `APPDATA[_ANY_ENV_SPECIFIC_FLAG]`

It would be nice to have a schema override setting within `MetadataExporter` so that I can create the `Q*` classes with a semantically correct schema name, so that at runtime may mapping would look like:

`APPDATA` => `APPDATA[_ANY_ENV_SPECIFIC_FLAG]`

As a little motivation :-) jOOQ code generator provides this feature.

timowest:
Fixes #1473

### DIFF
--- a/querydsl-docs/src/main/docbook/en-US/content/tutorials/sql.xml
+++ b/querydsl-docs/src/main/docbook/en-US/content/tutorials/sql.xml
@@ -346,9 +346,8 @@
 
     <para>Imports can be used to add cross-schema foreign keys support.</para>
 
-    <para>Renamings of schemas, tables and columns can also be applied using the plugin. At first an
-    example of a schema rename</para>
-
+    <para>Schemas, tables and columns can also be renamed using the plugin. Here are some examples:</para>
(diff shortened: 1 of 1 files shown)


johnktims:
I changed the wording a little bit.

Shredder121:
That looks good.
The PR looks good to me.

Issue #1212, opened 2015-02-18
`distinct on (field) field, other_field`  possible?


caarlos0:
In SQL (not sure if ANSI, but in at least in Postgres) we can distinct by one field using `distinct on`:

```sql
select distinct on (field) field, other_field
from table
where other_field like '%blah%';
```

I didn't find a way to write this with QueryDSL (using JPA).

Is this an uninplemented feature, a limitation by JPA (or something else), or I just didn't look well enough?

timowest:
This is unsupported in JPQL and currently also not implemented in Querydsl SQL.

caarlos0:
Figured.

Thanks.

PR #1490, opened 2015-08-21
add support for postgresql specific DISTINCT ON syntax (v4)

dharaburda:
Simple change to allow for using DISTINCT ON, a very useful postgres specific feature:
http://www.postgresql.org/docs/9.4/static/sql-select.html#SQL-DISTINCT

Another user inquired about this in [#1212].

Also submitted a PR for v3 [#1489].

### DIFF
--- a/querydsl-sql/src/main/java/com/querydsl/sql/postgresql/PostgreSQLQuery.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/postgresql/PostgreSQLQuery.java
@@ -95,9 +95,10 @@ public class PostgreSQLQuery<T> extends AbstractSQLQuery<T, PostgreSQLQuery<T>> 
      * @param exprs
      * @return
      */
-    public PostgreSQLQuery<T> distinct(Expression<?>... exprs) {
-        return addFlag(Position.AFTER_SELECT, Expressions
-            .template(Object.class, "distinct on({0}) ", ExpressionUtils.list(Object.class, exprs)));
+    public PostgreSQLQuery<T> distinctOn(Expression<?>... exprs) {
+        return addFlag(Position.AFTER_SELECT,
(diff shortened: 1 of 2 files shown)


querydsl-sql/src/main/java/com/querydsl/sql/postgresql/PostgreSQLQuery.java
```diff
@@ -87,6 +89,18 @@ public PostgreSQLQuery(Connection conn, Configuration configuration, QueryMetada
         return addFlag(Position.END, builder.toString());
     }
 
+    /**
+     * adds a DISTINCT ON clause
+     *
+     * @param exprs
+     * @return
+     */
+    public PostgreSQLQuery<T> distinct(Expression<?>... exprs) {
```
timowest: maybe `distinctOn` instead? Otherwise it clashes partly with `distinct()`

timowest:
Thanks for the PR! I added some comments.

querydsl-sql/src/main/java/com/querydsl/sql/postgresql/PostgreSQLQuery.java
```diff
@@ -87,6 +89,18 @@ public PostgreSQLQuery(Connection conn, Configuration configuration, QueryMetada
...
+        return addFlag(Position.AFTER_SELECT, Expressions
```
timowest: maybe format as

```
return addFlag(Position.AFTER_SELECT, 
    Expressions.template(Object.class, "distinct on({0}) ", 
    ExpressionUtils.list(Object.class, exprs)));
```

dharaburda:
I updated both PRs with name change and formatting.

johnktims:
Looks good to me. Thanks for the PR @dharaburda!

Shredder121:
I agree, it looks good. Thanks!

PR #1492, opened 2015-08-22
Remove unused READMEs

johnktims:
I think the comments made in these READMEs have been addressed.


Shredder121:
And then we can also backport this, no need to keep unused files in the repository.

PR #1494, opened 2015-08-23
Enable dynamic exclusion of test categories

johnktims:
Using a property to set excludedGroups enables developers to dynamically configure the groups of tests run by the surefire plugin.

I added this because I often want to turn off reporting and performance to speed up tests and I want to be able to easily enable the external database tests.

`mvn clean test -DexcludedGroups=com.querydsl.core.testutil.ExternalDB,com.querydsl.core.testutil.Performance,com.querydsl.core.testutil.ReportingOnly`

or `mvn clean test -DexcludedGroups=` to run all tests.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -25,6 +25,9 @@
     <project.homepage>http://www.querydsl.com</project.homepage>
     <project.githubpage>http://github.com/querydsl/querydsl</project.githubpage>
     <project.checkout>scm:git:git@github.com:querydsl/querydsl.git</project.checkout>
+
+    <!-- default groups to exclude from surefire plugin -->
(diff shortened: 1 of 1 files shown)


timowest:
Should the profiles then also override the property instead? That might be more consistent.

johnktims:
Good point. I've overridden the property in the profiles.

timowest:
looks good

Shredder121:
We merged the 'really slow tests' PR first, so you'll have to set that up as default behavior again.
After that we can merge it.

johnktims:
I rebased and excluded the SlowTest category by default.

Shredder121:
Yes, looks good.

PR #1495, opened 2015-08-23
Add ability to mark tests as slow and disable slow tests in default b…

johnktims:
…uild


timowest:
looks good

Shredder121:
Looks good to me as well.

PR #1498, opened 2015-08-25
Fix groupBy handling for fetchResults

### LINKED ISSUE #1472: fetchResults() generating wrong count query
QueryDSL: 4.0.3

Given the scenario:

### Document.java ###
```JAVA
@Entity
@Table(name = "document")
public class Document {

@Id
private Long id;

@ManyToOne(fetch = FetchType.LAZY)
@JoinColumn(name = "subDepartmentId", nullable = false)
private SubDepartment subDepartment;

@Column
private BigDecimal value;
...
}
```

### SubDepartment.java ###
```JAVA
@Entity
@Table(name = "subDepartment")
public class SubDepartment{

@Id
private Long id;

@Embedded
private Department department;
...
}
```

### Department.java ###
```JAVA
@Entity
@Table(name = "department")
public class Department{

@Id
private Long id;

@Column(name="departmentCode")
private String code;

@Column(name="departmentName")
private String name;
...
}
```

When i try to do an aggregation query with some group by, it generates a wrong count query in fetchResults:

```JAVA
jpaQueryFactory.from(qDocument)
.select(qDocument.value.sum(), qDocument.subDepartment.department.code)
.groupBy(qDocument.subDepartment.department.code)
.fetchResults();
```

The generated count is:

```SQL
SELECT count(s.deparmentCode) AS col_0_0_
FROM document d CROSS JOIN
  subDepartment s
WHERE d.subDepartmentId = s.id

GROUP BY s.departmentCode
```

That will generate multiple results, not a single result.

Is this a bug or am i doing something wrong?

timowest:
Fixes #1472

### DIFF
--- a/querydsl-jpa/src/main/java/com/querydsl/jpa/JPQLSerializer.java
+++ b/querydsl-jpa/src/main/java/com/querydsl/jpa/JPQLSerializer.java
@@ -171,20 +171,25 @@ public class JPQLSerializer extends SerializerBase<JPQLSerializer> {
             append(SELECT).append(projection).append("\n");
 
         } else if (forCountRow) {
-            if (!metadata.isDistinct()) {
-                append(SELECT_COUNT);
-            } else {
+            if (!groupBy.isEmpty()) {
                 append(SELECT_COUNT_DISTINCT);
-            }
-            if (select != null) {
-                if (select instanceof FactoryExpression) {
-                    handle(joins.get(0).getTarget());
-                } else {
-                    // TODO : make sure this works
-                    handle(select);
-                }
+                handle(", ", groupBy);
             } else {
-                handle(joins.get(0).getTarget());
+                if (!metadata.isDistinct()) {
+                    append(SELECT_COUNT);
+                } else {
+                    append(SELECT_COUNT_DISTINCT);
+                }
+                if (select != null) {
+                    if (select instanceof FactoryExpression) {
+                        handle(joins.get(0).getTarget());
+                    } else {
+                        // TODO : make sure this works
+                        handle(select);
+                    }
+                } else {
+                    handle(joins.get(0).getTarget());
+                }
             }
             append(")\n");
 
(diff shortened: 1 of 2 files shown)


querydsl-jpa/src/main/java/com/querydsl/jpa/JPQLSerializer.java
```diff
@@ -171,20 +171,25 @@ public void serialize(QueryMetadata metadata, boolean forCountRow, @Nullable Str
...
                 } else {
-                    // TODO : make sure this works
-                    handle(select);
+                    append(SELECT_COUNT_DISTINCT);
+                }
+                if (select != null) {
+                    if (select instanceof FactoryExpression) {
+                        handle(joins.get(0).getTarget());
+                    } else {
+                        // TODO : make sure this works
```
johnktims: I'm fine with the PR so long as you've made sure this works. ;)

Shredder121:
This looks good to me. I'll merge this.

Burhan1991:
Hey Timo @timowest ,

the Fix #1498 for the Issue #1472 [fetchResults() generating wrong count query]
works only for single groupBy expression. 
For queries with multiple groupBy expressions like:

select a, b, sum(c) from EntityA

the COUNT query would be 
		select count(disitinct a, b) 
that throws a SyntaxException (at least with Hibernate 5).

This Exception looks to be OK with JPA 2.1:

COUNT ([DISTINCT] identification_variable | state_valued_path_expression |
single_valued_object_path_expression)

Any ideas?

timowest:
@Burhan1991 No ideas for now, do you have good suggestions?

pointwisefixed:
I ran into this issue using mssql which doesn't allow multiple columns for a distinct and I get a sql syntax exception. It's a funny one, if I try to use aggregations, I need to have a group by. If I use a group by (in mssql) I need to add the columns to the projection. But I also need to use pagination so to get the totals, we need a count and then I get a syntax exception in mssql (mysql and h2 work fine). During my research I ran into this, which has a potential solution for the generated hql:
https://hibernate.atlassian.net/browse/HHH-11042?page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel&showAll=true

PR #1502, opened 2015-08-28
Suppress warnings on methods

johnktims:


Shredder121:
A few `@SuppressWarnings` could be remedied in a different way, how shall I preserve your commit?
Do you want me to include it in yours, or make an extra commit which reverses a few changes?

johnktims:
It doesn't matter to me.

Shredder121:
I guess we can merge this, the end result will be the same.

PR #1504, opened 2015-08-28
Fix NullHandling treatment

### LINKED ISSUE #1503: JPQLQuery loses nulls last/first for nested properties
I'm trying to generate a JPQLQuery which is order by a field in nested property. So I do something like this:

```java
.orderBy(device.status.connectionTime.desc().nullsLast())
```
And the relevant part in the resulting sql looks like this:

```sql
order by devic1_.connection_time desc  
```

So the nulls last is missing. 

Looking at my debugger, I can see that the place where the null handling value is reset is the com.mysema.query.support.QueryMixin -class.

```java
public final T orderBy(OrderSpecifier<?> spec) {
        Expression<?> e = convert(spec.getTarget(), true);
        if (!spec.getTarget().equals(e)) {
            metadata.addOrderBy(new OrderSpecifier(spec.getOrder(), e));
        } else {
            metadata.addOrderBy(spec);
        }
        return self;
    }
```
It seems that after the conversion, the first branch gets chosen and there is a call to the two argument OrderSpecifier -constructor that looks like this:
```java
public OrderSpecifier(Order order, Expression<T> target) {
        this(order, target, NullHandling.Default);        
    }
```
So much for my nulls last then.. Shouldn't the orderBy -method in the QueryMixin -class also copy the null handling value over to the new OrderSpecifier?

timowest:
Fixes #1503

### DIFF
--- a/querydsl-core/src/main/java/com/querydsl/core/support/QueryMixin.java
+++ b/querydsl-core/src/main/java/com/querydsl/core/support/QueryMixin.java
@@ -339,7 +339,7 @@ public class QueryMixin<T> {
     public final T orderBy(OrderSpecifier<?> spec) {
         Expression<?> e = convert(spec.getTarget(), Role.ORDER_BY);
         if (!spec.getTarget().equals(e)) {
-            metadata.addOrderBy(new OrderSpecifier(spec.getOrder(), e));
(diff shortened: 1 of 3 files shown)


Shredder121:
These PRs look good.

johnktims:
Looks good.

PR #1509, opened 2015-08-31
HibernateDomainExporter and inheritance with composite-id.

vveloso:
Ported from pull request #1460 and issue #1459.

### DIFF
--- a/querydsl-codegen/src/main/java/com/querydsl/codegen/EntitySerializer.java
+++ b/querydsl-codegen/src/main/java/com/querydsl/codegen/EntitySerializer.java
@@ -64,17 +64,25 @@ public class EntitySerializer implements Serializer {
         this.keywords = keywords;
     }
 
+    private boolean superTypeHasEntityFields(EntityType model) {
+        Supertype superType = model.getSuperType();
+        return null != superType && null != superType.getEntityType()
+                && superType.getEntityType().hasEntityFields();
+    }
+
     protected void constructors(EntityType model, SerializerConfig config,
             CodeWriter writer) throws IOException {
+
         String localName = writer.getRawName(model);
         String genericName = writer.getGenericName(true, model);
 
-        boolean hasEntityFields = model.hasEntityFields();
+        boolean hasEntityFields = model.hasEntityFields() || superTypeHasEntityFields(model);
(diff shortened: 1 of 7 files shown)


querydsl-codegen/src/main/java/com/querydsl/codegen/EntitySerializer.java
```diff
@@ -64,18 +64,25 @@ public EntitySerializer(TypeMappings mappings, @Named("keywords") Collection<Str
         this.keywords = keywords;
     }
 
+    private boolean superTypeHasEntityFields(EntityType model) {
+        Supertype superType = model.getSuperType();
+        return null != superType && null != superType.getEntityType() && superType.getEntityType().hasEntityFields();
```
johnktims: For consistency's sake, could you please add a newline right before the second `&&`?

vveloso: Sure. Done.

johnktims:
The code looks good now, but the commits are a bit noisy. Would you mind squashing them? Sorry for being so pedantic. :)

vveloso:
I did the best I could. Everything looked good locally, I then had to merge & push, now it looks that there are even more commits. Sorry, I'm no Git guru...

vveloso:
Is this what you intended? I'm no Git expert, :)

vveloso:
That's pretty much what I did and it looks right.

The CI is failing with a "Killed" message. I don't think that it is related to this squash as the log looks right. Can you confirm that is indeed the case?

Shredder121:
Yes, sorry about that. Occasionally Travis thinks the Scala compilation uses too much resources, so it kills it.
I'll restart it.

johnktims:
@vveloso Sorry about that. I hadn't refreshed my browser and github was showing the original commits as well as the squashed commit.

vveloso:
@johnktims No problem. It happened to me as well. :)

@Shredder121 It seems that CIs have minds of their own.

Thank you both.

johnktims:
I'll just merge this since it's the same as #1460.

@vveloso Thanks again for the contribution and for taking the time to port it to 4.x.

PR #1516, opened 2015-09-01
Use querydsl-parent

### LINKED ISSUE #1132: querydsl-root POM changes

timowest:
Fixes #1132

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -10,9 +10,9 @@
   <url>${project.homepage}</url>
   
   <parent>
-    <groupId>com.mysema.home</groupId>
-    <artifactId>mysema-source</artifactId>
-    <version>0.3.1</version>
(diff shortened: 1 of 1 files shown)


johnktims:
Nice. Looks good.

Shredder121:
Missed this one.
Looks good. I'll merge these.

Do we know yet what Intellij does with the prolog and last line?

timowest:
No, I haven't yet investigated why that change happens

Shredder121:
I'll merge these anyway, we'll find it one day.

PR #1530, opened 2015-09-04
Introduce typed version of coalesce

### LINKED ISSUE #1376: Performing coalesce on NumberPath in from or where clause
`ComparableExpressionBase`'s `public final Coalesce<T> coalesce(Expression<?>...exprs)` returns a `Coalesce<T>`. `Coalesce` does not have `SimpleExpression`'s `public BooleanExpression eq(T right)` method, so I assume the idea is to use `Coalesce`'s `public NumberExpression<?> asNumber()`. However if you try to E.G: (I know the query is nonsense)

``` java
query()
.from(someRelationalPathBase)
.innerJoin(someOtherRelationalPathBase)
.on(someRelationalPathBase.id.coalesce(someOtherRelationalPathBase.id).asNumber().eq(someRelationalPathBase.id));
```

you receive the compiler error `The method eq(capture#2-of ?) in the type SimpleExpression<capture#2-of ?> is not applicable for the arguments (NumberPath<Long>)`. I'm working around this with the following code:

``` java
 public static <T extends Number & Comparable<T>> NumberOperation<T> coalesceNumbers(NumberPath<T> arg0, NumberPath<T> arg1) {
   return ((NumberOperation<T>) arg0.coalesce(arg1).asNumber());
 }
```

timowest:
Fixes #1376

### DIFF
--- a/querydsl-core/src/main/java/com/querydsl/core/types/dsl/BooleanOperation.java
+++ b/querydsl-core/src/main/java/com/querydsl/core/types/dsl/BooleanOperation.java
@@ -68,4 +68,28 @@ public class BooleanOperation extends BooleanExpression implements Operation<Boo
         }
     }
 
+    @Override
+    public BooleanExpression nullif(Expression<Boolean> other) {
+        return Expressions.booleanOperation(Ops.NULLIF, this, other);
+    }
+
+    @Override
+    public BooleanExpression nullif(Boolean other) {
+        return nullif(ConstantImpl.create(other.booleanValue()));
+    }
+
+    @Override
+    public BooleanExpression coalesce(SimpleExpression<Boolean> other) {
+        return Expressions.booleanOperation(Ops.COALESCE, Expressions.list(this, other));
+    }
+
+    @Override
+    public BooleanExpression coalesce(SimpleExpression<Boolean>... exprs) {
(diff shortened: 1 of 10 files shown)


timowest:
This is more of a sketch at the moment. The signature clashes with the old version of `coalesce`.

The new version provides better return types, but different name should be used to avoid clashes.

robertandrewbain:
Looks good to me, even as a sketch.

querydsl-core/src/main/java/com/querydsl/core/types/dsl/BooleanOperation.java
```diff
@@ -68,4 +68,28 @@ public BooleanExpression not() {
...
+
+    @Override
+    public BooleanExpression coalesce(SimpleExpression<Boolean> other) {
+        return Expressions.booleanOperation(Ops.COALESCE, Expressions.list(this, other));
+    }
+
+    @Override
+    public BooleanExpression coalesce(SimpleExpression<Boolean>... exprs) {
+        if (exprs.length == 0) {
+            return this;
```
robertandrewbain: Throw `IllegalArgumentException`?

robertandrewbain:
I'd expect to see some unit tests for this.

PR #1533, opened 2015-09-05
Remove reference to querydsl-root

johnktims:
The root pom is now in the root directory so developers no longer need to switch to ./querydsl-root/ to build the project.


Shredder121:
That sounds good to me. This looks good.

Issue #1525, opened 2015-09-03
Spring-Data/MongoDB/QueryDSL Searching nested _id of ObjectId type


bc15:
When an document from a collection is nested inside a document of another collection, it's pretty standard to make a copy of the nested document verbatim instead of creating another type of document just for the sake of nesting. Ex:

    category {"_id": ObjectId("c1"), "name": "Category 1"}
    question {"_id": ObjectId("q1"), category: {"_id": ObjectId("c1"), "name": "Category 1"}}

When using queryDSL as follows:

    question.category.id = "c1"

queryDSL generates a query like this:

    "question.category._id":"c1"

where I expect:

    "question.category._id":ObjectId("c1")

This works for top level documents, not for nested. I think this is a valid case and Spring should do the same translations it does for top level search. Is there a workaround for that?

see http://stackoverflow.com/questions/31249917/spring-data-mongodb-querydsl-searching-nested-id-of-objectid-type

vespermetsu:
I'm still having the same issue as described above. Here's what I was doing:

@Entity

public abstract class BaseEntity implements Serializable {

    @Id

    private String id;

..



@Document

@Entity

public class Company extends BaseEntity {

..



public class ActivitySchedule extends BaseEntity {

    @NotNull

    private Company company;

..

My predicate is as follow... 

builder = new BooleanBuilder();

activitySchedule = QActivitySchedule.activitySchedule;

builder.and(activitySchedule.company.id.eq(companyId));

This works for top level document perfectly but when i try to search nested document by id, it just can't be found. Other string field is fine though, except id.

I suspect that the problem is due to the following:

MongodbSerializer.java

+    protected boolean isId(Path<?> arg) {
+        // TODO override in subclass
+        return false;
+    }

Shouldn't it be something like Morphia version? (MorphiaSerializer.java)
     @Override
+    protected boolean isId(Path<?> arg) {
+        return arg.getAnnotatedElement().isAnnotationPresent(Id.class);
+    }

Thanks!

timowest:
MongodbSerializer is an abstract superclass, the template methods should be overriden by subclasses. They have dummy implementations because of backwards compatibility.

vespermetsu:
[...]

PR #1534, opened 2015-09-06
Improve Mongodb reference handling

### LINKED ISSUE #1528: Querying by reference ID when using Morphia @Reference annotation
Please add support into QueryDSL/generated Q-classes to enable querying a collection by referencing ID.

A real-world example: I have two collections: VirtualRouter with a key (column) podInternalId , which is a "foreign key"/referencing ID into Pod collection.

This is how their entity classes look like: 

```
@Entity(value = "VirtualRouter", noClassnameStored = true)
public class VirtualRouterEntity extends AbstractTraceableEntity<Long> implements InfrastructureEntity {

	@Id
	private long internalId;
	...
	@Reference(value = "podInternalId", lazy = false, idOnly = true)
	private PodEntity pod;

	...
	
	@Override
	public long getInternalId() {
		return internalId;
	}
	
	...
	
	public PodEntity getPod() {
		return this.pod;
	}
}
```

and

```
@Entity(value = "Pod", noClassnameStored = true)
public class PodEntity extends AbstractEntity<Long> implements InfrastructureEntity {

	@Id
	private long internalId;

	...

	@Override
	public long getInternalId() {
		return internalId;
	}
}
```

(Their Q-classes get generated as well)

I have no idea though how to generate queries which include VirtualRouterEntity.podInternalId key/column. When I tried:

```
String query = baseQuery(qe)
  .where(qe.pod.internalId.eq(podInternalId), qe.customerInternalId.eq(customerInternalId))
  .toString()
```

I got serialized query like this (don't mind the deletedDate part, it comes from baseQuery() method):

```
{ "deletedDate" : { "$exists" : false} , "pod._id" : 1 , "customerInternalId" : 3}
```

The problem is of course with the "pod._id" : 1 filter - it means that I'm querying for virtual routers with Pod subdocuments with a specific ID.

What I'd like to get is:

```
{ "deletedDate" : { "$exists" : false} , "podInternalId" : 1 , "customerInternalId" : 3}
```

I understand that MongoDB doesn't allow cross-collection queries but querying by "foreign key" within one collection should still work? Actually, shouldn't generated Q-classes for MongoDB/Morphia ignore referenced entities and only generate path relevant to their referencing IDs/keys (because, again, MongoDB AFAIK doesn't support cross-collection querying)?

I managed to workaround the problem by creating VirtualRouterEntity.getPodInternalId() method with this.pod.getInternalId() body but it doesn't feel right - especially when I decide to use lazy loading of referenced entitities.

timowest:
Fixes #1528 and https://github.com/querydsl/querydsl/issues/1525

### DIFF
--- a/querydsl-mongodb/src/main/java/com/querydsl/mongodb/MongodbSerializer.java
+++ b/querydsl-mongodb/src/main/java/com/querydsl/mongodb/MongodbSerializer.java
@@ -19,6 +19,7 @@ import java.util.Map;
 import java.util.regex.Pattern;
 
 import org.bson.BSONObject;
+import org.bson.types.ObjectId;
 
 import com.google.common.collect.Sets;
 import com.mongodb.BasicDBList;
@@ -97,12 +98,11 @@ public abstract class MongodbSerializer implements Visitor<Object, Void> {
                 } else {
                     throw new UnsupportedOperationException("Illegal operation " + expr);
                 }
-            } else if (isReference(expr, 0)) {
-                return asDBObject(asDBKey(expr, 0), asReference(expr, 1));
-            } else {
-                return asDBObject(asDBKey(expr, 0), asDBValue(expr, 1));
+            } else if (expr.getArg(0) instanceof Path) {
+                Path<?> path = (Path<?>) expr.getArg(0);
+                Constant<?> constant = (Constant<?>) expr.getArg(1);
+                return asDBObject(asDBKey(expr, 0), convert(path, constant));
             }
-
         } else if (op == Ops.STRING_IS_EMPTY) {
             return asDBObject(asDBKey(expr, 0), "");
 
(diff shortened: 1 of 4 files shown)


querydsl-mongodb/src/main/java/com/querydsl/mongodb/MongodbSerializer.java
```diff
@@ -138,11 +138,9 @@ public Object visit(Operation<?> expr, Void context) {
             return asDBObject("$or", list);
 
         } else if (op == Ops.NE) {
-            if (isReference(expr, 0)) {
-                return asDBObject(asDBKey(expr, 0), asDBObject("$ne", asReference(expr, 1)));
-            } else {
-                return asDBObject(asDBKey(expr, 0), asDBObject("$ne", asDBValue(expr, 1)));
-            }
+            Path<?> arg1 = (Path<?>) expr.getArg(0);
+            Constant<?> arg2 = (Constant<?>) expr.getArg(1);
```
johnktims: I would rename these variables to `path` and `constant` to be consistent with your changes on lines 198 and 215.

Shredder121:
This looks good to me.

johnktims:
Looks good.

PR #1538, opened 2015-09-07
Improve test categories

johnktims:
Instead of using `ExternalDB` as a test category for all tests requiring a database, this PR adds database-specific classes so that they can be run individually.

Examples: `mvn clean test -Psql -Dgroups=com.querydsl.core.testutil.H2 -DexcludedGroups=`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -27,8 +27,15 @@
     <project.checkout>scm:git:git@github.com:querydsl/querydsl.git</project.checkout>
 
     <!-- default groups to exclude from surefire plugin -->
-    <excludedGroups>com.querydsl.core.testutil.ExternalDatabase,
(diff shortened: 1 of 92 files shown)


timowest:
Looks really good, the case of the category names could be aligned with the naming in the SQLTemplates subclasses.

johnktims:
Good point. I've updated the names accordingly.

timowest:
Looks good, maybe still MongoDb -> MongoDB

Shredder121:
Looks good in my opinion as well, I agree with the MongoDb -> MongoDB comment.
After that it's good to merge.

johnktims:
Sure thing. I'll fix that tonight.

johnktims:
Done.

Shredder121:
Then I'll merge this.

PR #1539, opened 2015-09-07
Remove compiler warnings

Shredder121:
SuppressWarnings moved to smaller scope
Generic array creations avoided
A lot of rawtypes warnings removed
Suppress more warnings in generated code - `all` is an Eclipse extension

### DIFF
--- a/querydsl-core/src/main/java/com/querydsl/core/types/dsl/NumberExpression.java
+++ b/querydsl-core/src/main/java/com/querydsl/core/types/dsl/NumberExpression.java
@@ -219,10 +219,10 @@ public abstract class NumberExpression<T extends Number & Comparable<?>> extends
      * @param right
      * @return this / right
      */
-    @SuppressWarnings("unchecked")
     public <N extends Number & Comparable<?>> NumberExpression<T> divide(Expression<N> right) {
-        Class<?> type = getDivisionType(getType(), right.getType());
-        return Expressions.numberOperation((Class<T>) type, Ops.DIV, mixin, right);
+        @SuppressWarnings("unchecked")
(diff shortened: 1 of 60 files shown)


querydsl-core/src/main/java/com/querydsl/core/types/dsl/NumberExpression.java
```diff
@@ -203,9 +203,11 @@ private T cast(Number number) {
         return ceil;
     }
 
-    private Class<?> getDivisionType(Class<?> left, Class<?> right) {
+    private <N> Class<? extends N> getDivisionType(Class<N> left, Class<?> right) {
         if (!left.equals(right)) {
-            return Double.class;
+            @SuppressWarnings("unchecked") // This is a good in the middle decision
+            Class<? extends N> newType = (Class<? extends N>) Double.class;
+            return newType;
```
Shredder121: I am not sure about this method, it's now safer, doesn't introduce a new signature (erasure is the same) but what do you think?
@timowest @johnktims

timowest: Maybe a little overkill.

Shredder121: Then I'll remove it.

Shredder121:
Just mentioning that I am waiting for at least a new build that includes querydsl/codegen#30.

[This](https://github.com/querydsl/querydsl/pull/1539/files#diff-54ec2c68b1dee3365f8a251272cebc7eR820) use the new `suppressWarnings(String...)` method.

timowest:
Looks good.

johnktims:
Looks good to me.

PR #1544, opened 2015-09-11
Use the versions plugin to bump all versions

johnktims:
The person making the release will now be prompted for the new
version.

Example: Enter the new version to set 4.0.5.BUILD-SNAPSHOT: : 4.0.6

I added this because I noticed the examples weren't updated in the last release.

### DIFF
--- a/dist.sh
+++ b/dist.sh
@@ -2,6 +2,8 @@
 rm -rf target/dist
 mkdir -p target/dist
 
+mvn versions:set -DgenerateBackupPoms=false
+
 echo "Creating javadocs"
 mvn javadoc:aggregate
 
--- a/pom.xml
+++ b/pom.xml
@@ -353,11 +353,6 @@
         </configuration>
       </plugin>
       <plugin>
-        <groupId>com.mysema.maven</groupId>
-        <artifactId>maven-version-plugin</artifactId>
-        <version>0.1.0</version>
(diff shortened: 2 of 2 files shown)


timowest:
Looks good, this declaration could also be removed in favor of the more flexible versions plugin https://github.com/querydsl/querydsl/blob/master/pom.xml#L355-L359

johnktims:
I'm not familiar with that plugin, but if you think it would be better I can change it. I just wanted to automate this process so that none of the modules are forgotten. How is the mysema plugin more flexible?

Shredder121:
I think he meant you can delete the declaration of the mysema version plugin, and use the more flexible versions plugin instead.

timowest:
No, I meant remove the declaration I linked to in favor of the versions plugin.

johnktims:
Sorry for the misunderstanding. I'll remove the `maven-version-plugin` tonight.

johnktims:
`maven-version-plugin` has been removed.

Shredder121:
To me this looks good, let's hope that it works then for Timo.
You may merge this one Timo :wink:.

johnktims:
Should this PR be backported?

Shredder121:
I vote yes.
At least the new versions, and consistent release processes are better for Timo.

PR #1546, opened 2015-09-11
Remove validation from toString()

### LINKED ISSUE #1543: JPAQueryBase.toString() throws IllegalArgumentException: "no joins given"
In Querydsl 3.x the following worked:

```java
new JPAQuery<Object>(em).
  select(QUser.user).
  orderBy(QUser.user.firstname.desc()).
  toString();
```

In 4.0 this results in an `IllegalArgumentException`:

```
java.lang.IllegalArgumentException: No joins given
	at com.querydsl.jpa.JPAQueryBase.serialize(JPAQueryBase.java:48)
	at com.querydsl.jpa.JPAQueryBase.toString(JPAQueryBase.java:215)
	at org.springframework.data.jpa.repository.support.QuerydslIntegrationTests.testname(QuerydslIntegrationTests.java:77)
```

Isn't it a bit harsh to outright reject `toString()` calls on queries without a join? ;)

timowest:
Fixes #1543

### DIFF
--- a/querydsl-jpa/src/main/java/com/querydsl/jpa/JPAQueryBase.java
+++ b/querydsl-jpa/src/main/java/com/querydsl/jpa/JPAQueryBase.java
@@ -47,8 +47,14 @@ public abstract class JPAQueryBase<T, Q extends JPAQueryBase<T, Q>> extends Fetc
     protected abstract JPQLSerializer createSerializer();
 
     protected JPQLSerializer serialize(boolean forCountRow) {
-        if (queryMixin.getMetadata().getJoins().isEmpty()) {
-            throw new IllegalArgumentException("No joins given");
+        return serialize(forCountRow, true);
+    }
+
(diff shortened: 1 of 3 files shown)


querydsl-jpa/src/main/java/com/querydsl/jpa/JPAQueryBase.java
```diff
@@ -46,9 +46,11 @@ protected JPQLTemplates getTemplates() {
 
     protected abstract JPQLSerializer createSerializer();
 
-    protected JPQLSerializer serialize(boolean forCountRow) {
```
Shredder121: If you keep the original method's signature (and delegate to the new signature, default `true`) you get a few bonuses:
- If others have overridden the method(and added `@Override`), the code will still compile.
- You won't have to change every call site to the new method's signature, so you'll have a cleaner PR.

johnktims:
Looks good to me.

Shredder121:
To me as well, I'll merge them.

PR #1550, opened 2015-09-13
Update unsafe class names test

Shredder121:
This allows classes to be loaded that begin with com.sun or com.apple
which are not in those packages.


johnktims:
Looks good to me.

Shredder121:
Thanks :smile:

PR #1557, opened 2015-09-15
Add QHierarchyTest_A2 to the expected 'failures'

Shredder121:
Since the field and property accessor types are different,
the effective type isn't consistent so the build might fail


johnktims:
Looks great!

PR #1562, opened 2015-09-16
Run MetaDataExporterTest case in parallel

timowest:


querydsl-core/src/test/java/com/querydsl/core/testutil/Parallelized.java
```diff
@@ -0,0 +1,40 @@
...
+    public Parallelized(Class<?> klass) throws Throwable {
+        super(klass);
+        setScheduler(new ThreadPoolScheduler());
+    }
+
+    private static class ThreadPoolScheduler implements RunnerScheduler {
+        private ExecutorService executor;
+
+        public ThreadPoolScheduler() {
+            int numThreads = Runtime.getRuntime().availableProcessors();
```
Shredder121: Did you test how well this works?
I have 8 available processors, did you test it in some way?
Are all involved elements either local to the thread or threadsafe?

timowest: I tested it with 4 and it works well.
I am not sure if DatabaseMetadata is guaranteed to be threadsafe, but it seems to work here.
Without parallelization it task 1min 23secs for me and parallelized 1min.

Shredder121: The time is not that important to me, but how flaky it is / how much it fails.
If it works well with 4 then it should work well with 8.

timowest: I tried it also with a ForkJoinPool, but the time was the time.

querydsl-sql-codegen/src/test/java/com/querydsl/sql/codegen/MetaDataExporterAllTest.java
```diff
@@ -0,0 +1,162 @@
...
+    public static void tearDownClass() throws SQLException {
+        connection.close();
+    }
+
+    @Parameterized.Parameters
+    public static List<Object[]> parameters() {
+        List<Object[]> params = Lists.newArrayList();
+
+        List<NamingStrategy> ns = Arrays.<NamingStrategy>asList(
+                new DefaultNamingStrategy(), new OriginalNamingStrategy());
```
Shredder121: I find that this reads a little nicer.
```java
        List<? extends NamingStrategy> ns = Arrays.asList(
                new DefaultNamingStrategy(), new OriginalNamingStrategy());

```

Shredder121:
But of course it's not necessary.
Looks good to me.

timowest:
Sorry, I can still do the change.

johnktims:
Looks good.

PR #1566, opened 2015-09-17
Improve property escaping

### LINKED ISSUE #1560: Code generation fails with the following error
I am getting compilation errors when generating the source code. 

Error:
 public final BooleanPath 8FRecord = createBoolean("8FRecord");

COMPILATION ERROR :Return type for the method is missing
	- Syntax error, insert ")" to complete ConstructorDeclaration
	- Syntax error on tokens, AnnotationName expected instead
	- Syntax error, insert "Identifier (" to complete 
	 ConstructorHeaderName

Can some one please help me to fix that.

timowest:
Fixes #1560

### DIFF
--- a/querydsl-codegen/src/main/java/com/querydsl/codegen/Property.java
+++ b/querydsl-codegen/src/main/java/com/querydsl/codegen/Property.java
@@ -51,8 +51,7 @@ public final class Property implements Comparable<Property> {
 
     public Property(EntityType declaringType, String name, Type type, List<String> inits,
             boolean inherited) {
-        this(declaringType, name, JavaSyntaxUtils.isReserved(name) ? (name + "$") : name, type,
-                inits, inherited);
+        this(declaringType, name, escapeName(name), type, inits, inherited);
     }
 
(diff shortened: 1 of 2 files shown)


querydsl-codegen/src/main/java/com/querydsl/codegen/Property.java
```diff
@@ -65,6 +64,16 @@ public Property(EntityType declaringType, String name, String escapedName, Type
         this.inherited = inherited;
     }
 
+    private static String escapeName(String name) {
+        if (JavaSyntaxUtils.isReserved(name)) {
+            name = name + "$";
+        }
+        if (!Character.isJavaIdentifierStart(name.charAt(0))) {
```
Shredder121: They are mutually exclusive?
Or is that a micro optimization?

timowest: Good point.

Shredder121:
To me this looks good.

johnktims:
This looks good to me, too.

PR #1569, opened 2015-09-17
Remove Hibernate workaround

### LINKED ISSUE #1500: Wrong type when searching from enum in collection in Hibernate 4.3.9+
Our class `User` has a property `roles` which is `@Enumerated(EnumType.STRING) private Set<Role> roles;`. When the predicate is `QUser.user.roles.contains(role)` then we are getting the following exception:

```
Caused by: java.lang.IllegalArgumentException: Parameter value [ADMIN] did not match expected type [com.example.Role (n/a)]
	at org.hibernate.jpa.spi.BaseQueryImpl.validateBinding(BaseQueryImpl.java:874) ~[hibernate-entitymanager-4.3.11.Final.jar:4.3.11.Final]
	at org.hibernate.jpa.internal.QueryImpl.access$000(QueryImpl.java:80) ~[hibernate-entitymanager-4.3.11.Final.jar:4.3.11.Final]
	at org.hibernate.jpa.internal.QueryImpl$ParameterRegistrationImpl.bindValue(QueryImpl.java:248) ~[hibernate-entitymanager-4.3.11.Final.jar:4.3.11.Final]
	at org.hibernate.jpa.internal.QueryImpl$JpaPositionalParameterRegistrationImpl.bindValue(QueryImpl.java:337) ~[hibernate-entitymanager-4.3.11.Final.jar:4.3.11.Final]
	at org.hibernate.jpa.spi.BaseQueryImpl.setParameter(BaseQueryImpl.java:674) ~[hibernate-entitymanager-4.3.11.Final.jar:4.3.11.Final]
	at org.hibernate.jpa.spi.AbstractQueryImpl.setParameter(AbstractQueryImpl.java:198) ~[hibernate-entitymanager-4.3.11.Final.jar:4.3.11.Final]
	at org.hibernate.jpa.spi.AbstractQueryImpl.setParameter(AbstractQueryImpl.java:49) ~[hibernate-entitymanager-4.3.11.Final.jar:4.3.11.Final]
	at com.mysema.query.jpa.impl.JPAUtil.setConstants(JPAUtil.java:55) ~[querydsl-jpa-3.6.6.jar:?]
	at com.mysema.query.jpa.impl.AbstractJPAQuery.createQuery(AbstractJPAQuery.java:130) ~[querydsl-jpa-3.6.6.jar:?]
	at com.mysema.query.jpa.impl.AbstractJPAQuery.count(AbstractJPAQuery.java:81) ~[querydsl-jpa-3.6.6.jar:?]
	at org.springframework.data.jpa.repository.support.QueryDslJpaRepository.findAll(QueryDslJpaRepository.java:141) ~[spring-data-jpa-1.8.2.RELEASE.jar:?]
...
```

For the predicates having only simple equality test with the enum it works. Also it works with Hibernate 4.3.8.Final and before.

The problem is probably at `com.mysema.query.jpa.JPQLSerializer.visitAnyInPath(Class<?>, Operator<?>, List<? extends Expression<?>>)`. Instead of `args = ImmutableList.of(ConstantImpl.create(constant.name()), args.get(1));` there should probably be `args = ImmutableList.of(ConstantImpl.create(constant), args.get(1));`

But maybe the problem is in the Hibernate and they've broken the compatibility. Could you please check it?

timowest:
Fixes #1500

### DIFF
--- a/querydsl-jpa/pom.xml
+++ b/querydsl-jpa/pom.xml
@@ -23,7 +23,7 @@
   </scm>
   
   <properties>
-    <hibernate.version>4.3.7.Final</hibernate.version>
+    <hibernate.version>4.3.11.Final</hibernate.version>
     <hibernate.validator.version>4.3.1.Final</hibernate.validator.version>
     <eclipselink.version>2.6.0-RC1</eclipselink.version>
     <osgi.import.package>javax.persistence.*;version="[1.1,3)",${osgi.import.package.root}</osgi.import.package>
--- a/querydsl-jpa/src/main/java/com/querydsl/jpa/HQLTemplates.java
+++ b/querydsl-jpa/src/main/java/com/querydsl/jpa/HQLTemplates.java
@@ -100,12 +100,6 @@ public class HQLTemplates extends JPQLTemplates {
     }
 
     @Override
-    public boolean isEnumInPathSupported() {
-        // related : http://opensource.atlassian.com/projects/hibernate/browse/HHH-5159
-        return false;
-    }
-
-    @Override
     public boolean wrapConstant(Object constant) {
         // related : https://hibernate.onjira.com/browse/HHH-6913
         Class<?> type = constant.getClass();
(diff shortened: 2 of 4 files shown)


johnktims:
Looks good. Although this technically breaks backwards compatibility, I think it makes sense here.

querydsl-jpa/pom.xml
```diff
@@ -23,7 +23,7 @@
   </scm>
   
   <properties>
-    <hibernate.version>4.3.7.Final</hibernate.version>
+    <hibernate.version>4.3.9.Final</hibernate.version>
```
johnktims: Just so you know, `4.3.11.Final` is available.

Shredder121:
Looks good to me as well.
I'll merge this then?

timowest:
yes please ;)

Issue #1561, opened 2015-09-16
querydsl 4 does not support calling isNull() on a SQLQuery


robertandrewbain:
In querydsl 3 you could write a subquery and call `isNull()` on it. However it took me a while to figure out how to do it using querydsl 4. I think the API could be improved.

To demonstrate, I was trying to do something like this:

```java
    private SQLQuery<String> getSomeSubquery() {
        return query()
                .select(someTable.someColumn)
                .from(someTable)
                .where(someTable.someColumn.eq("someValue"))
                .distinct();
    }

    private StringExpression getSomeCaseStatementForSomeSubquery() {
        return new CaseBuilder()
                .when(getSomeSubquery().isNull())
                .then("someValue")
                .otherwise("someOtherValue");
    }
```

The problem is that `isNull()` is not present. Looking at the `isNull()` method in `SimpleExpression` I realised it was calling `Expressions.booleanOperation(Ops.IS_NULL, mixin)` which led me to look at the usages of `Ops.IS_NULL` and discover `ExpressionUtils` `public static Predicate isNull(Expression<?> left)` method, which done the trick.

However, I think it should go in `ExtendedSubQuery` as `BooleanExpression isNull(Expression<? extends T> expr)`.

timowest:
What did you try? Didn't the following work?

```java
select(select(fooExpr).from(someTable)).from(someOtherTable)
```

robertandrewbain:
Hi Timo, in response to your question, I was completely barking up the wrong tree last night - long day. I've updated the issue and its title as I believe I've got to the bottom of the problem I was experiencing. Thanks as always for your help.

timowest:
@robertandrewbain Feel free to provide a PR for that.

robertandrewbain:
@timowest no problem, I'll attempt to get that done today. I'll put in `isNotNull()` too, for completeness.

PR #1571, opened 2015-09-20
Added isNull() and isNotNull() to ExtendedSubQuery

robertandrewbain:
#1561 Perhaps I'm missing something but having looked at usages of a few of its public methods, I could not see test coverage for `FetchableSubQueryBase`. If someone can point me in the direction for test coverage for this class, I will add coverage for these new methods.


timowest:
Looks good, also without the test. Test cases could be added for any non-abstract subclass of FetchableSubQueryBase.

Shredder121:
This looks good to me!
Thanks for the PR @robertandrewbain!

johnktims:
Looks good. Thanks for the PR, @robertandrewbain.

Issue #1573, opened 2015-09-21
AbstractJPAQuery#createQuery() triggers sometimes lazy loading


wrungel:
## Problem description
`AbstractJPAQuery#createQuery()` triggers unnecessary lazy loading if one or more query paramaters contains an entity where `toString()` imelementation accesses lazy references.

### Cause
The root cause for that is the query logging and eager query parameter to string transormation.
Even if the log level is low or debugging is off query parameters will be still transformed to string:

See [AbstractJPAQuery.java](https://github.com/querydsl/querydsl/blob/master/querydsl-jpa/src/main/java/com/querydsl/jpa/impl/AbstractJPAQuery.java#L231), line 231:

    protected void logQuery(String queryString, Map<Object, String> parameters) {
        String normalizedQuery = queryString.replace('\n', ' ');
        MDC.put(MDC_QUERY, normalizedQuery);
        MDC.put(MDC_PARAMETERS, String.valueOf(parameters));
        if (logger.isDebugEnabled()) {
            logger.debug(normalizedQuery);
        }
    }

## Example

Given the fllowing JPA entities:
```java
@Entity
public class Ereignis {
    @Id @GeneratedValue(strategy = IDENTITY)
    Long id;

    @OneToMany(mappedBy = "ereignis", cascade = ALL)
    List<Child> toStringChildren = new ArrayList<>();

    @OneToMany(mappedBy = "ereignis", cascade = ALL)
    List<Child> otherChildren = new ArrayList<>();

    @Override
    public String toString() {
        logger.debug("*** calling Ereignis.toString() ...");
        return "toStringChildren.size(): " + toStringChildren.size();
    }
}
```

```java
@Entity
public class Child {
    @Id @GeneratedValue(strategy = IDENTITY)
    Long id;

    @ManyToOne
    Ereignis ereignis;
}
```

The following query causes unnecessary lazy loading of property ```toStringChildren```:

```java
new JPAQueryFactory(em).selectFrom(qEreignis)
                .where(qEreignis.eq(ereignis)).createQuery();
````
which can slow down performance drammatically.

I created a GitHub test project with a JUnit test showing the issue described here:
https://github.com/wrungel/qdsl-bugs/tree/master/create-query-triggers-lazy-loading

Run the test:

    mvn test -Dtest=LazyLoadingTest

and look on console for the output:

    DEBUG TEST:45 - *** before createQuery() called
    DEBUG TEST:39 - *** calling Ereignis.toString() ...
[...]

PR #1574, opened 2015-09-21
Move adding objects to MDC inside the `isDebugEnabled` block

Shredder121:
Fixes https://github.com/querydsl/querydsl/issues/1573

Note that removing MDC parameters that are not there is okay.
It would complicate it too much anyway.


johnktims:
Looks good to me.

PR #1578, opened 2015-09-23
Improve delegate path handling

### LINKED ISSUE #1577: Path visitor doesn't work for MongoDB queries when having subtype
Consider the classes:

```
@Entity
class A {
    @Embedded
    private B b;
}

abstract class B {
}

class B1 extends B {
    private String name;
}

class B2 extends B {
    private String key;
}

```
When I generate query DSL classes for this hierarchy I'm able to write expressions like this:
`QA.a.b.as(QB1.class).name.eq("someName")` or `QA.a.b.as(QB2.class).key.eq("someKey")`

but the paths generated by MongodbSerializer for this expressions are wrong. In version 3.6.7 I just get "name" and "key", when they should be "b.name" and "b.key" respectively.

I also tried this in version 3.6.5 after I saw this issue: https://github.com/querydsl/querydsl/issues/1427 but the results are weird, I get: "b.a.b.name"

on method `String visit(Path<?> expr, Void context)` of `MongodbSerializer` I can see how the recursion is avoided if the path type is DELEGATE which is not the case in this scenario, but even if the recursion was reached I'm not sure it'd still generate the right path. 

Any help would be greatly appreciated.

timowest:
Fixes #1577

### DIFF
--- a/querydsl-mongodb/src/main/java/com/querydsl/mongodb/MongodbSerializer.java
+++ b/querydsl-mongodb/src/main/java/com/querydsl/mongodb/MongodbSerializer.java
@@ -330,13 +330,16 @@ public abstract class MongodbSerializer implements Visitor<Object, Void> {
     public String visit(Path<?> expr, Void context) {
         PathMetadata metadata = expr.getMetadata();
         if (metadata.getParent() != null) {
+            Path<?> parent = metadata.getParent();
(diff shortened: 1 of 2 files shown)


Shredder121:
Looks good to me.

johnktims:
Looks good.

Issue #797, opened 2014-06-10
Fine grained configuration possibility for Metadata Exporter


balazs-zsoldos:
_Background:_

We decided to use Liquibase - QueryDSL  pair for our persistence layer withing a modularized technology stack instead of JPA. The first issue was to generate QueryDSL classes for tables that were described in Liquibase changelog files. Each of our module contained a set of tables. E.g.: The authorization table contains only tables that belong to the authorization business logic and the localization contains only tables that belong to that API.

We created the lqmg project (Liquibase To QueryDSL Metadata Generator). It contains more and more configuration possibilities for MetadataExporter that might be useful for others, who want to generate QueryDSL Metadata classes. We would not like to keep a set of classes up-to-date but merge back our additional features to the QueryDSL code if it is possible.

_Concept 1_:

First, we created an XML schema. Each module could contain an XML that containes class name mapping definitions. The definitions could be defined as direct **Table to ClassName** mapping and also mapping with Regular expressions. 

_Concept 2_:

We created a set of additional tables:
- querydsl_table_mapping
- querydsl_property_mapping
- querydsl_foreignkey_mapping
- querydsl_primarykey_mapping

With these tables, it is possible to define code generation configuration. While it was not possible to define an exact property name with the XMLs, it is possible to define it in these tables.

_When is it useful?:_

In many cases the naming of the tables, fields and keys are pretty ugly in the database. When the table has the project name as the prefix, when the foreign key contains abbreviations, the source code becomes unreadable. With these fine grained configuration possibilities it is possible to generate Metadata classes that can be used to write beautiful queries in the source code.

_Goal:_

First, I would like to see the enhancement possibilities in QuerzDSL SQL Metadata generator in the way that I can add our features without copy-pasting the classes and write enhancements into the functions. I would like to write our features as extensions, not by changing the original QueryDSL code.

In our implementation, I created a class called ConfigurationContainer. At each point, where naming rules have to be decided, the configurationContainer instance is called. If there is a configuration available for the table in the container, that rule is stronger than the global rules coming from the MetadataExporter class.

See: https://github.com/everit-org/osgi-lqmg/blob/master/src/main/java/org/everit/osgi/dev/lqmg/internal/ConfigurationContainer.java

In case of a pull request, I would add some javadoc to this class and refactor it a bit to have nicer code in it.

_Question:_

[...]

PR #1586, opened 2015-09-28
Fine grained configuration possibility for Metadata Exporter

attila-kiss-it:
Fixes for #797


querydsl-sql-codegen/src/main/java/com/querydsl/sql/codegen/NamingStrategy.java
```diff
@@ -143,4 +157,10 @@
      */
     String normalizeSchemaName(String schemaName);
 
+    boolean shouldGenerateClass(SchemaAndTable schemaAndTable);
```
timowest: javadocs needed here

querydsl-sql-codegen/src/main/java/com/querydsl/sql/codegen/AbstractNamingStrategy.java
```diff
@@ -122,4 +124,25 @@ public void setReservedSuffix(String reservedSuffix) {
         this.reservedSuffix = reservedSuffix;
     }
 
+    @Override
+    public String getPackage(String basePackage, SchemaAndTable schemaAndTable) {
+        return basePackage;
```
timowest: shouldn't this delegate to appendSchema?

querydsl-sql-codegen/src/main/java/com/querydsl/sql/codegen/MetaDataExporter.java
```diff
@@ -372,15 +387,19 @@ private void serialize(EntityType type) {
         try {
             String fileSuffix = createScalaSources ? ".scala" : ".java";
 
+            String schemaName = (String) type.getData().get("schema");
+            String tableName = (String) type.getData().get("table");
+            SchemaAndTable schemaAndTable = new SchemaAndTable(schemaName, tableName);
```
timowest: the SchemaAndTable instance which is created in the previous could be given as an argument, then the normalized schema and table name would be used in both cases.

timowest:
This looks really good now, I found one thing could be still improved.

querydsl-sql-codegen/src/main/java/com/querydsl/sql/codegen/MetaDataExporter.java
```diff
@@ -123,9 +123,10 @@ public MetaDataExporter() { }
     protected EntityType createEntityType(@Nullable String schemaName, String tableName,
```
timowest: provide SchemaAndTable as argument

timowest:
Looks good now. @attila-kiss-it Thanks a lot for the contribution!

Shredder121:
Sorry for the delay, thank you @attila-kiss-it for the contribution.

PR #1587, opened 2015-09-28
Bump codegen version

timowest:


Shredder121:
These look good to me.
Since the PRs bump codegen to something that #1539 needs.

johnktims:
Looks good.

PR #1591, opened 2015-09-30
Querydsl 4.0.5

timowest:


Shredder121:
How is the new versions plugin working?

Shredder121:
It looks like all modules have their versions incremented.
That is good.

timowest:
Works fine, but generateBackupPoms=false should be moved to the plugin config in the pom

PR #1595, opened 2015-10-01
Release script

timowest:


release.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+git checkout master
+git pull
+mvn versions:set -DgenerateBackupPoms=false
+mvn clean deploy -Dtest=X -Dgpg.skip=false
```
johnktims: What does `-Dtest=X` do?

timowest: It skips the tests, but I'll change it to `-DskipTests`.

johnktims:
Looks fine to me.

Shredder121:
Oops, looks fine to me.

PR #1596, opened 2015-10-01
Obtain Connection in SQLQuery when needed

### LINKED ISSUE #1581: Using SQLQueryFactory's query(), subqueries are assigned connections and never close them
When using `SQLQueryFactory`'s `query()` method to create a subquery, each subquery is assigned a new connection, as the connection object is assigned when `query()` is called. Furthmore, using a `SQLCloseListener` does not close subqueries, I presume because a terminator E.G. `.list()` etc. has not been called on it. This caused my system test environment to grind to a halt when the querydsl 4 upgrade was pushed up.

After realising what was happening, I discovered `SQLExpressions.select` which will "Create a new detached SQLQuery instance with the given projection". After some debugging I realised this creates a `SQLQuery` instance with a null connection - so far so good. I assume detached means not attached to a connection. If so `SQLQueryFactory`'s  `public <T> SQLQuery<T> selectFrom(RelationalPath<T> expr)` method's documentation is wrong.

So for me, the issue here is that if you're using a `SQLQueryFactory`, creating a subquery using `query()` looks sensible and will work, but will swallow a connection. It feels like a connection should not be getting assigned on calling `query()`, surely the connection is not required until calling a terminator?

Secondly, `SQLCloseListener ` should close these connections.

timowest:
Fixes #1581

### DIFF
--- a/querydsl-sql/src/main/java/com/querydsl/sql/AbstractSQLQuery.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/AbstractSQLQuery.java
@@ -24,6 +24,7 @@ import java.util.List;
 import java.util.Map;
 
 import javax.annotation.Nullable;
+import javax.inject.Provider;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -56,7 +57,10 @@ public abstract class AbstractSQLQuery<T, Q extends AbstractSQLQuery<T, Q>> exte
     private static final QueryFlag rowCountFlag = new QueryFlag(QueryFlag.Position.AFTER_PROJECTION, ", count(*) over() ");
 
     @Nullable
-    private final Connection conn;
+    private Provider<Connection> connProvider;
+
+    @Nullable
+    private Connection conn;
 
     protected SQLListeners listeners;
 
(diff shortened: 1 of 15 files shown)


querydsl-sql/src/main/java/com/querydsl/sql/AbstractSQLQuery.java
```diff
@@ -81,6 +85,17 @@ public AbstractSQLQuery(@Nullable Connection conn, Configuration configuration,
         this.useLiterals = configuration.getUseLiterals();
     }
 
+    public AbstractSQLQuery(Provider<Connection> conn, Configuration configuration) {
+        this(conn, configuration, new DefaultQueryMetadata());
+    }
+
+    public AbstractSQLQuery(Provider<Connection> conn, Configuration configuration, QueryMetadata metadata) {
+        super(new QueryMixin<Q>(metadata, false), configuration);
+        this.connProvider = conn;
```
Shredder121: Maybe name these `connProvider`?

timowest:
Shall we deprecate some constructors? Like those which take a SQLTemplates argument?

timowest:
This PR breaks source level backwards compatibility, as can be seen in the `(Connection) null` casts.

Shredder121:
It at least looks good, but do you want to merge this for the upcoming release?

timowest:
Yes, I think this can be merged for the next release.

johnktims:
Looks good to me.

PR #1597, opened 2015-10-02
I1594 linked hash set in default mapper

elser:


Shredder121:
This looks already good to me. :D

elser:
Thanks :)

johnktims:
Looks good to me. Thanks for the contribution, @elser.

timowest:
I'd also change `AbstractMapper.getColumns()` to use `LinkedHashMap`. Then the order should be the same as `RelationalPath.getColumns()`. RelationalPath uses internally a LinkedHashMap, so that's already covered.

elser:
That should be it, I suppose?

timowest:
Looks good, could you maybe rebase your commits into a single one?

elser:
Sorry! Squashed.

PR #1599, opened 2015-10-03
Create test databases in target directory

johnktims:
... instead of in the user's home directory.


timowest:
Looks good.

Shredder121:
Looks good to me as well.
Thanks for the PR @johnktims!

PR #1600, opened 2015-10-07
Preserve column order

### LINKED ISSUE #1594: DefaultMapper should return maps with predictable order
I've came accross this issue while writing unit tests with asserts on SQL query strings. I've noticed that between multiple runs of the same test, the order of fields in generated SQLs varied. This rendered my tests flaky. 
Also, because of this, we could possibly be expecting different SQL queries for the same lists of fields produced on different JVMs in the production environment. This in turn would decrease database's ability to skip parsing similar statements.
The quickest fix would be to replace HashMap with a LinkedHashMap in DefaultMapper.

Thanks in advance

johnktims:
Fixes #1594
Backport of #1597

### DIFF
--- a/querydsl-sql/src/test/java/com/mysema/query/sql/dml/SQLInsertClauseTest.java
+++ b/querydsl-sql/src/test/java/com/mysema/query/sql/dml/SQLInsertClauseTest.java
@@ -1,11 +1,13 @@
 package com.mysema.query.sql.dml;
 
+import static org.junit.Assert.assertEquals;
+
+import org.junit.Test;
+
 import com.google.common.collect.ImmutableList;
-import com.mysema.query.sql.KeyAccessorsTest.QEmployee;
 import com.mysema.query.sql.SQLBindings;
 import com.mysema.query.sql.SQLTemplates;
-import org.junit.Test;
-import static org.junit.Assert.assertEquals;
+import com.mysema.query.sql.domain.QEmployee;
 
 public class SQLInsertClauseTest {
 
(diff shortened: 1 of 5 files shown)


querydsl-sql/src/test/java/com/mysema/query/sql/dml/SQLInsertClauseTest.java
```diff
@@ -20,4 +22,17 @@ public void GetSQL() {
         assertEquals(ImmutableList.of(1), sql.getBindings());
     }
 
+    @Test
+        public void GetSQLWithPreservedColumnOrder() {
```
timowest: indentation is off here

Shredder121: Haha, you think so?

Shredder121: @johnktims do you want me to cherry-pick the original commit?

johnktims: Yes.

Shredder121:
There you go!

johnktims:
Thanks :)

Shredder121:
One second.

querydsl-sql/src/test/java/com/mysema/query/sql/dml/SQLInsertClauseTest.java
```diff
@@ -20,4 +20,17 @@ public void GetSQL() {
         assertEquals(ImmutableList.of(1), sql.getBindings());
     }
 
+    @Test
+    public void GetSQLWithPreservedColumnOrder() {
+        com.querydsl.sql.domain.QEmployee emp1 = new com.querydsl.sql.domain.QEmployee("emp1");
```
Shredder121: Do you want me to update the fully qualified name or update the import of `QEmployee`?

timowest:
This could be maybe also backported in this PR in https://github.com/querydsl/querydsl/pull/1604

Otherwise the order isn't defined, just dependent on less variables than before.

Shredder121:
That's fine, I'll do that.

Issue #1580, opened 2015-09-24
JPA - Generate TypeSafe alias from QClass properties


darkbauer:
Hello QueryDsl team,

I would like to put a request for you, about aliasing. 

For example let's say I have this class: 
```java
public Class Customer {
    private Integer amount;

    @Transient
    @QueryType(PropertyType.SIMPLE)
    private Integer irr;
}
```
```java
QCustomer customer = new QCustomer("customer");

queryFactory()
.select(
   Projections.bean(Customer.class, customer.amount.as("irr"))
.from(customer)
.fetch();
```
This query works, and put amount value in "irr" property.

But my problem is that this aliasing is not typesafe : If I rename "irr" property in Customer class, I will not have any compilation issue.

So I tried to do :  
```java
queryFactory()
.select(
   Projections.bean(Customer.class, customer.amount.as(customer.irr))
.from(customer)
.fetch();
```
But this doesn't work, which is a problem. 

Is there a way to provide a typesafe way to name an alias ? It's very important, especially when projecting query results into a transient property.

Thanks !

timowest:
Here is a related issue https://github.com/querydsl/querydsl/issues/1572.

Maybe we can just fix the aliasing behaviour to make your second example work?

darkbauer:
Hello timo,

Yes, I think that if you can fix the aliasing behaviour to work like the second example, it would be really great and useful !

Thanks again for your quick answer.

PR #1601, opened 2015-10-07
Improve alias serialization

### LINKED ISSUE #1572: SQL aliasing contains tablename when path is used
I've used to do aliasing by reusing the original path `SFoo.bar.nullif(0).as(SFoo.bar)`. This nowadays leads to `NULLIF("foo"."bar", 0 ) AS "foo"."bar"` but it used to be  `NULLIF("foo"."bar", 0 ) AS "bar"`.

The new form doesn't work with neither H2 or MySQL and I guess that other databases do not appreciate it either. 

As a workaround I used `SFoo.bar.nullif(0).as(SFoo.bar.getMetadata().getName())` which turns the path into a variable and hence works as expected. 

I guess this has been done to support aliasing in some other parts of the query (`JOIN` maybe), but it doesn't work as `SELECT` anymore. I see that `SQLSerializer` has gimmicks involving `Ops.ALIAS` so could it be possible to intervene there how the result is actually printed out

timowest:
Fixes #1572 and #1580

### DIFF
--- a/src/main/resources/querydsl_checks_base.xml
+++ b/src/main/resources/querydsl_checks_base.xml
@@ -7,8 +7,9 @@
   <property name="charset" value="UTF-8"/>
   <property name="fileExtensions" value="java, properties, xml"/>
 
-  <!-- maximum file length 2000 lines -->
-  <module name="FileLength"/>
+  <module name="FileLength">
+    <property name="max" value="2500"/>
+  </module>
 
   <!-- check that there are no tab characters in the file -->
   <module name="FileTabCharacter">
--- a/src/main/resources/querydsl_checks_base.xml
+++ b/src/main/resources/querydsl_checks_base.xml
@@ -7,9 +7,8 @@
   <property name="charset" value="UTF-8"/>
   <property name="fileExtensions" value="java, properties, xml"/>
 
-  <module name="FileLength">
-    <property name="max" value="2500"/>
-  </module>
+  <!-- maximum file length 2000 lines -->
+  <module name="FileLength"/>
 
   <!-- check that there are no tab characters in the file -->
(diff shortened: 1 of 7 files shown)


src/main/resources/querydsl_checks_base.xml
```diff
@@ -90,6 +91,11 @@
   </module>
 
   <module name="SuppressWarningsFilter"/>
+  <module name="SuppressionCommentFilter">
+    <property name="offCommentFormat" value="CSOFF\: ([\w\|]+)"/>
+    <property name="onCommentFormat" value="CSON\: ([\w\|]+)"/>
```
johnktims: I'd prefer to use "checkstyle" instead of "cs".

Shredder121: So the same as the normal CHEKSTYLE:OFF?
Something like `CHECKSTYLE:OFF: ([\w|]+)`?

Shredder121: Shall I do `CHECKSTYLERULE:OFF: ([\w|]+)`?

johnktims: Adding RULE might be too verbose but either one is fine with me.

Shredder121: it already clashes with turning checkstyle off entirely: `<[[CHECKSTYLE:OFF]]>: ([\w]+) `.
Also, since it IS used to turn off a rule, it is descriptive enough.
And it won't be used that much, so verbosity I think is not that bad?

johnktims: You're right. Adding RULE makes sense.

Shredder121:
This looks good to me.

PR #1602, opened 2015-10-07
Make mapping configuration available in ant exporter

### LINKED ISSUE #1510: AntMetadataExporter documentation: numericMappings, typeMappings, customTypes not documented
Hi all,

I've try to set numericMappings, typeMappings and customTypes options to AntMetadataExporter, without success.

In documentation is not clear how to do it.

Best regards,

Agharta

timowest:
Fixes #1510


querydsl-sql-codegen/src/main/java/com/querydsl/sql/codegen/support/RenameMapping.java
```diff
@@ -23,9 +23,9 @@
  */
 public class RenameMapping implements Mapping {
 
-    String fromSchema, fromTable, fromColumn;
+    public String fromSchema, fromTable, fromColumn;
```
johnktims: Why are these fields public now?

timowest: the public fields are more an intermediate form, the goal is a form that can be populated by both Maven and Ant, haven't looked into Ant population yet.

Shredder121: Also the setters of the property you configure.
http://codefeed.com/tutorial/ant_config.html

querydsl-sql-codegen/src/main/java/com/querydsl/sql/codegen/ant/AntMetaDataExporter.java
```diff
@@ -644,4 +691,30 @@ public String getSourceEncoding() {
...
+        typeMappings.add(mapping);
+        return mapping;
+    }
+
+    public NumericMapping createNumericMapping() {
+        NumericMapping mapping = new NumericMapping();
+        numericMappings.add(mapping);
+        return mapping;
+    }
+
```
Shredder121: Maybe document these?
These were necessary for ANT's element syntax right?
Maybe you can do a shoutout on the Google group if this would also work for Gradle?

timowest: These is purely for Ant and so weird, that I doubt that any other build tool will use the same pattern.

Shredder121:
>As attributes are mapped to setXXX() style methods, nested elements were mapped to either addXXX() or createXXX() methods. The former allowed Ant to create the object corresponding to the nested element and pass it to the Task, while the latter gives the task responsibility for object creation. Ant then configures the created object from the build file. In general, tasks should use the addXXX() form where possible.

I think add would make it more explicit, since you _are_  actually adding the elements.

querydsl-maven-plugin/src/main/java/com/querydsl/maven/AbstractMetaDataExportMojo.java
```diff
@@ -33,6 +33,9 @@
 import com.querydsl.sql.codegen.DefaultNamingStrategy;
 import com.querydsl.sql.codegen.MetaDataExporter;
 import com.querydsl.sql.codegen.NamingStrategy;
+import com.querydsl.sql.codegen.support.NumericMapping;
+import com.querydsl.sql.codegen.support.RenameMapping;
+import com.querydsl.sql.codegen.support.TypeMapping;
```
Shredder121: Are these imports necessary?
They seem to be the only thing that changed in this file.
Or am I missing something?

timowest: The *Mapping classes have been moved to querydsl-sql-codegen, that's why they are now imported.

Shredder121: Oh yeah.
I didn't have my IDE, and didn't pay attention to the move.

johnktims:
Looks good to me.

PR #1604, opened 2015-10-09
Improvements to mapper order

timowest:
After this change both mappers will use column order in the resulting binding map


Shredder121:
This looks good to me.
It's also applied in #1600.

johnktims:
Looks good to me.

PR #1605, opened 2015-10-11
Fix projection wrapping

timowest:


johnktims:
@timowest What was the issue here? I don't think I completely understand the rationale behind the change.

timowest:
the serializer in Querydsl SQL expects flattened FactoryExpressions. The flattening happens in the SQLQueryMixin, here we are doing it explicitly since we use QueryMetadata directly.

Shredder121:
Then to me it looks good.

johnktims:
Agreed.

PR #1606, opened 2015-10-11
Improve SQLServer tests for JPA

timowest:


querydsl-jpa/src/test/resources/com/querydsl/jpa/testutil/mssql.properties
```diff
@@ -1,5 +1,5 @@
 ## MSSQL
-hibernate.dialect=org.hibernate.dialect.SQLServerDialect
+hibernate.dialect=org.hibernate.dialect.SQLServer2005Dialect
```
johnktims: How come you used `2005` here but `2008` in [Mode#GetSQLTemplates()](https://github.com/querydsl/querydsl/pull/1606/files#diff-38042974cf7ecb54afc1f8a1cf7a718cR22)?

johnktims:
I'm still seeing errors on appveyor when I run with sqlserver 2008 and 2012. Which version are you using?

querydsl-jpa/src/test/resources/com/querydsl/jpa/testutil/mssql.properties
timowest: Is there a Hibernate dialect for 2008? I might have missed that.

timowest:
I used 2008, but I didn't fix all the issues. Some might be caused by issues with the Hibernate dialect, like substring related failures.

johnktims:
Are all the tests passing for you locally now? I pushed your changes to my fork and switched my appveyor configuration to use mssql 2008, but I'm still [seeing errors](https://ci.appveyor.com/project/johnktims/querydsl/build/236#L742).

timowest:
Still failures, but this PR improves the situation.

timowest:
I left some of the issues unfixed, because they could be better fixed by looking if the Hibernate dialect could be improved, instead of excluding tests from test execution.

But that would be something for a followup PR.

PR #1611, opened 2015-10-16
Get rid of query level normalization

### LINKED ISSUE #1608: Normalization applied to String Literals
I tried to execute the following query:
SELECT * FROM POSITIONS WHERE ID='XXX-01-999-12345678-ZXB-1234'

Unfortunately the Normalizer of QueryDSL kicked in and changed it to:
SELECT * FROM POSITIONS WHERE ID='XXX-12346678-ZXB-1234'

The Query thusfailed to give me the expected results.

Why is it trying to normalize anything within String Literals?

timowest:
Fixes #1608


johnktims:
I really like this change, but maybe we should deprecate the functionality and allow it to remain for a few releases before removing it.

timowest:
Normalization is a non-public class, so removing it doesn't violate backwards compatibility. Most of the cases should be doable with a few changes to the templating logic.

timowest:
We currently support these parameter transformations
* {%0} - ends with like
* {%%0} - ends with like lowercase
* {0%} - starts with like
* {0%%} - starts with like lowercase
* {%0%} - contains
* {%%0%%} - contains lowercase
* {0l} - lowercase
* {0u} - uppercase
* {0s} - as string

the following additions might be needed
* arg1 +/- arg2
* arg +/- const
* const +/- arg

querydsl-core/src/main/java/com/querydsl/core/types/TemplateFactory.java
```diff
@@ -223,7 +229,11 @@ public Template create(String template) {
                         }
                         break;
                 }
-                if (asString) {
+                if (operand != null) {
+                    Operator operator = OPERATORS.get(operand.substring(0, 1));
+                    int index2 = Integer.parseInt(operand.substring(1));
```
Shredder121: I don't quite like these `substring`s, it's these that I removed in the first place. :/
What's your opinion @johnktims?

timowest: more regex groups?

querydsl-sql/src/test/java/com/querydsl/sql/SelectBase.java
```diff
@@ -142,6 +141,8 @@ public void Arithmetic() {
         NumberExpression<Integer> three = Expressions.numberTemplate(Integer.class, "(3.0)");
         NumberExpression<Integer> four = Expressions.numberTemplate(Integer.class, "(4.0)");
         arithmeticTests(one, two, three, four);
+        // the following one doesn't work with integer arguments
+        assertEquals(2, firstResult(four.multiply(one.divide(two))).intValue());
```
Shredder121: What do you mean by that?

timowest: `4 * (1 / 2)` is evaluated as zero because of integer conversion in between.

timowest: the results are consistent over databases, and the difference comes from missing normalization across operation boundaries.

Shredder121:
For reference and debugging/comparisons, the old regex:
`\{(%?%?)(\d+)([slu%]?%?)\}`

querydsl-sql/src/test/java/com/querydsl/sql/SelectBase.java
Shredder121: I see, thanks for the explanation.

querydsl-jpa/src/main/java/com/querydsl/jpa/JPQLTemplates.java
```diff
@@ -99,12 +99,12 @@ protected JPQLTemplates(char escape, QueryHandler queryHandler) {
         add(Ops.MATCHES, "{0} like {1}  escape '" + escape + "'"); // TODO : support real regexes
         add(Ops.MATCHES_IC, "{0} like {1} escape '" + escape + "'"); // TODO : support real regexes
         add(Ops.LOWER, "lower({0})");
-        add(Ops.SUBSTR_1ARG, "substring({0},{1s}+1)", Precedence.ARITH_LOW);
-        add(Ops.SUBSTR_2ARGS, "substring({0},{1s}+1,{2s}-{1s})", Precedence.ARITH_LOW);
+        add(Ops.SUBSTR_1ARG, "substring({0},{1+'1'})", Precedence.ARITH_LOW);
+        add(Ops.SUBSTR_2ARGS, "substring({0},{1+'1'},{2-1})", Precedence.ARITH_LOW);
```
Shredder121: Is this line the same as the old one?
I can't quite tell.

timowest: The `s` post processing suffix was mostly used to get more stuff for the normalization to convert. Now that the conversion is embedded in the templates it's not anymore necessary.

Shredder121: Gotcha.

querydsl-core/src/main/java/com/querydsl/core/support/SerializerBase.java
```diff
@@ -170,8 +168,12 @@ public void setAnonParamPrefix(String prefix) {
         this.anonParamPrefix = prefix;
     }
 
+    /**
+     * Not used anymore
```
johnktims: Maybe we should elaborate a little more here and explain why the `normalize` flag isn't used anymore and what the new behavior is so users know what's going on when they upgrade.

querydsl-core/src/test/java/com/querydsl/core/types/TemplateTest.java
```diff
@@ -31,6 +31,20 @@ public void test() {
...
+    public void operation() {
+        TemplateFactory factory = new TemplateFactory('\\');
+        match("[0 ADD 1]",               factory.create("{0+1}"));
+        match("[0 DIV 1]",               factory.create("{0/1}"));
+    }
+
+    @Test
+    public void operationConst() {
+        TemplateFactory factory = new TemplateFactory('\\');
+        match("[0 ADD 1]",               factory.create("{0+'1'}"));
```
johnktims: How come we differentiate between `1` and `'1'`?

querydsl-core/src/main/java/com/querydsl/core/support/SerializerBase.java
Shredder121: Or put that text in a `@deprecated` tag, since `// not used anymore` is already on the following line, once is enough.

querydsl-core/src/test/java/com/querydsl/core/types/TemplateTest.java
Shredder121: The separate `1` is an argument with _index 1_.
The `'1'` is a _constant 1_, which could also have been `'33'` for example.

querydsl-core/src/main/java/com/querydsl/core/types/Template.java
```diff
@@ -33,6 +38,8 @@
 
     private static final long serialVersionUID = -1697705745769542204L;
 
+    private static final Set<Operator> CONVERTABLES = ImmutableSet.<Operator>of(Ops.ADD, Ops.SUB);
```
Shredder121: Could you use `Sets.immutableEnumSet`?

timowest: Ops is an enum, but Operator isn't, so I guess not.

timowest:
Caching of Element instances could be maybe done in a separate PR.

Shredder121:
That's fine.

Shredder121:
How does the Template syntax cope with multiple elements?
Does it need to be documented?

timowest:
You mean operations with more than two operands? That's not (yet) supported, but can be added later when needed.

Shredder121:
But does it work now? Otherwise you can maybe get inconsistent results.

timowest:
Those are not parsed as operations and treated as normal text.

Shredder121:
I figured, then it's okay.

Shredder121:
I think we're there then?

PR #1617, opened 2015-10-22
Remove GenericExporterTest from jpa-codegen

johnktims:
This test doesn't really belong here, and `GenericExporter` is pretty well tested in other modules.


Shredder121:
Yeah, and furthermore, there aren't any _extra_ assertions, so to me it doesn't have added value.

To me this looks good.
Or do you have something against this @timowest?

timowest:
Nothing against it. Looks good ;)

PR #1620, opened 2015-10-25
Fixed detach for Collections.

Tillerino:
When automatically detaching results, if the result is a Collection, PersistenceManager.detachCopyAll(T...) was called instead of PersistenceManager.detachCopyAll(Collection) causing a ClassNotPersistableException. See unit test.


Shredder121:
Thanks for the PR!
@johnktims @timowest Just letting you know that this looks good to me.

timowest:
Looks good.

PR #1622, opened 2015-10-27
Extend Hibernate dialects

timowest:
Introduces Hibernate dialects which extend the official database specific Hibernate dialects with Querydsl based function definitions with date/time operations and other functionality that is not correctly defined in the original dialects.


Shredder121:
There is a shorthand for registering functions.
You could use `getFunctions().putAll(DialectSupport.createFunctions(templates))`, since the only thing the registerFunction does differently is converting the name to lowercase.

We already use lowercase function names, so it could be used.
Up to you whether you incorporate it though.

timowest:
That's a great suggestion, I will try it.

Shredder121:
Looks good to me.
Restarted the Scala build.

johnktims:
Looks good.

PR #1626, opened 2015-10-31
Replace list() with select().fetch() in docs

timowest:


querydsl-docs/src/main/docbook/en-US/content/general/creating-queries.xml
```diff
@@ -230,8 +230,8 @@ QBankAccount bankAccount = account.as(QBankAccount.class);
     <para>Literals can be selected by referring to them via Constant expressions. Here is a simple example </para>
 
     <programlisting language="java"><![CDATA[
-query.list(Expressions.constant(1),
-           Expressions.constant("abc"));
+query.select(Expressions.constant(1),
+             Expressions.constant("abc"));
```
Shredder121: Did you intentionally omit fetch?
(It's not necessary to have it, since it's only talking about the select part, but just wondering).

querydsl-docs/src/main/docbook/en-US/content/general/result-handling.xml
```diff
@@ -120,7 +120,7 @@ class Customer {
     <programlisting language="java"><![CDATA[
 QCustomer customer = QCustomer.customer;
 JPQLQuery query = new HibernateQuery(session);
-List<Customer> dtos = query.from(customer).list(QCustomer.create(customer.id, customer.name));
+List<Customer> dtos = query.select(QCustomer.create(customer.id, customer.name)).from(customer).fetch();
```
Shredder121: Why do you wrap some lines, and not these?

querydsl-docs/src/main/docbook/en-US/content/general/creating-queries.xml
timowest: Here it's on purpose, since this also applies to subqueries.

querydsl-docs/src/main/docbook/en-US/content/general/result-handling.xml
timowest: I will unify that.

Shredder121:
I think this looks good, and more importantly is pretty vital to merge.
@johnktims what do you think?

johnktims:
Looks good.

Issue #1624, opened 2015-10-30
MapPath problem with nested entitys and JPA / Hibernate 


Kernle32DLL:
Good morning,

I toyed around with querydsl the past few days, and ran into a strange problem with a Map<String, String>. Not sure if this is an error on querydsls or my side, so I'm posting it here.

Hibernate: 4.3.11.Final
Querydsl: 3.6.8

I have my entity set up as follows:

```java
@Entity
public class EvaluationCentreContext {
  [...]

  @ElementCollection(fetch=FetchType.EAGER)
  @CollectionTable(name = "evaluationcentrecontext_attribute",
      joinColumns = {@JoinColumn(name = "context_id")})
  @MapKeyColumn(name = "attribute_key")
  @Column(name = "attribute")
  private Map<String, String> attributes;

  [...]
}
```

The attributes map is backed by an MSSQL table as follows:

![clipboard01](https://cloud.githubusercontent.com/assets/1338408/10844257/8765a612-7efe-11e5-88e9-f5652ed1b95a.png)

The Q... class is generated correctly, too:

```java
@Generated("com.mysema.query.codegen.EntitySerializer")
public class QEvaluationCentreContext extends EntityPathBase<EvaluationCentreContext> {
    [...]
    public final MapPath<String, String, StringPath> attributes = this.<String, String, StringPath>createMap("attributes", String.class, String.class, StringPath.class);
    [...]
}
```

Now, I want to order my EvaluationCentreContext objects via an attribute in the attributes map. This works as intendend:

```java
evaluationCentreContextRepository.findAll(null, QEvaluationCentreContext.evaluationCentreContext.attributes.get("bla123").asc());
```

Which results in

```sql
select
	[...]
from
	evaluationcentrecontext evaluation0_ 
inner join
	evaluationcentrecontext_attribute attributes1_ 
		on evaluation0_.auto_id=attributes1_.context_id 
		and (
			attributes1_.attribute_key='bla123'
		) 
order by
	attributes1_.attribute asc
```

However, as soon as I introduce a new entity, that has an reference to said EvaluationCentreContex, the query breaks...

```java
@Entity
public class RadiationExposedActivity{
  [...]


  @OneToOne(optional = false)
  @JoinColumn(name = "context_id")
  private EvaluationCentreContext evaluationCentreContext;
    [...]
}
```

```java
@Generated("com.mysema.query.codegen.EntitySerializer")
[...]

PR #1623, opened 2015-10-29
Fix map access in order

timowest:
Fixes https://github.com/querydsl/querydsl/issues/1624

PR #1629, opened 2015-11-01
Fix map access in order

timowest:
Fixes https://github.com/querydsl/querydsl/issues/1624
backport of https://github.com/querydsl/querydsl/pull/1623

### DIFF
--- a/querydsl-core/src/main/java/com/mysema/query/types/ExpressionUtils.java
+++ b/querydsl-core/src/main/java/com/mysema/query/types/ExpressionUtils.java
@@ -13,9 +13,9 @@
  */
 package com.mysema.query.types;
 
-import javax.annotation.Nullable;
-
 import java.util.*;
+
+import javax.annotation.Nullable;
 
 import com.google.common.collect.ImmutableList;
 import com.mysema.query.BooleanBuilder;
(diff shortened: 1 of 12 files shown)


querydsl-core/src/main/java/com/mysema/query/types/ExpressionUtils.java
```diff
@@ -43,6 +43,193 @@ private UnderscoreTemplates() {
...
+        return new PathImpl<T>(type, variable);
+    }
+
+    /**
+     * Create a new Path expression
+     *
+     * @param type type of expression
+     * @param parent parent path
+     * @param property property name
+     * @return property path
```
Shredder121: Why does this say `property path` and the other places is `path expression`?
I think `path expression` sounds more in line with `ExpressionUtils`.

querydsl-core/src/main/java/com/mysema/query/types/ParametrizedPathImpl.java
```diff
@@ -0,0 +1,37 @@
...
+        super(type, metadata);
+        this.parameterTypes = parameterTypes;
+    }
+
+    @Override
+    public Class<?> getParameter(int index) {
+        return parameterTypes[index];
+    }
+
+}
```
Shredder121: Could you add a newline?

querydsl-core/src/main/java/com/mysema/query/types/ExpressionUtils.java
```diff
@@ -43,6 +43,193 @@ private UnderscoreTemplates() {
...
+
+    /**
+     * Create a new Operation expression
+     *
+     * @param type type of expression
+     * @param operator operator
+     * @param args operation arguments
+     * @return operation expression
+     */
+    public static <T> Operation<T> operation(Class<? extends T> type, Operator operator,
```
Shredder121: `Operator` is generic.

Shredder121:
I think it now looks good.

PR #1632, opened 2015-11-01
Improve docs

### LINKED ISSUE #1311: Add a link from querydsl-sql documentation to GenericExporter
I wanted to use querydsl-maven-plugin:export to generate code used only by unit tests (meaning, I wanted to export into `test/generated-test-sources` instead of `test/generated-sources`. After much digging, I discovered http://www.querydsl.com/static/querydsl/3.6.3/reference/html/ch03s03.html.

Problem is, the original querydsl-sql documentation (found at http://www.querydsl.com/static/querydsl/3.6.3/reference/html/ch02s03.html) does not link to the former so new users will get stuck.

Please consider linking from the querydsl-sql documentation to the paragraph about GenericExporter (telling users they should look here for advanced functionality).

timowest:
Fixes #1311

### DIFF
--- a/querydsl-docs/src/main/docbook/en-US/content/tutorials/sql.xml
+++ b/querydsl-docs/src/main/docbook/en-US/content/tutorials/sql.xml
@@ -85,8 +85,7 @@
 
     <para>
       Use the goal <emphasis>test-export</emphasis>
-      to add the targetFolder as a test compile source root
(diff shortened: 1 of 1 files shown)


johnktims:
Looks fine to me.

Shredder121:
Agreed.

PR #1637, opened 2015-11-03
Fix Column handling

### LINKED ISSUE #1635: NativeSQLSerializer makes invalid assumption about column name of projection args
The serialize(QueryMetadata metadata, boolean forCountRow) method contains the following code (multiple times) within the method body:

if (path.getAnnotatedElement().isAnnotationPresent(Column.class)) {
    aliases.put(path, path.getAnnotatedElement().getAnnotation(Column.class).name());
}

This code is problematic because the name field of the Column annotation is not compulsory. From a JPA perspective it defaults to the field name (see Javadoc), but from a Java perspective it simply defaults to an empty String.

Issue discovered when running on JPA/Hibernate.
If a JPASQLQuery is created for a query with a result type which references an entity which does not explicitly provide all the column names, then Hibernate throws an obscure internal error (during CustomLoader.autoDiscoverTypes because one of the "queryReturn" objects has an empty String for an alias).

example code causing error:
jpaSqlQuery.list(QMyEntity.myTable.data)
where QMyEntity is the metamodel information for entity bean MyEntity

Workaround is to ensure that entity beans explicitly define the name attribute for every column annotation.

timowest:
Fixes #1635

### DIFF
--- a/querydsl-jpa/src/main/java/com/querydsl/jpa/NativeSQLSerializer.java
+++ b/querydsl-jpa/src/main/java/com/querydsl/jpa/NativeSQLSerializer.java
@@ -118,7 +118,12 @@ public final class NativeSQLSerializer extends SQLSerializer {
                 projection = ExpressionUtils.as(projection, alias);
                 modified = true;
             } else if (path.getAnnotatedElement().isAnnotationPresent(Column.class)) {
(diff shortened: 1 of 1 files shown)


Shredder121:
Looks good together with the backport.

johnktims:
Looks good.

PR #1636, opened 2015-11-03
Add keywords

timowest:
Fixes #1114

PR #1644, opened 2015-11-07
Add keywords 2

Shredder121:
Follow up to #1636.

It would really help if each file also has a comment in it where the keywords were found.
@timowest could you maybe add the sources for the keywords?


timowest:
Nice! Shall I add the sources as a new commit in this PR?

Shredder121:
Yes, that would be great!

johnktims:
Looks good to me.

PR #1645, opened 2015-11-08
Remove Java 7 from Travis build

timowest:


Shredder121:
What is the motivation?

timowest:
I felt it would be good for consistency and maybe to avoid less unnecessary use of travis resources.

But maybe this doesn't make that much sense for a patch or minor release.

Issue #1308, opened 2015-04-06
Clarify schemaPattern


cowwoc:
Please clarify the `schemaPattern` documentation found at  http://www.querydsl.com/static/querydsl/latest/reference/html/ch02s03.html. It's not clear what syntax the "pattern" is in (is it a regular expression, glob, etc?) or how to specify multiple patterns (e.g. give me pattern1 or pattern2).

PR #1648, opened 2015-11-10
Clarify pattern format

johnktims:
Backport of #1308


Shredder121:
Looks good already, since it's just documentation.

PR #1526, opened 2015-09-03
Remove metadata reset usage

timowest:
Fixes #1524

PR #1651, opened 2015-11-12
Remove metadata reset comment in docs

johnktims:
This no longer applies. See #1526 for context.


timowest:
Looks good

Shredder121:
Looks good to me as well.

PR #1652, opened 2015-11-12
Remove metadata reset comment in docs

johnktims:
This no longer applies. See #1537 for context.


Shredder121:
Merging this as well.

PR #1654, opened 2015-11-12
Add configurable field/accessor usage

### LINKED ISSUE #1650: Codegen : Ability to configure useGetters/useFields
By default, fields and getters are analyzed during code generation, but I need fields only generation. I found an option for that in `DefaultConfiguration`,  but there is no way to set it using maven plugin. I also didn't find a workaround for now.

Shredder121:
Fixes #1650.

### DIFF
--- a/querydsl-docs/src/main/docbook/en-US/content/general/codegen.xml
+++ b/querydsl-docs/src/main/docbook/en-US/content/general/codegen.xml
@@ -196,7 +196,14 @@ import com.mysema.query.annotations.Config;
             <entry>querydsl.excludedClasses</entry>
             <entry>comma separated list of class names to be excluded from code generation (default: none)</entry>
           </row>
-
+          <row>
+            <entry>querydsl.useFields</entry>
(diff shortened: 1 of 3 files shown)


querydsl-docs/src/main/docbook/en-US/content/general/codegen.xml
```diff
@@ -196,7 +196,14 @@ import com.mysema.query.annotations.Config;
             <entry>querydsl.excludedClasses</entry>
             <entry>comma separated list of class names to be excluded from code generation (default: none)</entry>
           </row>
-
+          <row>
+            <entry>querydsl.useFields</entry>
+            <entry>set whether fields are used as metadata source (default: true)</entry>
```
andrei-ivanov: both are true? I see bellow <querydsl.useFields>false</querydsl.useFields>

Shredder121: Yes, by default we process everything.
However with some submodules we inspect the presence of relevant annotations first.

PR #1658, opened 2015-11-16
Replace Puppet with Shell provisioner

timowest:
Support
- [x] MySQL 5.6
- [x] PostgreSQL 9.3
- [x] Mongodb
- [ ] CUBRID
- [ ] Firebird
- [ ] Oracle XE


johnktims:
What do you think about adding Mongo to the list?

Shredder121:
:+1:
I think that would be good.

johnktims:
This might be a better approach to adding mongo support: https://github.com/flapdoodle-oss/de.flapdoodle.embed.mongo

johnktims:
We're using docker to manage the test environments.

Issue #180, opened 2012-06-26
Support annotation processor default discovery process for javac


renataogarcia:
Would it be possible to add support for the default discovery  process used by javac described [here](http://docs.oracle.com/javase/6/docs/technotes/tools/solaris/javac.html)? Basically a new file named javax.annotation.processing.Processor wil be added to the META-INF/services dir containing the processors on the querydsl-apt jar.

Thanks!

PR #1661, opened 2015-11-20
#180 added documentation regarding APT based code generation with Maven

kaczynskid:
This provides documentation of how changes introduced in #180 can be used with `maven-compiler-plugin`. Also hints how this can be used to process annotations on Groovy sources. I've spent a whole day looking that up over the internet and some claimed it [can't be done](https://twitter.com/dkaczynski/status/667292975915859970). I think this should be documented for the ones that might follow.

A working example of this configuration can be found [here](https://github.com/kaczynskid/querydsl-groovy-maven).

### DIFF
--- a/querydsl-docs/src/main/docbook/en-US/content/general/codegen.xml
+++ b/querydsl-docs/src/main/docbook/en-US/content/general/codegen.xml
@@ -242,6 +242,116 @@ import com.querydsl.core.annotations.Config;
 </project>
 ]]></programlisting>
 
+    <para>Alternatively <code>maven-compiler-plugin</code> can be used as well:
+    </para>
+
+    <programlisting language="xml"><![CDATA[
+      <project>
+        <build>
+        <plugins>
+          ...
+          <plugin>
+            <artifactId>maven-compiler-plugin</artifactId>
+            <configuration>
+              <generatedSourcesDirectory>target/generated-sources/java</generatedSourcesDirectory>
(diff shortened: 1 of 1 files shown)


timowest:
Thanks for the PR! I added some comments.

querydsl-docs/src/main/docbook/en-US/content/general/codegen.xml
```diff
@@ -242,6 +242,116 @@ import com.querydsl.core.annotations.Config;
 </project>
 ]]></programlisting>
 
+    <para>Alternatively <code>maven-compiler-plugin</code> can be used as well:
```
timowest: could you also document what this will do? It looks like it will hook APT directly into the compilation. Also I think this configuration doesn't add the generated sources folder to the project source roots, or does it?

kaczynskid: Yes. It does exactly what you thought it would: runs APT processing during compilation. It also adds the generated sources folder to project sources, as generated classes are also compiled on the same pass (also IntelliJ Idea can see the generated sources folder).

timowest:
This looks good to me.

Shredder121:
Looks good to me as well.

Shredder121:
@johnktims you can merge this one.

johnktims:
Thanks, @kaczynskid. Looks good to me.

PR #1662, opened 2015-11-21
Default variable name customization

Shredder121:


querydsl-codegen/src/main/java/com/mysema/query/codegen/EntityType.java
```diff
@@ -233,10 +260,10 @@ public void include(Supertype supertype) {
     }
     
     private Property validateField(Property field) {
-        if (field.getName().equals(uncapSimpleName) || field.getEscapedName().equals(uncapSimpleName)) {
+        if (field.getName().equals(modifiedSimpleName) || field.getEscapedName().equals(modifiedSimpleName)) {
             do {
-                uncapSimpleName = StringUtils.uncapitalize(getType().getSimpleName()) + (escapeSuffix++);
-            } while (propertyNames.contains(uncapSimpleName));
```
Shredder121: @johnktims @timowest Looking back, what do you think of this?
@robertandrewbain You can chime in as well if you want.

timowest: that logic could be maybe moved into the injected function, but then the function would need to be applied later

PR #1663, opened 2015-11-21
Get rid of capitalized test method names

timowest:


querydsl-apt/src/test/java/com/querydsl/apt/DateExtensionsTest.java
```diff
@@ -34,7 +34,7 @@
     private static final String packagePath = "src/test/java/com/querydsl/apt/";
 
     @Test
-    public void Handles_Date_Extensions_Correctly() throws IOException, InterruptedException {
+    public void handles_Date_Extensions_Correctly() throws IOException, InterruptedException {
```
Shredder121: I understand the 'getting rid of capitalized test method names' part.
I do think that the PR should maybe be 'unify naming of method names'.
If we go from one non-standard way into another, that might distract from the development.

timowest: I see the following naming patterns we could use:
* camelCase
* camelCase_variant_in_lower_case_snake_case
* requirement_in_lowercase_snake_case

What do you think?

Shredder121: I don't quite see what you mean with `camelCase_variant_in_lower_case_snake_case`.
But I think that `camelCase` is fine.

querydsl-apt/src/test/java/com/querydsl/apt/domain/QueryEmbeddedTest.java
```diff
@@ -55,17 +55,17 @@
...
 
     @Test
-    public void Parent_Child_ChildProperty() {
+    public void parent_child_childProperty() {
         assertNotNull(QQueryEmbeddedTest_Parent.parent.child.childProperty);
     }
 
     @Test
-    public void Parent_Children_Any_ChildrenProperty() {
+    public void parent_children_any_childrenProperty() {
```
Shredder121: Should probably be `childProperty`.

querydsl-codegen/src/test/java/com/querydsl/codegen/BeanSerializerTest.java
```diff
@@ -111,7 +111,7 @@ public void Interfaces2() throws IOException {
     }
 
     @Test
-    public void ToString() throws IOException {
+    public void toString_() throws IOException {
```
Shredder121: Lol. That happens as well then yes.

querydsl-codegen/src/test/java/com/querydsl/codegen/EntitySerializerTest.java
```diff
@@ -78,7 +78,7 @@ public void No_Package() throws IOException {
     }
 
     @Test
-    public void OriginalCategory() throws IOException {
+    public void originalCategory() throws IOException {
```
Shredder121: Don't you think this one stands out between all the `snake_case`?
If it were a class or property name, I'd say preserve the casing.
Maybe we should `snake_case` all 'test specifications'?

What do you think?

querydsl-codegen/src/test/java/com/querydsl/codegen/GenericExporterTest.java
```diff
@@ -57,7 +57,7 @@ public void Export_With_Keywords() throws IOException {
     }
 
     @Test
-    public void Export_With_Stopclass() {
+    public void export_with_stopclass() {
```
Shredder121: With preserve casing I'd say that `uncapSimpleName_escaped` is perfect, but this one would be `export_with_stopClass`.

What's your take on this? @johnktims @robertandrewbain

querydsl-collections/src/test/java/com/querydsl/collections/LeftJoinTest.java
```diff
@@ -62,7 +62,7 @@ public void List() {
     }
 
     @Test
-    public void Alias() {
+    public void alias1() {
```
Shredder121: If something clashes, I'd say it's nice if we have a convention on how to 'escape' it.
Shall we use `1` or `_`? Or if you know something else?

querydsl-core/src/test/java/com/querydsl/core/BooleanBuilderTest.java
```diff
@@ -31,26 +31,26 @@
...
-    public void AndAnyOf() {
+    public void andAnyOf() {
         BooleanBuilder builder = new BooleanBuilder();
         builder.andAnyOf(first, null);
         assertEquals(first, builder.getValue());
     }
 
     @Test
-    public void AndAnyOf2() {
+    public void andAnyOf2() {
```
Shredder121: What about `and_anyOf`, `and_anyOf2`, `or_allOf` etc..?

querydsl-codegen/src/test/java/com/querydsl/codegen/EntitySerializerTest.java
timowest: Yes, that was the plan.

Shredder121: Cool, yes I didn't get what you meant with `camelCase_variant_in_lower_case_snake_case`.
I'm all for it.

querydsl-collections/src/test/java/com/querydsl/collections/LeftJoinTest.java
timowest: You are right, it should be consistent. Any preference?

Shredder121: To me the `_` made sense, since it is the smallest, non-intrusive, way of escaping.
if you use any number I instantly look for the other `2`, `3` etc.. cases.
@johnktims Your opinion?

querydsl-apt/src/test/java/com/querydsl/apt/domain/EntityTest.java
```diff
@@ -87,7 +87,7 @@ public void Inheritance() {
     }
 
     @Test
-    public void SupertypePaths() {
+    public void supertypePaths() {
```
Shredder121: I would say constructs like these should then be separated by  `_`?

timowest: I'll fix the rest tomorrow, unless someone else is faster ;)

timowest:
Can we do fine-tuning of the test names in a separate PR? The naming patterns used here are
* underscore based test case / requirement
* camel case based method name to test

Shredder121:
Yes, it at least got rid of the Capitalized starting of method names.

So looks good to me.

PR #1664, opened 2015-11-21
Simplify codegen tests

timowest:


Shredder121:
I think that using `ErrorCollector` would be nice.

querydsl-jpa-codegen/src/test/java/com/querydsl/jpa/codegen/HibernateDomainExporterTest.java
```diff
@@ -13,204 +13,54 @@
...
+        Set<File> files = exporter.getGeneratedFiles();
+        assertFalse(files.isEmpty());
+        for (File file : files) {
+            String path = file.getAbsolutePath().replace(
+                    folder.getRoot().getAbsolutePath(), origRoot.getAbsolutePath());
+            String reference = Files.toString(new File(path), Charsets.UTF_8);
+            String content = Files.toString(file, Charsets.UTF_8);
+            try {
+                assertEquals(file.getPath(), reference, content);
+            } catch (Throwable e) {
```
Shredder121: I would write it like this, it does away with the try-catch:
```java
errors.checkThat(file.getPath(), content, is(equalTo(reference)));
```

robertandrewbain:
Looks good to me.

querydsl-jpa-codegen/src/test/java/com/querydsl/jpa/codegen/HibernateDomainExporterTest.java
```diff
@@ -13,203 +13,50 @@
...
+
+        File origRoot = new File("../querydsl-jpa/target/generated-test-sources/java");
+        Set<File> files = exporter.getGeneratedFiles();
+        assertFalse(files.isEmpty());
+        for (File file : files) {
+            String path = file.getAbsolutePath().replace(
+                    folder.getRoot().getAbsolutePath(), origRoot.getAbsolutePath());
+            String reference = Files.toString(new File(path), Charsets.UTF_8);
+            String content = Files.toString(file, Charsets.UTF_8);
+            errors.checkThat(content, is(equalTo(reference)));
```
Shredder121: Don't you want to keep the `reason` parameter in this?

querydsl-jpa-codegen/src/test/java/com/querydsl/jpa/codegen/ant/AntJPADomainExporterTest.java
```diff
@@ -1,25 +1,43 @@
...
-        assertTrue(new File("target/AntJPADomainExporterTest/com/querydsl/jpa/codegen/ant/QDepartments.java").exists());
+        File origRoot = new File("../querydsl-jpa/target/generated-test-sources/java");
+        Set<File> files = exporter.getGeneratedFiles();
+        assertFalse(files.isEmpty());
+        for (File file : files) {
+            String path = file.getAbsolutePath().replace(
+                    folder.getRoot().getAbsolutePath(), origRoot.getAbsolutePath());
+            String reference = Files.toString(new File(path), Charsets.UTF_8);
+            String content = Files.toString(file, Charsets.UTF_8);
+            assertEquals(file.getPath(), reference, content);
```
Shredder121: You don't want to move them all to hamcrest?

Shredder121:
Could @timowest maybe bump the minimum coverage to 75% on coveralls after this is merged?
Then we at least are anchored and secured.

querydsl-jpa-codegen/src/test/java/com/querydsl/jpa/codegen/HibernateDomainExporterTest.java
timowest: how to get the reason in?

Shredder121: You're now using this one.
`public <T> void checkThat(T value, Matcher<T> matcher) ...`
You can also use this one
`public <T> void checkThat(String reason, T value, Matcher<T> matcher) ...`

Is `String path` a readable path of the file?
Or else `file.toString()`?

Shredder121: Oh it seems we're on an older version of junit than I usually am.
I am at `junit-4.12` and use ErrorCollector myself, so thought that the `reason` would make more sense.
You can either bump the junit version (and see which surefire version works) or just leave it like this.

timowest:
Yes, i'll bump the minimum coverage.

Shredder121:
They added some validation regarding `@Rule` and static `TestRule`s/`MethodRule`s.
I am seeing if I can rewrite them as `TestRule`s, since they are a little cleaner.

Shredder121:
I am done, tell me how it looks?

timowest:
Looks good.

johnktims:
Looks good to me.

PR #1666, opened 2015-11-22
Fix modulo override

### LINKED ISSUE #1665: Wrong serialization of modulo operator with JDOQL
Hi, I just found that queries that use the modulo operator are serialized wrongly for JDOQL and the reason is the corresponding template pattern is incorrectly overridden in JavaTemplates class (https://github.com/querydsl/querydsl/blob/master/querydsl-core/src/main/java/com/querydsl/core/types/JavaTemplates.java):

    add(Ops.MOD, "{0} % {0}"); 

I could gladly submit a PR for this—apparently—easy to fix issue but as I'm not really familiarized with the code I'm not sure what the reason for that override might be (perhaps instead of changing it, it should be removed from that class? the pattern found on the base class is the same :) ).

timowest:
Fixes #1665


Shredder121:
Looks good to me.

johnktims:
Looks good.

Issue #1668, opened 2015-11-25
NumberExpression class init deadlock


margothi:
(Found using version 4.0.3, suspected problem persists in head revision)

We have been encountering a class init deadlock in our system. I believe I have tracked it to be because of com.querydsl.core.types.dsl.NumberExpression:38 calling in static init:

    private static final NumberExpression<Double> random = Expressions.numberOperation(Double.class, MathOps.RANDOM);

Which requires loading of com.querydsl.core.types.dsl.Expressions. However, another thread may end up concurrently loading _that_ one, leading to Expressions:612:

    return new NumberTemplate<T>(cl, template, args);

Which induces loading of NumberTemplate, which induces loading of NumberExpression, and we have a deadlock. I may be wrong as I have never really delved into QueryDSL source. Please find below the significant portions of a thread dump:

	Thread 22: (state = BLOCKED)
	 - com.querydsl.core.types.dsl.NumberExpression.<clinit>() @bci=9, line=38 (Interpreted frame)
	 - com.querydsl.core.types.dsl.BeanPath.createNumber(java.lang.String, java.lang.Class) @bci=1, line=248 (Interpreted frame)
	 - my.querydsl.generated.domain.QDomainObjectA.<init>(java.lang.String) @bci=20, line=27 (Interpreted frame)
	 - my.querydsl.generated.domain.QDomainObjectA.<clinit>() @bci=6, line=25 (Interpreted frame)

	Thread 21: (state = BLOCKED)
	 - com.querydsl.core.types.dsl.Expressions.numberTemplate(java.lang.Class, com.querydsl.core.types.Template, com.google.common.collect.ImmutableList) @bci=0, line=612 (Interpreted frame)
	 - com.querydsl.core.types.dsl.Expressions.numberTemplate(java.lang.Class, java.lang.String, java.lang.Object[]) @bci=9, line=573 (Interpreted frame)
	 - com.querydsl.core.types.dsl.Expressions.<clinit>() @bci=8, line=32 (Interpreted frame)
	 - com.querydsl.core.types.dsl.DateExpression.currentDate(java.lang.Class) @bci=8, line=56 (Interpreted frame)
	 - com.querydsl.core.types.dsl.DateExpression.<clinit>() @bci=2, line=36 (Interpreted frame)
	 - com.querydsl.core.types.dsl.BeanPath.createDate(java.lang.String, java.lang.Class) @bci=1, line=190 (Interpreted frame)
[...]

PR #1669, opened 2015-11-25
Delay static field init to avoid class deadlock

margothi:
Should fix https://github.com/querydsl/querydsl/issues/1668


timowest:
Thanks for the PR! I used the initialization-on-demand holder idiom instead here https://github.com/querydsl/querydsl/pull/1673

It should be safer than this solution.

PR #1671, opened 2015-11-25
Fix getResults handling

timowest:
Related to #1254


Shredder121:
Looks good to me.

I have a suggestion for maybe later.
What do you think of using a `Closer` to collect all the things that must be closed?

timowest:
A Closer sounds good.

johnktims:
Looks good to me.

PR #1672, opened 2015-11-25
Deprecate getResults with arguments

timowest:


johnktims:
This looks like a fine change.

querydsl-sql/src/main/java/com/querydsl/sql/AbstractSQLQuery.java
```diff
@@ -251,10 +251,23 @@ protected void endContext(SQLListenerContext context) {
...
     public ResultSet getResults(Expression<?>... exprs) {
-        queryMixin.setProjection(exprs);
+        if (exprs.length > 0) {
+            queryMixin.setProjection(exprs);
+        }
+        return getResults();
+    }
 
+    /**
+     * Get the results as an JDBC result set
```
Shredder121: I would turn this into
`Get the results as a JDBC result set`
or
`Get the results as a JDBC ResultSet`

Shredder121:
I made a small adjustment to the Javadoc, to me this looks good now.
@johnktims can you look one more time and merge if it's okay?

johnktims:
Looks good to me.

PR #1673, opened 2015-11-25
Delay constant initialization via static inner class

### LINKED ISSUE #1668: NumberExpression class init deadlock

timowest:
Fixes #1668

### DIFF
--- a/querydsl-core/src/main/java/com/querydsl/core/types/ConstantImpl.java
+++ b/querydsl-core/src/main/java/com/querydsl/core/types/ConstantImpl.java
@@ -28,50 +28,53 @@ public final class ConstantImpl<T> extends ExpressionBase<T> implements Constant
 
     private static final int CACHE_SIZE = 256;
 
-    @SuppressWarnings({"rawtypes", "unchecked"}) //generic array creation not possible
-    private static final Constant<Character>[] CHARACTERS = new Constant[CACHE_SIZE];
+    private static class Constants {
 
-    @SuppressWarnings({"rawtypes", "unchecked"}) //generic array creation not possible
-    private static final Constant<Byte>[] BYTES = new Constant[CACHE_SIZE];
+        @SuppressWarnings({"rawtypes", "unchecked"}) //generic array creation not possible
+        private static final Constant<Character>[] CHARACTERS = new Constant[CACHE_SIZE];
 
-    @SuppressWarnings({"rawtypes", "unchecked"}) //generic array creation not possible
-    private static final Constant<Integer>[] INTEGERS = new Constant[CACHE_SIZE];
+        @SuppressWarnings({"rawtypes", "unchecked"}) //generic array creation not possible
+        private static final Constant<Byte>[] BYTES = new Constant[CACHE_SIZE];
 
-    @SuppressWarnings({"rawtypes", "unchecked"}) //generic array creation not possible
-    private static final Constant<Long>[] LONGS = new Constant[CACHE_SIZE];
+        @SuppressWarnings({"rawtypes", "unchecked"}) //generic array creation not possible
+        private static final Constant<Integer>[] INTEGERS = new Constant[CACHE_SIZE];
 
-    @SuppressWarnings({"rawtypes", "unchecked"}) //generic array creation not possible
-    private static final Constant<Short>[] SHORTS = new Constant[CACHE_SIZE];
+        @SuppressWarnings({"rawtypes", "unchecked"}) //generic array creation not possible
+        private static final Constant<Long>[] LONGS = new Constant[CACHE_SIZE];
 
-    private static final Constant<Boolean> FALSE = new ConstantImpl<Boolean>(Boolean.FALSE);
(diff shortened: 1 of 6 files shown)


querydsl-core/src/main/java/com/querydsl/core/types/ConstantImpl.java
```diff
@@ -28,32 +28,34 @@
 
     private static final int CACHE_SIZE = 256;
 
-    @SuppressWarnings({"rawtypes", "unchecked"}) //generic array creation not possible
-    private static final Constant<Character>[] CHARACTERS = new Constant[CACHE_SIZE];
+    private static final Constant<Boolean> FALSE = new ConstantImpl<Boolean>(Boolean.FALSE);
```
johnktims: Using `create(false)` would utilize the cache instead of creating another instance, right?

timowest: Yes, but here we can't use it

Shredder121: Why didn't you also move these in the `Constants` class?

timowest: Yes, I'll move them, initially I thought they could be out, since it's only two boolean constants, but for consistency ...

querydsl-core/src/main/java/com/querydsl/core/types/ConstantImpl.java
```diff
@@ -28,50 +28,52 @@
 
     private static final int CACHE_SIZE = 256;
 
-    @SuppressWarnings({"rawtypes", "unchecked"}) //generic array creation not possible
-    private static final Constant<Character>[] CHARACTERS = new Constant[CACHE_SIZE];
```
Shredder121: Is it also possible to reorder these to the original ordering?
Or do you want it in this order?

Shredder121: Hmm, it's not entirely what I'd hoped for, but thanks anyway!

timowest: Could you be more specific then?

Shredder121: No it's fine, I just thought the diff would be a little more cleaner.
Since there is a new line introduced, git mismatches the lines a bit.

Shredder121:
I finished the test, how does it look?

johnktims:
Looks good to me now.

Issue #1670, opened 2015-11-25
executeWithKey and MySQL throw SQLException


serphacker:
executeWithKey return the following exception : 

```
Exception in thread "main" com.querydsl.core.QueryException: Caught SQLException for insert into `user` (`email`)
values (?)
	at com.querydsl.sql.DefaultSQLExceptionTranslator.translate(DefaultSQLExceptionTranslator.java:50)
	at com.querydsl.sql.Configuration.translate(Configuration.java:453)
	at com.querydsl.sql.dml.SQLInsertClause.executeWithKeys(SQLInsertClause.java:364)
	at com.querydsl.sql.dml.SQLInsertClause.executeWithKey(SQLInsertClause.java:172)
	at com.querydsl.sql.dml.SQLInsertClause.executeWithKey(SQLInsertClause.java:153)
	at com.serphacker.qdslbug.App.main(App.java:33)
Caused by: java.sql.SQLException: getGeneratedKeys error. Statement.RETURN_GENERATED_KEYS must be precised on connection.prepareStatement(String sql, int autoGeneratedKeys) or statement.executeUpdate(String sql, int autoGeneratedKeys)
	at org.mariadb.jdbc.MariaDbStatement.getGeneratedKeys(MariaDbStatement.java:800)
	at com.querydsl.sql.dml.SQLInsertClause.executeWithKeys(SQLInsertClause.java:351)
	... 3 more
```

Tested with querydsl 4.0.6 

```
        drop table if exists `user`;
        create table `user` (
            id int primary key auto_increment,
            email varchar(255)
        );
```

```
        Configuration conf = new Configuration(new MySQLTemplates());
        QUser t_user = QUser.user;

        Integer key = new SQLInsertClause(connection, conf, t_user)
            .set(t_user.email, "email")
            .executeWithKey(t_user.id);
```

Related to #1040

timowest:
Which version of the Maria driver do you use?

timowest:
Also do you use it with MySQL or MariaDB and which version?

serphacker:
@timowest I tested against mariadb 10.0.22 but it looks like it's a mariadb driver issue : 

mariadb-java-client 1.3.x : exception triggered
mariadb-java-client <= 1.2.x : ok, return generated keys
mysql-connector-java 5.1.37 : ok, return generated keys

serphacker:
https://mariadb.com/kb/en/mariadb/mariadb-connector-j-130-release-notes/

[...]

PR #1675, opened 2015-11-26
Use MariaDB driver instead of MySQL driver

timowest:
more of a try

### DIFF
--- a/querydsl-sql/pom.xml
+++ b/querydsl-sql/pom.xml
@@ -78,10 +78,18 @@
       <version>${derby.version}</version>
       <scope>test</scope>
     </dependency>
+    <!--
     <dependency>
       <groupId>mysql</groupId>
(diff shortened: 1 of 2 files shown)


querydsl-sql/pom.xml
```diff
@@ -78,12 +78,19 @@
...
       <artifactId>mysql-connector-java</artifactId>
       <version>${mysql.version}</version>
       <scope>test</scope>
     </dependency>
+    -->
+    <dependency>
+      <groupId>org.mariadb.jdbc</groupId>
+      <artifactId>mariadb-java-client</artifactId>
+      <version>1.3.2</version>
+    </dependency>
```
Shredder121: `<scope>test</scope>`?

timowest:
Closing, since switching to the MariaDB driver doesn't seem to work now https://github.com/querydsl/querydsl/issues/1670

PR #1682, opened 2015-11-29
Issue 1680: Add utility methods to Expression to get typed expressions

### LINKED ISSUE #1680: Improve API for concatenating a subquery with a literal
The best I can come up with is:
```java
Expressions.stringOperation(
    Ops.CONCAT, 
    SQLExpressions.select(foo.fooString).from(foo), 
    Expressions.constant(" "));
````
I feel that using `Ops` in non-querydsl code is a leaky abstraction and that instead, support for this should be added to the `Expressions` class, offering the following API:

````java
	public static StringExpression append(Expression<String> left, Expression<String> right)

	public static StringExpression concat(Expression<String> left, Expression<String> right)

	public static StringExpression append(Expression<String> left, String right) 

	public static StringExpression concat(Expression<String> left, String right)
````

Having `append` and `concat` provides consistency with `StringExpression`.

### PR DESCRIPTION
**Pull request by robertandrewbain**
Fixes #1680

### DIFF
--- a/querydsl-core/src/main/java/com/querydsl/core/types/dsl/Expressions.java
+++ b/querydsl-core/src/main/java/com/querydsl/core/types/dsl/Expressions.java
@@ -57,7 +57,7 @@ public final class Expressions {
         if (source == null) {
             return as(Expressions.<D>nullExpression(), alias);
         } else {
-            return Expressions.operation(alias.getType(), Ops.ALIAS, source, alias);
+            return operation(alias.getType(), Ops.ALIAS, source, alias);
         }
     }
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by timowest** [comment]
For completeness the following types should also be supported: ComparableExpression, DateTimeExpression, TimeExpression, EnumExpression

**Comment by timowest** [comment]
Also maybe overloaded methods for constants?

**Comment by timowest** [comment]
Unwrapping should also be done:

```
public static StringExpression asString(Expression<String> expr) {
    Expression<String> underlyingMixin = ExpressionUtils.extract(expr);
    return new StringExpression(underlyingMixin) {
        @Override
        public <R, C> R accept(Visitor<R, C> v, C context) {
            return this.mixin.accept(v, context);
        }
    };
}
```

File: querydsl-core/src/main/java/com/querydsl/core/types/dsl/Expressions.java [hunk 1]
```diff
@@ -1463,4 +1463,84 @@ private static Template createTemplate(String template) {
...
 
+    /**
+     * Create a new BooleanExpression
+     *
+     * @param expr Expression of type Boolean
+     * @return new BooleanExpression
+     */
+    public static BooleanExpression asBoolean(Expression<Boolean> expr) {
+
+        return new BooleanExpression(expr) {
```
Shredder121: I'd say extract the underlying mixin of the incoming `expr` by using `Expression<Boolean> underlyingMixin = ExpressionUtils.extract(expr)`.

then `return new BooleanExpression(underlyingMixin) {...}`.

The same for the rest also?
You never know what assumptions code elsewhere makes.

robertandrewbain:
Unwrapping added as per review feedback along with support for `ComparableExpression`, `DateTimeExpression`, `TimeExpression`, `EnumExpression` and overloaded methods for constants.

querydsl-core/src/main/java/com/querydsl/core/types/dsl/Expressions.java
```diff
@@ -1463,4 +1463,244 @@ private static Template createTemplate(String template) {
...
+    }
+
+    /**
+     * Create a new BooleanExpression
+     *
+     * @param value boolean
+     * @return new BooleanExpression
+     */
+    public static BooleanExpression asBoolean(boolean value) {
+        return asBoolean(Expressions.constant(value));
```
timowest: this is inside Expressions, so maybe just `asBoolean(constant(value))`, the same with the others

Shredder121:
Had to close because of Travis issues.

robertandrewbain:
Adding support for #1691

timowest:
This PR is related and would simplify the wrapping https://github.com/querydsl/querydsl/pull/1705

timowest:
Looks good.

Shredder121:
Looks good to me as well.

johnktims:
Looks good to me.

PR #1689, opened 2015-12-01
Fix setting root path to null

### LINKED ISSUE #1686: set null values to column causes NullPointerException when inserting or updating
I don't use class generation for mapping my entities. I'm using queryDSL version 4.0.5

Here is my code :

    SQLTemplates templates = MySQLTemplates.builder().printSchema().build();
    Configuration configuration = new Configuration( templates );

    try
    {
        Class.forName( "com.mysql.jdbc.Driver" );

        final Connection connection = DriverManager.getConnection( "jdbc:mysql://localhost:3306/test", "root", "" );

        SQLQuery sqlQuery = new SQLQuery( connection, configuration );

        // querying
        NumberPath< Long > volume = Expressions.numberPath( Long.class, "volume" );

        RelationalPathBase< Object > r = new RelationalPathBase< Object >( Object.class, "", "test", "mstf" );

        Provider< java.sql.Connection > provinv = new Provider< Connection >()
        {

            @Override
            public Connection get()
            {
                return connection;
            }
        };

        SQLQueryFactory queryFactory = new SQLQueryFactory( configuration, provinv );

        queryFactory.update( r ).setNull( volume ).execute();

    }
    catch( Exception e )
    {
        e.printStackTrace();
    }
When executing, I got a wonderful NullPointerException because of this code in class ColumnMetadata:

public int getJdbcType() {
    return jdbcType;
}
Of course we have a jdbcType of type INTEGER, it is null and we must return an int...

Shredder121:
Fixes #1686

### DIFF
--- a/querydsl-sql/src/main/java/com/querydsl/sql/Configuration.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/Configuration.java
@@ -223,7 +223,13 @@ public final class Configuration {
     @SuppressWarnings({ "unchecked", "rawtypes" })
     public <T> void set(PreparedStatement stmt, Path<?> path, int i, T value) throws SQLException {
         if (Null.class.isInstance(value)) {
-            Integer sqlType = path != null ? ColumnMetadata.getColumnMetadata(path).getJdbcType() : null;
+            Integer sqlType = null;
+            if (path != null) {
+                ColumnMetadata columnMetadata = ColumnMetadata.getColumnMetadata(path);
+                if (columnMetadata.hasJdbcType()) {
+                    sqlType = columnMetadata.getJdbcType();
+                }
+            }
             if (sqlType != null) {
                 stmt.setNull(i, sqlType);
             } else {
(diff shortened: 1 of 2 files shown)


querydsl-sql/src/main/java/com/querydsl/sql/Configuration.java
```diff
@@ -223,7 +223,7 @@ public String getColumnOverride(SchemaAndTable key, String column) {
     @SuppressWarnings({ "unchecked", "rawtypes" })
     public <T> void set(PreparedStatement stmt, Path<?> path, int i, T value) throws SQLException {
         if (Null.class.isInstance(value)) {
-            Integer sqlType = path != null ? ColumnMetadata.getColumnMetadata(path).getJdbcType() : null;
+            Integer sqlType = path != null && !path.getMetadata().isRoot() ? ColumnMetadata.getColumnMetadata(path).getJdbcType() : null;
```
timowest: The test case is ok, but I'd fix this differently, since root path usage is only one case where this breaks. Maybe
```
Integer sqlType = null;
if (path) {
    ColumnMetadata metadata = ColumnMetadata.getColumnMetadata(path);
    if (metadata.hasJdbcType()) {
        jdbcType = metadata.getJdbcType();
    }
}
```

timowest:
Looks good now

johnktims:
Looks good to me.

PR #1691, opened 2015-12-02
Update JTSGeometryExpressions: added transform

JWGmeligMeyling:
Added the transform operation to JTSGeometryExpressions

### DIFF
--- a/querydsl-spatial/src/main/java/com/querydsl/spatial/jts/JTSGeometryExpressions.java
+++ b/querydsl-spatial/src/main/java/com/querydsl/spatial/jts/JTSGeometryExpressions.java
@@ -72,6 +72,19 @@ public final class JTSGeometryExpressions {
     }
 
     /**
+     * Transform the SRID on a geometry to a particular integer value.
+     *
+     * @param expr geometry
+     * @param srid SRID
+     * @param <T>
+     * @return geometry
+     */
+    public static <T extends Geometry> JTSGeometryExpression<T> transformSRID(Expression<T> expr, int srid) {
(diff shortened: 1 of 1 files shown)


timowest:
Transform is already available in JTSGeomtryExpression, so no need to duplicate it as a static method.

JWGmeligMeyling:
How would I do that with a `ConstantImpl.create(geometry)`?

JWGmeligMeyling:
Nevermind, I missed that one. Thanks!

timowest:
You could transform the JTS geometry via JTS tools before using it in a query, I think that makes more sense.

JWGmeligMeyling:
After scrolling a bit through all the available classes, I still wasn't able to find how to get from a `Constant` expression to a `JTSGeometryExpression`. I ended up writing:

```java
	private class JTSConstantGeometry<T extends Geometry> extends JTSGeometryExpression<T> {

		public JTSConstantGeometry(T constant) {
			super(ConstantImpl.create(constant));
		}

		@Nullable
		@Override
		public <R, C> R accept(Visitor<R, C> v, C context) {
			return v.visit((ConstantImpl<Geometry>) mixin, context);
		}
	}
```

Am I still miserably missing something? :wink:

timowest:
@robertandrewbain Could this be covered also in https://github.com/querydsl/querydsl/pull/1682 ?

The relevant factory classes are `GeometryExpressions` and `JTSGeometryExpressions`.

robertandrewbain:
@timowest, yes, I believe that it could. I'll add that tonight.

timowest:
Closing this in favor of a constant -> typed expression conversion approach.

PR #1693, opened 2015-12-02
Improve JDO tests

timowest:


Shredder121:
Yes, looks good.
I didn't like that the tests failed for no reason.
So thanks for this fix!

johnktims:
Looks good.

PR #1694, opened 2015-12-02
Connection on demand for DML clauses

### LINKED ISSUE #1683: Obtain Connection in SQLInsert/Update When Needed
As with #1596, also inserts and updates should obtain connection when needed instead of construction time. Obtaining it at construction time causes problems especially with multi-table batch inserts/updates and direct DataSource, as multiple connections are acquired and held unnecessarily long.

timowest:
Fixes #1683

### DIFF
--- a/querydsl-sql/src/main/java/com/querydsl/sql/dml/AbstractSQLClause.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/dml/AbstractSQLClause.java
@@ -18,6 +18,9 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 
+import javax.annotation.Nullable;
+import javax.inject.Provider;
+
 import org.slf4j.Logger;
 import org.slf4j.MDC;
 
@@ -47,10 +50,24 @@ public abstract class AbstractSQLClause<C extends AbstractSQLClause<C>> implemen
 
     protected SQLListenerContextImpl context;
 
-    public AbstractSQLClause(Configuration configuration) {
+    @Nullable
+    private Provider<Connection> connProvider;
+
+    @Nullable
+    private Connection conn;
+
+    public AbstractSQLClause(Configuration configuration, Provider<Connection> connProvider) {
         this.configuration = configuration;
         this.listeners = new SQLListeners(configuration.getListeners());
         this.useLiterals = configuration.getUseLiterals();
+        this.connProvider = connProvider;
+    }
+
+    public AbstractSQLClause(Configuration configuration, Connection conn) {
+        this.configuration = configuration;
+        this.listeners = new SQLListeners(configuration.getListeners());
+        this.useLiterals = configuration.getUseLiterals();
+        this.conn = conn;
     }
 
     /**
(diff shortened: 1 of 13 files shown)


querydsl-sql/src/main/java/com/querydsl/sql/dml/AbstractSQLClause.java
```diff
@@ -47,10 +50,24 @@
...
+    private Provider<Connection> connProvider;
+
+    @Nullable
+    private Connection conn;
+
+    public AbstractSQLClause(Configuration configuration, Provider<Connection> connProvider) {
         this.configuration = configuration;
         this.listeners = new SQLListeners(configuration.getListeners());
         this.useLiterals = configuration.getUseLiterals();
+        this.connProvider = connProvider;
```
Shredder121: The parameter passed in here should be non-null.

querydsl-sql/src/main/java/com/querydsl/sql/dml/AbstractSQLClause.java
```diff
@@ -47,10 +50,24 @@
...
+    }
+
+    public AbstractSQLClause(Configuration configuration, Connection conn) {
+        this.configuration = configuration;
+        this.listeners = new SQLListeners(configuration.getListeners());
+        this.useLiterals = configuration.getUseLiterals();
+        this.conn = conn;
```
Shredder121: The parameter passed in here should be non-null.

querydsl-sql/src/main/java/com/querydsl/sql/dml/AbstractSQLClause.java
timowest: for serialization usage it could be null

querydsl-sql/src/main/java/com/querydsl/sql/dml/AbstractSQLClause.java
timowest: for serialization usage it could be null

querydsl-sql/src/main/java/com/querydsl/sql/dml/AbstractSQLClause.java
```diff
@@ -220,6 +237,13 @@ protected void reset() {
         cleanupMDC();
     }
 
+    protected Connection connection() {
+        if (conn == null && connProvider != null) {
+            conn = connProvider.get();
```
Shredder121: Then put an assert here?
Just so there is an exception, that indicates that it's deliberate.

Shredder121: Then put an assert here?
Just so there is an exception, that indicates that it's deliberate.

querydsl-sql/src/main/java/com/querydsl/sql/dml/AbstractSQLClause.java
```diff
@@ -220,6 +237,13 @@ protected void reset() {
...
+        }
+        return conn;
```
Shredder121: Then put an assert here?
Just so there is an exception, that indicates that the connection was null.

Shredder121:
This now looks good to me.

johnktims:
Looks good to me.

PR #1698, opened 2015-12-06
Use TotalHitCountCollector for count queries

### LINKED ISSUE #1676: querydsl-lucene: AbstractLuceneQuery.innerCount() should use TotalHitCountCollector instead of TopDocs.totalHits
TotalHitCountCollector is the preferred way for count only queries since it can be more performant, especially for large indexes. 

I suggest changing the method:

```java
private long innerCount() {
    try {
        final int maxDoc = searcher.getIndexReader().maxDoc();
        if (maxDoc == 0) {
            return 0;
        }
        return searcher.search(createQuery(), getFilter(), maxDoc, Sort.INDEXORDER, false, false).totalHits;
    } catch (IOException e) {
        throw new QueryException(e);
    } catch (IllegalArgumentException e) {
        throw new QueryException(e);
    }
}
```

to:

```java
private int innerCount() {
    try {
        final int maxDoc = searcher.getIndexReader().maxDoc();
        if (maxDoc == 0) {
            return 0;
        }
        TotalHitCountCollector collector = new TotalHitCountCollector();
        searcher.search(createQuery(), getFilter(), collector);
        return collector.getTotalHits();
    } catch (IOException e) {
        throw new QueryException(e);
    } catch (IllegalArgumentException e) {
        throw new QueryException(e);
    }
}
```

timowest:
Fixes #1676

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ env:
     - MODULE=!querydsl-scala
     - MODULE=querydsl-scala
 install:
-  - mvn -B -q install -DskipTests=true
+  - mvn -B -q --projects ${MODULE} install -DskipTests=true
 before_script:
   - mysql -u root -e "source travis/mysql.sql"
   - psql -U postgres -f travis/postgresql.sql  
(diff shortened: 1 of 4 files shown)


johnktims:
This looks fine to me. I don't know much about Lucene, but I was unable to find any information showing that `TotalHitCountCollector` is any better than what we had.

robertandrewbain:
Looks good to me.

Shredder121:
Shall we do [the suggestion](https://github.com/querydsl/querydsl/issues/1676#issuecomment-164409030) in a different PR?

timowest:
> Shall we do the suggestion in a different PR?

Yes, maybe.

Shredder121:
I think this is fine as well, although I couldn't really find the benefits of this instead of the searcher.
For the fetchResults improvement, I think we'll have to revert back to the Searcher, but for just the count, this is fine.

PR #1701, opened 2015-12-08
Bump jts version

### LINKED ISSUE #1697: Missing  com.vividsolutions:jts:jar:1.10
The last version of querydsl code gen depends on  com.vividsolutions:jts:jar:1.10 that does not exists in maven central.

```xml
<dependency>
    <groupId>com.querydsl</groupId>
    <artifactId>querydsl-sql-codegen</artifactId>
    <version>4.0.7</version>
</dependency>
```

The only versions available are 1.8, 1.11, 1.12 and 1.13

image

jts is a transitive dependency of querydsl-sql-spacial. Can't querydsl-sql-spacial be made \<optional\>?

timowest:
Fixes #1697


querydsl-sql-spatial/pom.xml
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="UTF-8"?>
+<?xml version="1.0" encoding="UTF-8"?>
```
Shredder121: Did you find out yet what causes these changes?
Is it required by spec to end the prelude in CRLF?
Or are they now LF?

timowest: I didn't check.

timowest: It was CRLF -> LF, I checked now.

Shredder121:
Looks good.

robertandrewbain:
Looks good to me.

Issue #1703, opened 2015-12-09
Fix Cubrid build issues on Travis


timowest:

Shredder121:
[I submitted the issue at Travis' side](https://github.com/travis-ci/travis-ci/issues/5254), let's see if they know what's going on.

PR #1704, opened 2015-12-09
Skip CUBRID on Travis

timowest:
Related to #1703


Shredder121:
@robertandrewbain !
This means we will have no more issues :wink:

PR #1705, opened 2015-12-09
Make Expression types instantiable

timowest:


timowest:
@johnktims @Shredder121 @robertandrewbain Do you see problems with this approach? Maybe we could make the constructors of the *Expression types protected?

Shredder121:
Depends on what you want to achieve.
Do you want to not directly extend *Expression types?

timowest:
I thought that directly providing the support to wrap expressions would be beneficial.

timowest:
But then I realized that there are serious drawbacks in this approach, e.g. a wrapped path won't be a path anymore.

timowest:
I'll provide direct instantiation in a different (and safer) way later.

PR #1706, opened 2015-12-09
Rename <groups> property to <includedGroups> to prevent clashing

Shredder121:


johnktims:
For more info: https://maven.apache.org/plugins/maven-javadoc-plugin/aggregate-mojo.html#groups

johnktims:
Looks good to me. This doesn't need to be backported since #1538 wasn't backported.

Shredder121:
But if it is backported, then this needs to be included.
But nice that it looks good. Let's see what Timo thinks.

timowest:
Both PRs are not really bugfixes, so backporting is not really mandatory.

PR #1709, opened 2015-12-12
Escape query.toString()

### LINKED ISSUE #1633: LuceneQuery.toString produces non printable characters
It may be related to issue #1555 When you use IN operator with numbers and use toString method on such query, output will contain binary characters. 
In my case toString produces something like that

 +(clientId:! clientId: clientId: clientId:d clientId: clientId:E clientId: clientId:G clientId:g clientId: clientId:H clientId:`  clientId:+ clientId:+ clientId: clientId: clientId:T clientId: clientId:v clientId: clientId:X clientId:8)

every clientId was a number, but was not printed.

It can also clear your screen or mess your terminal (when some control characters are produced) - as it happens in my case - for some queries screen is cleared. I'm sure it may have some security impact on users, as control sequences for terminals are quite powerful.

timowest:
Fixes #1633


robertandrewbain:
Looks good to me.

Shredder121:
Agreed.

PR #1711, opened 2015-12-13
Normalize line endings to lf

Shredder121:
This should also prevent Windows line ending issues.

We don't really need anything with `crlf` line endings, do we?
Thoughts? Problem cases?


robertandrewbain:
Looks good to me.

timowest:
Looks good

johnktims:
I'd remove the outdated performance documents (e.g. querydsl-collections/doc/perf-20090313.txt).

Shredder121:
Done.

PR #1720, opened 2015-12-20
Better preserve original Expression type

timowest:


Shredder121:
Looks good.
Little late, sorry Timo.

johnktims:
Looks good.

PR #1722, opened 2015-12-27
Remove reference to junit.framework

johnktims:


Shredder121:
You could also add a checkstyle configuration for [`IllegalImport`](http://checkstyle.sourceforge.net/config_imports.html#IllegalImport) if you feel like it.
Or if you think that's not necessary that's fine too.

```xml
<module name="IllegalImport">
    <property name="illegalPkgs" value="junit.framework, ..." />
</module>

```

johnktims:
Good idea. I'll add that.

Shredder121:
Looks good :smile:

timowest:
Looks good

PR #1725, opened 2015-12-28
Add support for boosting Lucene queries

Shredder121:


Shredder121:
Let me know what you think.

querydsl-lucene3/src/main/java/com/querydsl/lucene3/BoostElement.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+package com.querydsl.lucene3;
+
+import com.querydsl.core.types.Constant;
+import com.querydsl.core.types.ConstantImpl;
+import com.querydsl.core.types.Expression;
+import com.querydsl.core.types.dsl.BooleanOperation;
+
+/**
+ * {@code BoostElement} is an element that allows boosting a Query.
```
johnktims: Doesn't it just boost a term?

johnktims:
It looks fine to me. Are you going to add this functionality for versions 4 and 5?

Do you think we need an integration test?

Shredder121:
Yes I am, but I figured it would be good to also hear from Timo.

Besides, for 5 they changed the API to use decoration instead of setters to mixin the boost factor.

querydsl-lucene3/src/main/java/com/querydsl/lucene3/LuceneSerializer.java
```diff
@@ -120,6 +120,12 @@ private Query toQuery(Operation<?> operation, QueryMetadata metadata) {
             @SuppressWarnings("unchecked") //This is the expected type
             Query rv = ((Constant<Query>) operation.getArg(0)).getConstant();
             return rv;
+        } else if (op == LuceneOps.BOOST) {
+            BoostElement element = (BoostElement) operation;
```
timowest: this is not a safe cast, since the mixin type is just an operation, you need to get the predicate and boost factor from the operation

Shredder121: Okay, I thought I'd beat the system, but you're right.
I'll change it back to the usual stuff.

johnktims:
@Shredder121 is there anything left to do here?

PR #1732, opened 2016-01-05
Normalize line endings to lf

Shredder121:
'Backport' of #1711


querydsl-codegen/src/main/java/com/mysema/query/codegen/AnnotationHelper.java
```diff
@@ -1,49 +1,49 @@
...
-    /**
-     * Get the {@link TypeCategory} according with object Class and Annotation.
-     * @param cl Class of type.
-     * @param annotation Annotation found on element.
-     * @return Custom {@link TypeCategory}.
-     */
-    TypeCategory getTypeByAnnotation(Class<?> cl, Annotation annotation);
-}
+/*
+ * Copyright 2015, The Querydsl Team (http://www.querydsl.com/team)
```
johnktims: Should this be bumped to 2016?

timowest:
Looks good

querydsl-codegen/src/main/java/com/mysema/query/codegen/AnnotationHelper.java
Shredder121: I don't think that's necessary, nothing has changed.
What does @timowest think?

timowest: I don't think the bumping is necessary

johnktims:
LGTM

Shredder121:
Thanks!

PR #1734, opened 2016-01-06
Add getBatchCount for SQL DML clauses

### LINKED ISSUE #1700: SQLDeleteClause.isEmpty()
SQLDeleteClause seems to lack isEmpty method. It is very handy when using batched 0-to-n deletes. It would be nice to have this also for 3.x version.

timowest:
Fixes #1700


Shredder121:
This looks good to me.
Did you also want to add a test for this? (that only after a specific point the batchCount is incremented?)

johnktims:
Looks good.

PR #1740, opened 2016-01-08
Improve addLastJoin behaviour

### LINKED ISSUE #1737: SQLQuery.toString should not have side effects
Calling ProjectableSQLQuery.toString registers pending join clauses as a side effect. If one happens to call query.toString() between join() and on() then the query breaks as a result: 

    query.from(x);
    query.leftJoin(y) // toString() call e.g. by debugger
    .on(x.foo.equals(y.foo));
    query.leftJoin(z).on(z.foo.equals(y.foo));

Results in 

    from x 
    left join y 
    left join z on x.foo = y.foo and z.foo = y.foo

ToString with side effects is especially disastrous with modern debuggers as they call toString every time execution pauses. This makes debugging query builders very hard and error-prone.

timowest:
Fixes #1737

### DIFF
--- a/querydsl-core/src/test/java/com/querydsl/core/DefaultQueryMetadataTest.java
+++ b/querydsl-core/src/test/java/com/querydsl/core/DefaultQueryMetadataTest.java
@@ -19,6 +19,7 @@ import java.util.Arrays;
 
 import org.junit.Test;
 
+import com.google.common.collect.ImmutableSet;
 import com.querydsl.core.QueryFlag.Position;
 import com.querydsl.core.types.ConstantImpl;
 import com.querydsl.core.types.dsl.Expressions;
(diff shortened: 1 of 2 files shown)


querydsl-core/src/test/java/com/querydsl/core/DefaultQueryMetadataTest.java
```diff
@@ -17,6 +17,7 @@
 
 import java.util.Arrays;
 
+import com.google.common.collect.ImmutableSet;
```
Shredder121: Did you change your import settings?

timowest: Sorry, I forgot to configure them on my new machine

Shredder121: Ah, that explains it.
I was just wondering.

Shredder121:
Looks good now.

robertandrewbain:
Looks good.

PR #1751, opened 2016-01-17
Unify SQLQueryFactory signatures

timowest:


querydsl-sql/src/main/java/com/querydsl/sql/AbstractSQLQueryFactory.java
```diff
@@ -90,4 +91,60 @@ public final Connection getConnection() {
...
+     */
+    public abstract AbstractSQLQuery<Integer, ?> selectZero();
+
+    /**
+     * Create a new SQLQuery instance with one as the projection
+     *
+     * @return select(1)
+     */
+    public abstract AbstractSQLQuery<Integer, ?> selectOne();
+    /**
```
Shredder121: What do you think of adding a space here?

johnktims:
It looks like the subclasses are overriding the abstract methods in `AbstractSQLQueryFactory`'s with the same implementation. Maybe that functionality should be pulled up into the parent class.

Shredder121:
They are overridden to specify the new return type.

Shredder121:
I think this now looks good.

querydsl-sql/src/main/java/com/querydsl/sql/AbstractSQLQueryFactory.java
```diff
@@ -90,4 +91,61 @@ public final Connection getConnection() {
         return connection.get();
     }
 
+    /**
+     * Create a new SQLQuery instance with the given projection
```
robertandrewbain: It's not returning a `SQLQuery` instance, it's returning a concrete impl of `AbstractSQLQuery` instance. Comment applies throughout changes to class.

querydsl-sql/src/main/java/com/querydsl/sql/AbstractSQLQueryFactory.java
```diff
@@ -90,4 +91,61 @@ public final Connection getConnection() {
...
+     *
+     * @param expr projection
+     * @param <T>
```
robertandrewbain: Param is Expression<T>, not <T>

robertandrewbain: Incorrect param type

querydsl-sql/src/main/java/com/querydsl/sql/AbstractSQLQueryFactory.java
```diff
@@ -90,4 +91,61 @@ public final Connection getConnection() {
...
+     * @return select(expr)
```
robertandrewbain: Elaborate

querydsl-sql/src/main/java/com/querydsl/sql/AbstractSQLQueryFactory.java
```diff
@@ -90,4 +91,61 @@ public final Connection getConnection() {
...
+     *
+     * @return select(1)
+     */
+    public abstract AbstractSQLQuery<Integer, ?> selectOne();
+
+    /**
+     * Create a new SQLQuery instance with the given projection and source
+     *
+     * @param expr query source and projection
+     * @param <T>
```
robertandrewbain: ?

timowest: because the type parameter is also used in the return value

timowest: AbstractSQLQuery has T for the projection type and Q for the concrete subclass type. Q is necessary for fluent method call invocations.

johnktims:
This looks good now.

PR #1755, opened 2016-01-19
Add test

timowest:


Shredder121:
This also looks fine to me.
Does the same bug exist in master?

timowest:
No, but I added this test to assert that as well.

Shredder121:
Ah I see, yes that's a good idea.

johnktims:
LGTM

PR #1756, opened 2016-01-20
Add convert tuple to map

dubu:
In vtl (velocity template engine )
get (0) to get the value of this approach in arrary form as discomfort.

It adds a tuple convert map feature so that you can use the value value to the string key.

### DIFF
--- a/querydsl-core/src/main/java/com/querydsl/core/Tuple.java
+++ b/querydsl-core/src/main/java/com/querydsl/core/Tuple.java
@@ -17,6 +17,8 @@ import javax.annotation.Nullable;
 
 import com.querydsl.core.types.Expression;
 
+import java.util.Map;
+
 /**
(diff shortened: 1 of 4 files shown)


timowest:
There is already QMap which provides a Expression to Value mapping in Map form. The toString() of the expression is also not guaranteed to be unique, so I wouldn't promote such a mapping in the core module.

In your case it might be best your own FactoryExpression that populates a Map with expr.toString() as keys.

dubu:
Thank you for your advice.

The following information has been helpful in my case.

```
List<Map<Expression<?>,?> result = query.from(employee).select(Projections.map(employee.firstName, employee.lastName)).fetch();
 for (Tuple row : result) {
     System.out.println("firstName " + row.get(employee.firstName));
     System.out.println("lastName " + row.get(employee.lastName));
 }

```

PR #1758, opened 2016-01-20
Fix modulo normalization

timowest:
Fixes https://github.com/querydsl/querydsl/issues/1759

### DIFF
--- a/querydsl-core/src/main/java/com/mysema/query/support/Normalization.java
+++ b/querydsl-core/src/main/java/com/mysema/query/support/Normalization.java
@@ -27,11 +27,11 @@ public final class Normalization {
     // TODO simplify
     private static final Pattern FULL_OPERATION = Pattern.compile(
             "(?<![\\d*/\"?' ])" + "(\\b|\\(|\\s+)"  +
-            "(" + NUMBER + WS + "[+\\-/*]" + WS + ")+" + NUMBER + WS +
+            "(" + NUMBER + WS + "[+\\-/*%]" + WS + ")+" + NUMBER + WS +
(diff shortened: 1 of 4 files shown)


johnktims:
Looks good.

Shredder121:
Keep in mind that since it now also uses `%` to look for operations, it triggers earlier (when using like `%`, it will look for operations, so maybe some more tests?)

timowest:
good point with the like patterns, I'll add some tests

Shredder121:
Also, If I add this:

```java
@Test
@Category(Performance.class)
public void Performance2() throws Exception {
    Runner.run("NormalizationTest Performance2", new Benchmark() {
        @Override
        public void run(int times) throws Exception {
            for (int i = 0; i < times; i++) {
                Normalization.normalize("select name from product where description like '%on sale%'");
            }
        }
    });
}
```

Then my test output becomes this:

```
- - NormalizationTest Performance2 5,230 s
- - NormalizationTest Performance 49,50 ms
```

Do we do anything about that? Or do we think that correctness is key and take the hit?

timowest:
The fix for various issues related to normalization is in Querydsl 4. I'd say we focus on correctness for Querydsl 3. But easy improvements to performance can of course be applied.

Shredder121:
Okay, just wanted you to be aware of that.

Shredder121:
And on that note, LGTM.
Shall I merge it or does someone else want to?

timowest:
You can merge it.

Shredder121:
I figured maybe John's hands were itching to merge some PRs, but I'll do this one then.

PR #1767, opened 2016-01-25
Use the keyword list from SourceVersion

Shredder121:
Comments appreciated.

This way we don't have to maintain the reserved word list.

### DIFF
--- a/querydsl-core/src/main/java/com/querydsl/core/util/JavaSyntaxUtils.java
+++ b/querydsl-core/src/main/java/com/querydsl/core/util/JavaSyntaxUtils.java
@@ -13,9 +13,7 @@
  */
 package com.querydsl.core.util;
 
-import java.util.Set;
-
-import com.google.common.collect.ImmutableSet;
+import javax.lang.model.SourceVersion;
 
 /**
  * JavaSyntaxUtils defines reserved keywords for Java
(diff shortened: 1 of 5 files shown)


timowest:
Looks good

Shredder121:
Then, @johnktims what would you change regarding the Javadoc?

Shredder121:
Or @robertandrewbain of course

querydsl-core/src/main/java/com/querydsl/core/util/JavaSyntaxUtils.java
```diff
@@ -27,63 +25,8 @@
 
     private JavaSyntaxUtils() { }
```
robertandrewbain: Why not delete this class, and just call `SourceVersion.isKeyword`? This is just a static utility class that wraps a call to a static method with the same signature.

Shredder121: I'm not sure whether end users use it :P 
I'd rather keep it for a while?
Or what does @timowest think?

timowest: JavaSyntaxUtils is public, so we can't delete it in 4.*.

Shredder121: Shall we then deprecate it?
Then the next cleanup with the next major release it will pop up.

timowest: Yes, maybe add the deprecation and replace the usage with SourceVersion.isKeyword

robertandrewbain:
I like the deprecation - that's the solution! +1

querydsl-core/src/main/java/com/querydsl/core/util/JavaSyntaxUtils.java
timowest: Yes, deprecation is fine.

PR #1772, opened 2016-01-30
Improve addLastJoin behaviour

### LINKED ISSUE #1737: SQLQuery.toString should not have side effects

timowest:
Fixes #1737

### DIFF
--- a/querydsl-core/src/main/java/com/mysema/query/DefaultQueryMetadata.java
+++ b/querydsl-core/src/main/java/com/mysema/query/DefaultQueryMetadata.java
@@ -22,6 +22,7 @@ import com.google.common.base.Objects;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Lists;
 import com.mysema.query.types.*;
 import static com.mysema.query.util.CollectionUtils.*;
 
@@ -113,7 +114,6 @@ public class DefaultQueryMetadata implements QueryMetadata, Cloneable {
 
     @Override
     public void addGroupBy(Expression<?> o) {
-        addLastJoin();
         // group by elements can't be validated, since they can refer to projection elements
         // that are declared later
         groupBy = add(groupBy, o);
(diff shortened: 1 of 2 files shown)


querydsl-core/src/main/java/com/mysema/query/DefaultQueryMetadata.java
```diff
@@ -385,8 +386,7 @@ public boolean equals(Object o) {
 
     @Override
     public int hashCode() {
-        addLastJoin();
-        return Objects.hashCode(flags, groupBy, having, joins, modifiers,
+        return Objects.hashCode(flags, groupBy, having, getJoins(), modifiers,
                 orderBy, params, projection, unique, where);
```
robertandrewbain: I note that `where` is in hashCode() but not `equals(Object o)`. I would have expected to see it in neither or both.

querydsl-core/src/test/java/com/mysema/query/DefaultQueryMetadataTest.java
```diff
@@ -94,6 +95,20 @@ public void GetJoins2() {
     }
 
     @Test
+    public void GetJoins3() {
```
robertandrewbain: Poorly named.

querydsl-core/src/main/java/com/mysema/query/DefaultQueryMetadata.java
timowest: see line 380

querydsl-core/src/test/java/com/mysema/query/DefaultQueryMetadataTest.java
timowest: I agree, but this is a backport, in general we do only fixes in Querydsl 3, and not stylistic improvements

PR #1778, opened 2016-01-31
Bump version

timowest:


Shredder121:
Why now of all times, Travis?

PR #1782, opened 2016-02-02
When MathUtils.cast gets a null input, return null

### LINKED ISSUE #1781: NullPointerException with NumberExpression.in
I want to express the following SQL expression: 

```sql
select * from table where table.number_column in (1, 2, null);
```

When I try this with Querydsl like this:

```java
factory
    .select(table.all())
    .from(table)
    .where(table.numberColumn.in(asList(1, 2, null)));
```

Then a `NullPointerException` is thrown.

```java
java.lang.NullPointerException
	at com.querydsl.core.util.MathUtils.cast(MathUtils.java:52)
	at com.querydsl.core.types.dsl.NumberExpression.convert(NumberExpression.java:777)
	at com.querydsl.core.types.dsl.NumberExpression.in(NumberExpression.java:766)
	at com.querydsl.core.types.dsl.NumberExpression.in(NumberExpression.java:33)
```

I am using version 4.0.3 of Querydsl.

Shredder121:
Closes #1781 

You can cast `null` to any type, so this should be the least surprising change..
What do you think?

### DIFF
--- a/querydsl-core/src/main/java/com/querydsl/core/util/MathUtils.java
+++ b/querydsl-core/src/main/java/com/querydsl/core/util/MathUtils.java
@@ -60,7 +60,7 @@ public final class MathUtils {
 
     public static <D extends Number> D cast(Number num, Class<D> type) {
         D rv;
-        if (type.isInstance(num)) {
+        if (num == null || type.isInstance(num)) {
             rv = type.cast(num);
         } else if (type.equals(Byte.class)) {
             rv = type.cast(num.byteValue());
--- a/querydsl-core/src/test/java/com/querydsl/core/util/MathUtilsTest.java
+++ b/querydsl-core/src/test/java/com/querydsl/core/util/MathUtilsTest.java
@@ -13,8 +13,7 @@
  */
 package com.querydsl.core.util;
 
-import static org.junit.Assert.assertEquals;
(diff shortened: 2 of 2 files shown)


timowest:
Looks good

ThijsWouters:
Now there is an IllegalArgumentException:

```java
java.lang.IllegalArgumentException: eq(null) is not allowed. Use isNull() instead
	at com.querydsl.core.types.dsl.SimpleExpression.eq(SimpleExpression.java:127)
	at com.querydsl.core.types.dsl.SimpleExpression.in(SimpleExpression.java:192)
	at com.querydsl.core.types.dsl.NumberExpression.in(NumberExpression.java:769)
	at com.querydsl.core.types.dsl.NumberExpression.in(NumberExpression.java:33)
```

ThijsWouters:
It seems that I actually wanted to express: 
```sql
select * from table where table.number_column in (null);
```

Is there a reason why you would prefer `=` in favor of `in` when there is only one value?

ThijsWouters:
I did some tests with Postgresql and it seems that both statements generate the same plan.

timowest:
We can maybe fix the expr in single arg case in a separate PR?

ThijsWouters:
Of course. I would not mind that.

johnktims:
Looks good to me.

PR #1789, opened 2016-02-09
Fix listener handling

### LINKED ISSUE #1786: Pooled Connection doesn't work in SQLMergeClause with compositeMerge
Dear all,

I really appreciate your improvements regarding connection pooling in the last version. Currently, I am using QueryDSL 4.0.7 at a client's project and experienced some issue while using BasicDataSources for our Oracle-connectivity and the new connection pooling mechanism of QueryDSL. In detail, I would like to use the SQLMergeClause, but it doesn't work. The OracleTemplates don't support nativeMerge (isNativeMerge() returns false). During the invocation of the SQLMergeClause.execute()-method the method executeCompositeMerge() is run and here is the problem. This method begins with an invocation of hasRow() which closes the used connection. But this connection keeps referenced and is reused for the following update or insert query of that method. The result is "SQLExeption: Connection is null". From my point of view the connection has to be nulled in the SQLMergeClause when it is going to be closed. In this case the DataSource would deliver a new connection from its pool.
Are you aware of this problem? Or do you have any suggestion how I can fix this problem?

Thank you and kind regards
Lukas

timowest:
Fixes #1786

### DIFF
--- a/querydsl-sql/src/main/java/com/querydsl/sql/SQLListeners.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/SQLListeners.java
@@ -245,4 +245,9 @@ public class SQLListeners implements SQLDetailedListener {
             listener.exception(context);
         }
     }
+
+    public Set<SQLDetailedListener> getListeners() {
+        return listeners;
+    }
+
 }
--- a/querydsl-sql/src/main/java/com/querydsl/sql/SQLNoCloseListener.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/SQLNoCloseListener.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2015, The Querydsl Team (http://www.querydsl.com/team)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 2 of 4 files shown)


querydsl-sql/src/main/java/com/querydsl/sql/SQLListeners.java
```diff
@@ -245,4 +245,9 @@ public void exception(final SQLListenerContext context) {
             listener.exception(context);
         }
     }
+
+    public Set<SQLDetailedListener> getListeners() {
```
Shredder121: Does this specifically need to be a `Set`?
Just wondering if you had plans with this.

Shredder121:
Looks good.

querydsl-sql/src/main/java/com/querydsl/sql/SQLListeners.java
timowest: I like the emphasis on no duplicates instead of ordering, although a LinkedHashSet is used

Shredder121: Okay, since we only iterate over it to copy it, so `Iterable` would work, but if you need to add anything `Collection` would work as well.
Keeping it `Set` is also an option, I was just wondering though.

johnktims:
Looks good to me, too.

PR #1790, opened 2016-02-09
Add clear method to DML clauses

### LINKED ISSUE #1785: StoreClause: Clear Batches on Execute
Would it be possible to clear batches on execute, so that it's easy to configure when to flush? Currently we get OutOfMemoryError because of a large batch and the simplest course of action would be just to flush more often.

timowest:
Fixes #1785


timowest:
TODO: tests

Shredder121:
Looks good.

Shredder121:
I was thinking of maybe extracting a common AbstractDMLTest class, but that's not really better I think?

timowest:
> I was thinking of maybe extracting a common AbstractDMLTest class, but that's not really better I think?

Yes, that could be a followup PR if we want to test something more consistenly.

johnktims:
Looks good, Timo.

PR #1797, opened 2016-02-17
Fix custom type serialization

### LINKED ISSUE #1796: SQL MetaDataExporter with custom sql types
I've a use case where I would like the use the PostgreSQL full text support ('tsvector' sql type) within the context of a QueryDSL query.

In order to be able to do this I had to implement the following classes:

- querydsl.TsVector
- querydsl.TsVectorExpression (contains methods to build the expressions we need for our use cases)
- querydsl.TsVectorPath

With these classes I can create every query I need. 

(I'm currently only interested in creating custom expressions on 'tsvector' columns. Selecting, inserting or updating is not required and I'm aware how to implement this in case I do need it.) 

Obviously the code generator of QueryDSL doesn't know about these classes. In order to avoid writing my own metamodel classes by hand I investigated the possibility to have the MetaDataExporter generate classes with fields using the custom TsPath type. I came up with the following:

```java
Connection c = ...;

SQLTemplates templates = new PostgreSQLTemplates();
Configuration configuration = new Configuration(templates);
configuration.registerType("tsvector", querydsl.TsVector.class)

TypeMappings typeMappings = new JavaTypeMappings();
Type type = new ClassType(querydsl.TsVector.class);
typeMappings.register(type, new ClassType(querydsl.TsVector.class, type));

MetaDataExporter exporter = new MetaDataExporter();
exporter.setConfiguration(configuration);
exporter.setTypeMappings(typeMappings);
...
exporter.export(c.getMetaData());
```

This setup results in a metamodel that is very close to what I would like it to be. It results in fields like this:

```java
public final querydsl.TsVectorPath<querydsl.TsVector> tsv = createTsVector("tsv", querydsl.TsVector.class);
```

Unfortunately this doesn't compile because 'createTsVector' doesn't exist. While analyzing the source of the code generator I discovered that this behavior (i.e. expecting these create methods to be available for custom types) is actually hard coded (https://github.com/querydsl/querydsl/blob/master/querydsl-sql-codegen/src/main/java/com/querydsl/sql/codegen/MetaDataSerializer.java#L321).

I figured that in order to get this code compiling, all I had to do was figure out a way to include an 'import static' that brings an appropriate 'creatTsVector' method in scope. The MetaDataExporter doesn't have a setting to include an additional 'import static' but it does have one to include an ordinary 'import' that can be used to trick the code generator into generating a 'static import':

```java
exporter.setImports(new String[]{"static querydsl.factory"});
```

results in:

```java
import static querydsl.factory.*;
```

Which does exactly what I need, provided there is a class named 'querydsl.factory' (with lowercase 'f' to trick the code generator into thinking that I want to import a package, resulting in '.*' to be added) with:

```java
public static <T extends querydsl.TsVector> querydsl.TsVectorPath<T> createTsVector(String var, Class<T> type) {
...
}
```

Although this solution works very well, it does feel like an ugly hack. I've been looking for a better strategy but everything I came up with (e.g. implementing my own Serializer, extending MetaDataExporter or creating a SQLCodegenModule.ENTITYPATH_TYPE like was done to implement spatial support) required a significant amount of additional work.

It would be great if support for using custom database types in the code generator would be easier and fully supported. It would enable me to leverage the power of specialized sql types (there are quite a few of those in the PostgreSQL universe!) without loosing the ability to use QueryDSL to create my queries.

timowest:
Fixes #1796

### DIFF
--- a/querydsl-sql-codegen/src/main/java/com/querydsl/sql/codegen/MetaDataSerializer.java
+++ b/querydsl-sql-codegen/src/main/java/com/querydsl/sql/codegen/MetaDataSerializer.java
@@ -321,11 +321,15 @@ public class MetaDataSerializer extends EntitySerializer {
 
     @Override
     protected void customField(EntityType model, Property field, SerializerConfig config,
-            CodeWriter writer) throws IOException {
+                               CodeWriter writer) throws IOException {
         Type queryType = typeMappings.getPathType(field.getType(), model, false);
-        String localRawName = writer.getRawName(field.getType());
-        serialize(model, field, queryType, writer, "create" + field.getType().getSimpleName(),
-                writer.getClassConstant(localRawName));
+        if (queryType.getPackageName().startsWith("com.querydsl")) {
+            String localRawName = writer.getRawName(field.getType());
+            serialize(model, field, queryType, writer, "create" + field.getType().getSimpleName(),
+                    writer.getClassConstant(localRawName));
+        } else {
+            super.customField(model, field, config, writer);
+        }
     }
 
     protected void serializePrimaryKeys(EntityType model, CodeWriter writer,
--- a/querydsl-sql-codegen/src/test/java/com/querydsl/sql/codegen/MetaDataSerializerTest.java
+++ b/querydsl-sql-codegen/src/test/java/com/querydsl/sql/codegen/MetaDataSerializerTest.java
@@ -13,6 +13,8 @@
  */
 package com.querydsl.sql.codegen;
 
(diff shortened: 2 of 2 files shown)


copierrj:
Thanks for the quick response to by bug report. 

While doing my analysis for #1796 I noticed that the method removed in this PR was introduced when spatial support was added to the code generator. When spatial support is enabled this method creates the appropriate calls to RelationalPathSpatial.createGeometry. 

What does the output of the code generator for custom types look like after this PR is merged?

Shredder121:
How can we reliably test this?

PR #1798, opened 2016-02-17
Improve String id handling

### LINKED ISSUE #1795: MongodbSerializer fails when entity id is assigned a String and is not an ObjectId
This was introduced in the following changelist https://github.com/querydsl/querydsl/commit/3d09b9581bf1ef2c52118364c372ce640ec7f844

The following code expects entity Ids to be object Ids:

```java
protected Object convert(Path<?> property, Constant<?> constant) {
	        if (isReference(property)) {
	            return asReference(constant.getConstant());
	        } else if (isId(property)) {
	            if (isReference(property.getMetadata().getParent())) {
	                return asReferenceKey(property.getMetadata().getParent().getType(), constant.getConstant());
	            } else if (constant.getType().equals(String.class)) {
	                return new ObjectId((String) constant.getConstant());
	            }
	        }
	        return visit(constant, null);
	    }
```

Here is a simple unit test that fails:

```java
@Entity
public class WithStringId {
	@Id	String id;
}
```

```java
	@Test
	public void test_QueryDSL_StringIds()
	{
		IPrimaryMongoConnector mongo = ClassContainer.getInstance(IPrimaryMongoConnector.class);

		Morphia morphia = new Morphia();
		Datastore ds = morphia.createDatastore(mongo.getMongo(), "sampledb"); 
		MorphiaSerializer ser = new MorphiaSerializer(new Morphia());
		
		MorphiaQuery<WithStringId> query = new MorphiaQuery<>(morphia, ds, WithStringId.class);
		query = query.where(QWithStringId.withStringId.id.eq("a string value"));
		query.toString();
	}
```

An exception is thrown in the constructor of ObjectId, called in the MorphiaSerializer on the following line:

```java
                return new ObjectId((String) constant.getConstant());
```
In this case, the code seems to be a special case for when a String expression is provided for an ObjectId instead of an actual ObjectId.  The problem is that it does not ensure that the property type is an ObjectId.  A check is missing to make sure that property.type is indeed an ObjectId.

I would suggest the following change but I haven't tested:

```java
            } else if (constant.getType().equals(String.class)) {
```

should be

```java
            } else if (property.getType.equals(ObjectId.class) && constant.getType().equals(String.class)) {
```

Gaspard

timowest:
Fixes #1795

### DIFF
--- a/querydsl-mongodb/src/main/java/com/querydsl/mongodb/MongodbSerializer.java
+++ b/querydsl-mongodb/src/main/java/com/querydsl/mongodb/MongodbSerializer.java
@@ -305,11 +305,16 @@ public abstract class MongodbSerializer implements Visitor<Object, Void> {
         } else if (isId(property)) {
             if (isReference(property.getMetadata().getParent())) {
                 return asReferenceKey(property.getMetadata().getParent().getType(), constant.getConstant());
-            } else if (constant.getType().equals(String.class)) {
-                return new ObjectId((String) constant.getConstant());
+            } else if (constant.getType().equals(String.class) && isImplicitObjectIdConversion()) {
+                String id = (String) constant.getConstant();
+                return ObjectId.isValid(id) ? new ObjectId(id) : id;
             }
         }
         return visit(constant, null);
+    }
+
+    protected boolean isImplicitObjectIdConversion() {
+        return true;
     }
 
     protected DBRef asReferenceKey(Class<?> entity, Object id) {
--- a/querydsl-mongodb/src/main/java/com/querydsl/mongodb/morphia/MorphiaSerializer.java
+++ b/querydsl-mongodb/src/main/java/com/querydsl/mongodb/morphia/MorphiaSerializer.java
@@ -71,6 +71,12 @@ public class MorphiaSerializer extends MongodbSerializer {
     }
 
     @Override
+    protected boolean isImplicitObjectIdConversion() {
+        // see https://github.com/mongodb/morphia/wiki/FrequentlyAskedQuestions
+        return false;
+    }
+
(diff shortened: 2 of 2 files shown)


gaspardpetit:
Hi Timo,

Thanks for the improvement. I am not convinced however that the fix will address the case where the id is a String type and not an ObjectId.  It seems to me that with this change, if the string value happens to be a valid ObjectId (whether intended or by chance), the field will be assigned an ObjectId type even though it is a String (i.e. {value: {$oid: "theid"}} instead of plainly {value: "theid"}). The serialized query might be valid for the DB, but the type associated type of the value in the database will be incorrect.

I will test it on my end and get back to you.

Gaspard

timowest:
For Morphia the implicit conversion was removed
https://github.com/mongodb/morphia/wiki/FrequentlyAskedQuestions

But for Spring Data it's still used
http://docs.spring.io/spring-data/mongodb/docs/1.3.3.RELEASE/reference/html/mapping-chapter.html
(section 7.1.1)

gaspardpetit:
I see, thanks for the clarification, I did not realize Spring Data had implicit conversion to ObjectId when the value is a compatible String.  To me this is clearly something that should not happen, but I understand how QueryDSL is stuck in the middle.

timowest:
You can take a look at the PR. The default is now implicit conversion, but for the Morphia setup it's disabled. That will be incompatible with older versions of Morphia, but I didn't find a simple way to get the runtime version number of Morphia in the code.

Shredder121:
I also didn't find a good way to get the version.
But this looks fine to me.

johnktims:
Looks good to me.

PR #1802, opened 2016-02-20
Fix mapping

### LINKED ISSUE #1801: bug in SQLTemplatesRegistry and MS SQL Server
QueryDSL version - **4.0.7**

JDBC Driver:
com.microsoft.sqlserver
sqljdbc
**4.1**

SQLTemplatesRegistry does not determine right way SQLTemplates for MS SQL Server 
because misspelling:
```java
        // sqlserver
        if (name.equals("**microsft sql server**")) {
            switch (md.getDatabaseMajorVersion()) {
                case 12:
                case 11: return sqlserver2012;
                case 10: return sqlserver2008;
                case 9:  return sqlserver2005;
                default: return sqlserver;
            }
        }
```
it must be "microsoft sql server". At least my driver gives this string for md.getDatabaseProductName().toLowerCase();

timowest:
Fixes #1801

### DIFF
--- a/querydsl-sql/src/main/java/com/querydsl/sql/SQLTemplatesRegistry.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/SQLTemplatesRegistry.java
@@ -81,7 +81,7 @@ public class SQLTemplatesRegistry {
         }
 
         // sqlserver
-        if (name.equals("microsft sql server")) {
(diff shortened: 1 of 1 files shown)


Shredder121:
This looks fine to me.

johnktims:
Looks good.

PR #1811, opened 2016-03-05
Fix empty value handling

### LINKED ISSUE #1810: Don't skip null prefix/suffix values in Maven goals
Related to https://groups.google.com/forum/#!topic/querydsl/HupqtP28LRM

Also related http://mail-archives.apache.org/mod_mbox/maven-users/200708.mbox/%3C5a2cf1f60708090246l216f156esf46cc1e968b37ccd@mail.gmail.com%3E

timowest:
Fixes #1810

### DIFF
--- a/querydsl-maven-plugin/src/main/java/com/querydsl/maven/AbstractMetaDataExportMojo.java
+++ b/querydsl-maven-plugin/src/main/java/com/querydsl/maven/AbstractMetaDataExportMojo.java
@@ -395,7 +395,7 @@ public class AbstractMetaDataExportMojo extends AbstractMojo {
             }
 
             MetaDataExporter exporter = new MetaDataExporter();
-            exporter.setNamePrefix(Strings.nullToEmpty(namePrefix));
+            exporter.setNamePrefix(emptyIfSetToBlank(namePrefix));
             exporter.setNameSuffix(Strings.nullToEmpty(nameSuffix));
             exporter.setBeanPrefix(Strings.nullToEmpty(beanPrefix));
             exporter.setBeanSuffix(Strings.nullToEmpty(beanSuffix));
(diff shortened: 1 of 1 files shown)


Shredder121:
This doesn't actually fix it.
Looking into a fix.

Shredder121:
[I found the culprit][culprit].
The value must be nonempty, otherwise it falls back to the `default-value`.

What we could do is maybe recognize `BLANK` (case insensitive?).

[culprit]: http://grepcode.com/file/repo1.maven.org/maven2/org.eclipse.sisu/org.eclipse.sisu.plexus/0.3.0/org/codehaus/plexus/component/configurator/converters/basic/AbstractBasicConverter.java#AbstractBasicConverter.fromExpression%28org.codehaus.plexus.configuration.PlexusConfiguration%2Corg.codehaus.plexus.component.configurator.expression.ExpressionEvaluator%2Cjava.lang.Class%29

Shredder121:
It's been about a month ago, what's your view on my suggestion?

timowest:
I can't come up with anything better, so `BLANK`is fine. Can you do the change?

querydsl-maven-plugin/src/main/java/com/querydsl/maven/AbstractMetaDataExportMojo.java
```diff
@@ -394,18 +395,10 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
-            if (beanPrefix != null) {
-                exporter.setBeanPrefix(beanPrefix);
-            }
-            if (beanSuffix != null) {
-                exporter.setBeanSuffix(beanSuffix);
-            }
+            exporter.setNamePrefix(emptyIfSetToBlank(namePrefix));
+            exporter.setNameSuffix(Strings.nullToEmpty(nameSuffix));
+            exporter.setBeanPrefix(Strings.nullToEmpty(beanPrefix));
+            exporter.setBeanSuffix(Strings.nullToEmpty(beanSuffix));
```
Shredder121: I figured changing it also for these is not necessary, since only `namePrefix` has a `default-value` set to a non-empty String.

Shredder121:
I was looking into a custom `@configurator`, using an extra doclet `@nullable` that would actually set things to null if the aren't supplied.
Alas, it didn't work well enough to warrant the change.

PR #1812, opened 2016-03-05
Add more null guards for jdbcType

### LINKED ISSUE #1809: com.querydsl.sql.ColumnMetadata jdbcType is reference while null
The field definition in com.querydsl.sql.ColumnMetadata has
    private final Integer jdbcType;

but the getter is

    public int getJdbcType() {
        return jdbcType;
    }

but the constructor of ColumnMetadata allows null values.

in the com.querydsl.sql.Configuration#set you have this code

 public <T> void set(PreparedStatement stmt, Path<?> path, int i, T value) throws SQLException {
        if (Null.class.isInstance(value)) {
            Integer sqlType = path != null ? ColumnMetadata.getColumnMetadata(path).getJdbcType() : null;
            if (sqlType != null) {
                stmt.setNull(i, sqlType);
            } else {
                stmt.setNull(i, Types.NULL);
            }
        } else {
            getType(path, (Class) value.getClass()).setValue(stmt, i, value);
        }
    }

So it looks up column metadata but if the jdbcType is null it will NPE

The jdbcType can be null for Enum types say

com.querydsl.core.types.dsl.BeanPath#createEnum

    protected <A extends Enum<A>> EnumPath<A> createEnum(String property, Class<A> type) {
        return add(new EnumPath<A>(type, forProperty(property)));
    }

    protected <P extends Path<?>> P add(P path) {
        return path;
    }

in this case the RelationPath code does NOT add any metadata for that column and hence its "synthesises" some based on name like

    public static ColumnMetadata getColumnMetadata(Path<?> path) {
        Path<?> parent = path.getMetadata().getParent();
        if (parent instanceof EntityPath) {
            Object columnMetadata = ((EntityPath<?>) parent).getMetadata(path);
            if (columnMetadata instanceof ColumnMetadata) {
                return (ColumnMetadata) columnMetadata;
            }
        }
        return ColumnMetadata.named(path.getMetadata().getName());
    }

The end result of all this is that NPEs happen

I think the query generation code should use com.querydsl.sql.ColumnMetadata#hasJdbcType to check if in fact there is a JDBC type in play.  It appears like it wants to do this but its not happening in effect, specifically for Enum columns

timowest:
Fixes #1809

### DIFF
--- a/querydsl-sql-codegen/src/main/java/com/querydsl/sql/codegen/MetaDataSerializer.java
+++ b/querydsl-sql-codegen/src/main/java/com/querydsl/sql/codegen/MetaDataSerializer.java
@@ -225,11 +225,13 @@ public class MetaDataSerializer extends EntitySerializer {
             columnMeta.append("ColumnMetadata");
             columnMeta.append(".named(\"" + metadata.getName() + "\")");
             columnMeta.append(".withIndex(" + metadata.getIndex() + ")");
-            String type = String.valueOf(metadata.getJdbcType());
-            if (typeConstants.containsKey(metadata.getJdbcType())) {
-                type = "Types." + typeConstants.get(metadata.getJdbcType());
+            if (metadata.hasJdbcType()) {
+                String type = String.valueOf(metadata.getJdbcType());
+                if (typeConstants.containsKey(metadata.getJdbcType())) {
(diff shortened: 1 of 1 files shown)


Shredder121:
Looks good to me.

PR #1816, opened 2016-03-13
Fix precedence mappings

### LINKED ISSUE #1814: Issue with SQLServerTemplates, CollectionExpression and IN operator
QueryDSL version - 4.0.7

I've found issue with using SQLServerTemplates and CollectionExpression as an argument of IN operator.

Code to reproduce and illustrate issue:
```java
public class QueryDslInListSQLServerTemplatesIssueTest
{
	private Logger logger = LoggerFactory.getLogger(QueryDslInListSQLServerTemplatesIssueTest.class);
	
	@Test
	public void test() throws Exception
	{
		QNavigationItem item = new QNavigationItem("r");
		
		List<String> authorities = Arrays.asList("1", "2", "3");
		
		CollectionExpression<Collection<String>, String> nameConstants = null;
		for(String a : authorities)
		{
			StringTemplate t = Expressions.stringTemplate("'" + a + "'");
			if (nameConstants ==  null)
			{
				nameConstants = Expressions.collectionOperation(String.class, Ops.SINGLETON, t);
			}
			else
			{
				nameConstants = Expressions.collectionOperation(String.class, Ops.LIST, nameConstants, t);
			}
		}
		
		SQLQuery<?> query = queryFactory().query();
		query
			.select(item.id)
			.from(item)
			.where(item.name.in(nameConstants));
		
		String sql = query.getSQL().getSQL();
		
		logger.info("\n" + sql);
/*
output is:
select r.id
from navigation_item r
where r.name in '1', '2', '3'
		 */
		
//		Assert.assertThat(sql, StringContains.containsString("in (" + nameConstants.toString() + ")"));
	}
	
	{
		conf.setUseLiterals(true);
		
	}
	
	@Test
	public void testWorkaround() throws Exception
	{
		QNavigationItem item = new QNavigationItem("r");
		
		List<String> authorities = Arrays.asList("1", "2", "3");
		
		SQLQuery<?> query = queryFactory().query();
		query
			.select(item.id)
			.from(item)
			.where(item.name.in(authorities));
		
		String sql = query.getSQL().getSQL();
		
		logger.info("\n" + sql);
/*
output is:
select r.id
from navigation_item r
where r.name in ('1', '2', '3')
		 */		
		
//		Assert.assertThat(sql, StringContains.containsString("in (" + nameConstants.toString() + ")"));
	}
}
```

As you can see the test that uses CollectionExpression does not add parentheses for constants list.
This is because that for SQLServerTemplates precedence (110) of the Ops.IN is greater then precedence (70) of the Ops.LIST.

this is a bug or correct behavior?

timowest:
Fixes #1814

### DIFF
--- a/querydsl-sql/src/main/java/com/querydsl/sql/SQLServerTemplates.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/SQLServerTemplates.java
@@ -78,8 +78,8 @@ public class SQLServerTemplates extends SQLTemplates {
         setPrecedence(Precedence.ARITH_LOW, Ops.NEGATE);
         setPrecedence(Precedence.COMPARISON, Ops.EQ, Ops.EQ_IGNORE_CASE, Ops.NE);
         setPrecedence(Precedence.OR, Ops.BETWEEN, Ops.IN, Ops.NOT_IN, Ops.LIKE, Ops.LIKE_ESCAPE);
-
         setPrecedence(Precedence.OR, OTHER_LIKE_CASES);
+        setPrecedence(Precedence.OR + 1, Ops.LIST, Ops.SINGLETON);
 
         add(SQLOps.WITH_REPEATABLE_READ, "\nwith (repeatableread)");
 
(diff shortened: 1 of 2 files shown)


Shredder121:
This also looks fine to me.

PR #1822, opened 2016-03-14
Ignore unreliable test

timowest:


querydsl-jdo/src/test/java/com/querydsl/jdo/BasicsTest.java
```diff
@@ -93,10 +94,14 @@ public void list_distinct() {
 
     @Test
     public void list_distinct_two_sources() {
-        // XXX List implementation of JDO provider has weird equals implementation
-        assertEquals(
-                ImmutableList.copyOf(query().from(product, product2).select(product, product2).fetch()),
-                ImmutableList.copyOf(query().from(product, product2).distinct().select(product, product2).fetch()));
+        try {
+            // XXX List implementation of JDO provider has weird equals implementation
```
johnktims: What is the `XXX` for?

Shredder121: Then it shows up in a list of TODOs, HACKs, BUGs, XXXs, etc

Shredder121:
This looks fine to me now.

querydsl-jdo/src/test/java/com/querydsl/jdo/BasicsTest.java
Shredder121: I myself use `XXX` for unclassifiable action items, not a `BUG` or `HACK`.

timowest: Ideally we should abstract that away in Querydsl, so it's kind of an action item

Shredder121: Do you think so? Do you mean copying it over into a new list?
I don't quite know if it's worth the overhead, personally.
What's your view on this?

timowest: It's a bug in the underlying JDO implementation, and we cover lots of Hibernate bugs, so users might also expect that we cover this one.

Shredder121: Okay, I get you.

johnktims:
About time!

PR #1826, opened 2016-03-17
Serialize Expressions.TRUE and Expressions.FALSE as boolean literals

### LINKED ISSUE #1824: Wrong sql generated when using Expressions.TRUE (FALSE) for SQLServerTemplates
QueryDSL version - 4.0.7

Code to reproduce and illustrate issue:

```java
public class TrueAndFalseExpressionSQLServerTemplatesTest
{
	private Logger logger = LoggerFactory.getLogger(TrueAndFalseExpressionSQLServerTemplatesTest.class);
	
	@Test
	public void test()
	{
		QAclSid sid = QAclSid.aclSid;
		
		SQLQuery<?> query = queryFactory()
				.select(sid.all())
				.from(sid)
				.where( sid.principal.eq(Expressions.TRUE) )
				;
		
		String sql = query.getSQL().getSQL();
		
		logger.info("\n" + sql);
	}
	
	{
		conf.setUseLiterals(true);
		
	}
}
```

Where **sid.principal** is **BooleanPath** and **sql** is
`
select acl_sid.id, acl_sid.principal, acl_sid.sid
from acl_sid acl_sid
where acl_sid.principal = true
`

When I run this query I getting error:

Msg 207, Level 16, State 1, Line 3
Invalid column name 'true'.

Workaround is to use **Expressions.booleanTemplate("1")** instead of **Expressions.TRUE**

timowest:
Fixes #1824

### DIFF
--- a/querydsl-sql/src/main/java/com/querydsl/sql/SQLSerializer.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/SQLSerializer.java
@@ -13,6 +13,7 @@
  */
 package com.querydsl.sql;
 
+import java.sql.Types;
 import java.util.*;
 
 import javax.annotation.Nullable;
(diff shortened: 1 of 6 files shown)


Shredder121:
Lol, it's the first time I've seen the bot comment.
Anyway, I see a lot of failures too.
Is it an idea to add something to SelectBase instead?
```java
@Test
public void select_booleanExpr3() {
    assertTrue(query().select(Expressions.TRUE).fetchFirst());
    assertFalse(query().select(Expressions.FALSE).fetchFirst());
}
```
Or maybe as an extra?

timowest:
Added both, to make sure we get a specific literal serialization.

Shredder121:
Okay, that's good.

PR #1835, opened 2016-03-29
Improve limit/offset handling

### LINKED ISSUE #1834: Wrong sql generated when using offset and no order by clause specified
I'm getting error form MS SQL Server when using query like that:

```java
		QAclSid sid = QAclSid.aclSid;
		
		SQLQuery<?> query = queryFactory()
				.select(sid.all())
				.from(sid)
				.limit(50)
				.offset(50);
```

error:

> The ranking function "row_number" must have an ORDER BY clause.

perhaps it makes sense to add automiticly order by clause if it is not specified
as it is made in hibernate:

> SELECT inner_query.*, ROW_NUMBER() OVER (ORDER BY CURRENT_TIMESTAMP) as __hibernate_row_nr__ 

to avoid generating incorrect query
somthing like this:

```java
.orderBy(new OrderSpecifier<>(Order.ASC, DateExpression.currentDate()))
```

timowest:
Fixes #1834

### DIFF
--- a/querydsl-sql/src/main/java/com/querydsl/sql/SQLServer2005Templates.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/SQLServer2005Templates.java
@@ -91,6 +91,9 @@ public class SQLServer2005Templates extends SQLServerTemplates {
                 for (OrderSpecifier<?> os : metadata.getOrderBy()) {
                     rn.orderBy(os);
                 }
+                if (metadata.getOrderBy().isEmpty()) {
+                    rn.orderBy(Expressions.currentTimestamp().asc());
+                }
                 FactoryExpression<?> pr = Projections.appending(metadata.getProjection(), rn.as("rn"));
                 metadata.setProjection(FactoryExpressionUtils.wrap(pr));
                 metadata.clearOrderBy();
--- a/querydsl-sql/src/main/java/com/querydsl/sql/SQLServer2012Templates.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/SQLServer2012Templates.java
@@ -13,12 +13,15 @@
  */
 package com.querydsl.sql;
 
+import java.util.Map;
 import java.util.Set;
 
 import com.querydsl.core.QueryFlag;
 import com.querydsl.core.QueryFlag.Position;
 import com.querydsl.core.QueryMetadata;
 import com.querydsl.core.QueryModifiers;
+import com.querydsl.core.types.Expression;
+import com.querydsl.core.types.Path;
 import com.querydsl.core.types.dsl.Expressions;
 
 /**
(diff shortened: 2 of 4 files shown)


toughpheeckouse:
@timowest , I see what you fix offset handling

> +                metadata.addOrderBy(Expressions.ONE.asc());

I think it will cause an error for MS SQL Server

> Windowed functions do not support integer indices as ORDER BY clause expressions.

that's because need to order by current timestamp

toughpheeckouse:
I'm realy didn't try new fix only review it.
As for windows functions the original issue was about an error which i getting when execute query with offset but did not specify order by clause.
query is serialized for MS SQL Server 2008. configuration is very primitive:
```java
	{
		conf.setUseLiterals(true);
		
	}
```
MS SQL server does not support native implementation for offset functionality and it is implemeted using window function **row_number**(). But row_number() requires order by clause in "over" of window function.

for query in original issue I'm getting next sql:

> select * from (
>   select acl_sid.id, acl_sid.principal, acl_sid.sid, **row_number() over ()** as rn
> from acl_sid acl_sid) a where rn > 50 order by rn


if this sql run it will cause an error:

> The ranking function "row_number" must have an ORDER BY clause.

In original issue I offered a way to avoid generating such sql - if order by clause is empty add it.

In this fix I see that you add order by if it's empty

> metadata.addOrderBy(Expressions.ONE.asc());

may be I misread the changes? but if I'm rigtht and this code adds "order by 1" string to result sql
I will get the next query after this fix:

> select * from (
>    select acl_sid.id, acl_sid.principal, acl_sid.sid, **row_number() over (order by 1)** as rn
> from acl_sid acl_sid) a where rn > 50 order by rn

and it will cause an error again, but error will be

> Windowed functions do not support integer indices as ORDER BY clause expressions.

but if fixe would be

> metadata.addOrderBy(DateExpression.currentDate().asc());

no error will be raised.

I hope that I have made a clear

Excuse me for my confused thoughts and my english

timowest:
Ok, thanks for the clarifications. I'll apply the fix also to SQLServer2008Templates.

toughpheeckouse:
I did not pay attention that first fix will be made only for SQLServer2012Templates.

Thank you!

Shredder121:
Hi @timowest.
I saw that the `SQLServer2012Templates` overrides some methods, in particular related to serialization of modifiers.
However, it is skipped.

I think the same fix should be applied there then.

![`isRestricting()` == true][restricting]
![`metadata.getOrderBy()`][orderby]



[restricting]: https://cloud.githubusercontent.com/assets/4105066/14154082/06b73460-f6bb-11e5-8135-5890492b7413.png
[orderby]: https://cloud.githubusercontent.com/assets/4105066/14154314/13fbe138-f6bc-11e5-84fd-5c2d906b2c47.png

timowest:
Thanks for the heads up, Ruben! Fixed now also the DML part.

PR #1837, opened 2016-04-01
Querydsl 4.1.0

timowest:


Shredder121:
We should probably fix the Scala build matrix.
But, for now this looks fine to me.

timowest:
I agree

Issue #1838, opened 2016-04-04
Projection not working


damienhollis:
Hi,

We have a query that looks similar to the following:

`getHibernateQueryFactory().from(QAbstractOrderItemImpl.abstractOrderItemImpl).select(QAbstractOrderItemImpl.abstractOrderItemImpl.order.countDistinct()).fetchOne();`

But it does not return the count of distinct orders, instead it returns the count of distinct order items (this query is made up, it is obviously nonsensical way of counting all distinct orders).

The problem appears to be in the JPQLSerializer:

```java
    public void serialize(QueryMetadata metadata, boolean forCountRow, @Nullable String projection) {
        final Expression<?> select = metadata.getProjection();
        final List<JoinExpression> joins = metadata.getJoins();
        final Predicate where = metadata.getWhere();
        final List<? extends Expression<?>> groupBy = metadata.getGroupBy();
        final Predicate having = metadata.getHaving();
        final List<OrderSpecifier<?>> orderBy = metadata.getOrderBy();

        // select
        boolean inProjectionOrig = inProjection;
        inProjection = true;
        if (projection != null) {
            append(SELECT).append(projection).append("\n");

        } else if (forCountRow) {
            if (!groupBy.isEmpty()) {
                append(SELECT_COUNT_DISTINCT);
                handle(", ", groupBy);
            } else {
                if (!metadata.isDistinct()) {
                    append(SELECT_COUNT);
                } else {
                    append(SELECT_COUNT_DISTINCT);
                }
                if (select != null) {
                    if (select instanceof FactoryExpression) {
                        handle(joins.get(0).getTarget());
                    } else {
                        // TODO : make sure this works
                        handle(select);
                    }
                } else {
                    handle(joins.get(0).getTarget());
                }
            }
            append(")\n");

        } else if (select != null || !joins.isEmpty()) {
            if (!metadata.isDistinct()) {
                append(SELECT);
            } else {
                append(SELECT_DISTINCT);
            }
            if (select != null) {
                handle(select);
            } else {
                handle(joins.get(0).getTarget());
            }
            append("\n");

        }
        inProjection = inProjectionOrig;

        // from
        if (!joins.isEmpty()) {
            append(FROM);
            serializeSources(forCountRow, joins);
        }

        // where
        if (where != null) {
            append(WHERE).handle(where);
        }

        // group by
        if (!groupBy.isEmpty() && !forCountRow) {
            append(GROUP_BY).handle(COMMA, groupBy);
        }

[...]

PR #1841, opened 2016-04-06
Add test

timowest:
Related to #1838


Shredder121:
I think this looks fine.
@johnktims What do you think?

PR #1847, opened 2016-04-13
Fix orderby handling

### LINKED ISSUE #1846: ERROR: for SELECT DISTINCT, ORDER BY expressions must appear in select list
The following code:
```java
JPAQuery query = new JPAQuery(em);
query.distinct();
query.from(QPlayer.player);
query.orderBy(new OrderSpecifier(Order.ASC, QPlayer.player.team.id));
query.list(QPlayer.player, QPlayer.player.team.id);
```

generates wrong JPQL:
```sql
SELECT DISTINCT player,
                player.team.id
FROM Player player
LEFT JOIN player.team AS player_team
ORDER BY player_team.id ASC
```
In code we are using the same expression `QPlayer.player.team.id` two times: in [order by] and [select] sections. But the resulting query has different expressions:
- in [select] section we have `player.team.id`
- but in [order by] it is collapsed and looks differently: `player_team.id`

Wrong JPQL leads to wrong SQL:
```sql
SELECT DISTINCT player0_.team_id AS col_0_0_,
                player0_.id AS col_1_0_,
                ...
FROM Player player0_
LEFT OUTER JOIN Team team1_ ON player0_.team_id=team1_.id
ORDER BY team1_.id ASC
```
And this query cannot be executed due to error

> ERROR: for SELECT DISTINCT, ORDER BY expressions must appear in select list

_We are using QueryDSL **v.3.6.8**_

timowest:
Fixes #1846

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -87,6 +87,12 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+        <groupId>org.hamcrest</groupId>
+        <artifactId>hamcrest-library</artifactId>
+        <version>1.3</version>
+        <scope>test</scope>
+    </dependency>
 
     <dependency>
       <groupId>org.easymock</groupId>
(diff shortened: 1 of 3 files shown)


Shredder121:
So the build fails because?
Did you hand count the result of the query? Or is something else wrong?

timowest:
I think EclipseLink handles implict joins differently from Hibernate.

Shredder121:
I think this looks fine.
@johnktims What do you think?

johnktims:
I would like the tests to assert something about the order of the data instead of just looking at the size.

Shredder121:
Oh that sounds good yes.
Who's going to do it?

Shredder121:
I can do it, but I was wondering if we want Hamcrest for the tests.
It makes tests focused on ordering of elements clearer.

PR #1850, opened 2016-04-17
Fix orderby handling

### LINKED ISSUE #1846: ERROR: for SELECT DISTINCT, ORDER BY expressions must appear in select list

timowest:
Fixes #1846

### DIFF
--- a/querydsl-jpa/src/test/java/com/mysema/query/AbstractJPATest.java
+++ b/querydsl-jpa/src/test/java/com/mysema/query/AbstractJPATest.java
@@ -22,6 +22,7 @@ import java.util.*;
 import java.util.Calendar;
 import java.util.Map.Entry;
 
+import org.hamcrest.Matchers;
 import org.junit.Before;
 import org.junit.Ignore;
 import org.junit.Test;
@@ -31,6 +32,7 @@ import com.google.common.collect.Lists;
 import com.mysema.commons.lang.Pair;
 import com.mysema.query.group.Group;
 import com.mysema.query.group.GroupBy;
+import com.mysema.query.group.MockTuple;
 import com.mysema.query.group.QPair;
 import com.mysema.query.jpa.JPAExpressions;
 import com.mysema.query.jpa.JPASubQuery;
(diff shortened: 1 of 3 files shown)


querydsl-jpa/src/test/java/com/mysema/query/AbstractJPATest.java
```diff
@@ -547,6 +547,26 @@ public void DistinctResults() {
...
+                .distinct()
+                .from(cat)
+                .orderBy(cat.mate.id.asc())
+                .list(cat.id, cat.mate.id).size());
+    }
+
+    @Test
+    public void Distinct_OrderBy2() {
+        QCat cat = QCat.cat;
+        assertEquals(4, query()
```
johnktims: Why do you expect this one to return `4` and the previous (looks identical) query to return `6`?

johnktims: Also, wouldn't you want to assert something about the order or the uniqueness of the elements instead of looking at how many are returned?

timowest: Good point, I missed the annotations. I'll improve the assertions.

Shredder121:
What was wrong with this backport PR?

timowest:
Maybe the different definition order. The order by part is expanded as it get's defined in the query.
In Querydsl 4, the select part is then already known, in Querydsl 3 it isn't.

I'll try to fix it.

johnktims:
3.x hasn't been supported for several years.

PR #1851, opened 2016-04-17
Exposed query as DBObject

### LINKED ISSUE #1842: Expose the QueryMetadata on the AbstractMongodbQuery so it can be converted to a Morphia Query
I use QueryDSL not just for querying, but also for deleting and updating mongo entities.  Since QueryDSL only support queries, I have to convert the AbstractMongodbQuery into a Morphia query (or break it down into a DBObject and obtain query modifiers to talk directly using the java driver).

The only workaround I know to convert a QueryDSL to a Morphia Query / DBObject is a hack whereby the query is serialized to JSON, deserialized to a DBObject and then jacked into a Morphia query using reflection to call an otherwise private method:

See https://groups.google.com/forum/#!topic/querydsl/ItCdp5wR6Vs

This works in simple cases where there is no limit or sorting on the query, but we lose modifier information.

Would it be possible to expose a getMetadata() on AbstractMongodbQuery so that we can get the limit, order and projection modifiers?

Alternatively, a "toMorphiaQuery" method on the AbstractMongodbQuery would also be a good addition.

I wouldn't mind contributing with a patch - just let me know which solution seems more appropriate.

Gaspard

timowest:
Fixes #1842

### DIFF
--- a/querydsl-mongodb/src/main/java/com/querydsl/mongodb/AbstractMongodbQuery.java
+++ b/querydsl-mongodb/src/main/java/com/querydsl/mongodb/AbstractMongodbQuery.java
@@ -411,9 +411,18 @@ public abstract class AbstractMongodbQuery<K, Q extends AbstractMongodbQuery<K, 
         this.readPreference = readPreference;
     }
 
+    /**
+     * Get the where definition as a DBObject instance
+     *
+     * @return
+     */
+    public DBObject asDBObject() {
+        return createQuery(queryMixin.getMetadata().getWhere());
+    }
+
     @Override
     public String toString() {
-        return createQuery(queryMixin.getMetadata().getWhere()).toString();
+        return asDBObject().toString();
     }
 
 }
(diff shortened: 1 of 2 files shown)


querydsl-mongodb/src/main/java/com/querydsl/mongodb/AbstractMongodbQuery.java
```diff
@@ -411,6 +411,15 @@ public void setReadPreference(ReadPreference readPreference) {
...
+     *
+     * @return
+     */
+    public DBObject asDBObject() {
+        return createQuery(queryMixin.getMetadata().getWhere());
+    }
+
     @Override
     public String toString() {
         return createQuery(queryMixin.getMetadata().getWhere()).toString();
```
Shredder121: Would we be able to reuse the `asDBObject` here in the `toString()`?
I don't know if they will ever diverge, but it reduces repetition right now.

timowest: good point

johnktims:
Looks good to me.

Shredder121:
To me as well.
I'll merge it then.

PR #1855, opened 2016-04-21
Further improve normalization

timowest:


Shredder121:
I think this looks good.
Shall I try to break the regex?
It looks at least fine again to me.

timowest:
> Shall I try to break the regex?

Sure, give it a try

Shredder121:
I didn't force push, you can rebase if you want (those two commits could be combined).

Shredder121:
There is one more issue, but it's up to you whether we wan't to tackle it.
 ```java
assertUntouched("'INPS-ISET-0000-0000-0000.0000'");
```
These inner 'operations' are too far away from the quotes, so they are seen by the regex as candidates.
3286c3d826bd03e372bbe1de0770a65b0cb663a3

PR #1856, opened 2016-04-21
Add a profile that skips validation

Shredder121:
This might be useful for end users, since there could be lots of time wasted in the checks.

Let me know what you think.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,8 +33,9 @@ Free support is provided in the [Querydsl Google Group](https://groups.google.co
 Querydsl provides releases via public Maven repositories, but you can also build the sources yourself like this
 
 ```BASH
-$ mvn -DskipTests=true clean install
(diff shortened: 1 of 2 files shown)


timowest:
Looks good

Shredder121:
Do you think we can also use it with Travis, as initial install?

timowest:
> Do you think we can also use it with Travis, as initial install?

Makes sense.

Shredder121:
Do you want to quit the build early if there are issues?
Or do you want to have it in the verification process?

Shredder121:
Because if you want it to quit the build early, then it's better to leave Travis the way it is.

Shredder121:
I noticed that for Travis the time doesn't really matter, since it has to do a two pass anyway.
So I suppose that's it then? Let me know what you think.

johnktims:
This looks good to me.

PR #1859, opened 2016-04-25
querydsl.variableNameFunctionClass swallows ClassNotFoundExceptions

Shredder121:
I was either going to introduce a new try-catch around the `Class.forName` or adding a catch clause to the existing try-catch.
Which one do you prefer?

### DIFF
--- a/querydsl-apt/src/main/java/com/querydsl/apt/DefaultConfiguration.java
+++ b/querydsl-apt/src/main/java/com/querydsl/apt/DefaultConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015, The Querydsl Team (http://www.querydsl.com/team)
+ * Copyright 2016, The Querydsl Team (http://www.querydsl.com/team)
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -195,8 +195,10 @@ public class DefaultConfiguration implements Configuration {
                 variableNameFunction = variableNameFunctionClass.newInstance();
             } catch (RuntimeException e) {
                 throw e;
+            } catch (ClassNotFoundException e) {
+                throw new RuntimeException(e);
             } catch (Exception e) {
-               variableNameFunction = DefaultVariableNameFunction.INSTANCE;
+                variableNameFunction = DefaultVariableNameFunction.INSTANCE;
             }
         } else {
             variableNameFunction = DefaultVariableNameFunction.INSTANCE;
--- a/querydsl-codegen/src/main/java/com/querydsl/codegen/NamingFunction.java
+++ b/querydsl-codegen/src/main/java/com/querydsl/codegen/NamingFunction.java
@@ -15,24 +15,12 @@
  */
 package com.querydsl.codegen;
 
-import javax.lang.model.SourceVersion;
-
-import com.mysema.codegen.StringUtils;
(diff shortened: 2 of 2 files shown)


querydsl-apt/src/main/java/com/querydsl/apt/DefaultConfiguration.java
```diff
@@ -195,6 +196,8 @@ public DefaultConfiguration(
                 variableNameFunction = variableNameFunctionClass.newInstance();
             } catch (RuntimeException e) {
                 throw e;
+            } catch (ClassNotFoundException e) {
+                throw Throwables.propagate(e);
```
natros: Guava intends to deprecate this method, and recommends throwing a RuntimeException directly
https://github.com/google/guava/blob/2bcf189a62275bb23fd92edf4dc72e9ad849a89e/guava/src/com/google/common/base/Throwables.java#L201-L203

Shredder121: Thanks for the heads up!
It's still a WIP, but I appreciate the comment.

timowest:
The proposed solution is fine

Shredder121:
Shall we maybe for Querydsl 5 (or now) add a NamingFunction class, so people can just implement that?
So you can be sure you get the generics right?
Just an idea.

timowest:
NamingFunction sounds good, it could extend `Function<EntityType, String>`, that way we could introduce it already now.

querydsl-apt/src/main/java/com/querydsl/apt/DefaultConfiguration.java
```diff
@@ -195,6 +195,8 @@ public DefaultConfiguration(
                 variableNameFunction = variableNameFunctionClass.newInstance();
             } catch (RuntimeException e) {
                 throw e;
+            } catch (ClassNotFoundException e) {
+                throw new RuntimeException(e);
             } catch (Exception e) {
                variableNameFunction = DefaultVariableNameFunction.INSTANCE;
```
Shredder121: Should we also log this?
Or write it the syserr?

Shredder121: I mean the `Exception`.

querydsl-apt/src/main/java/com/querydsl/apt/NamingFunction.java
```diff
@@ -0,0 +1,27 @@
...
+
+import com.google.common.base.Function;
+import com.querydsl.codegen.EntityType;
+
+/**
+ * The {@link Function} signature to convert {@link EntityType} to {@link String}.
+ *
+ * @author Shredder121
+ */
+public interface NamingFunction extends Function<EntityType, String> {
```
timowest: should we change the existing naming functions to implement this interface?

Shredder121: That's up to you.
We'll probably have to add a new `catch` for `CCE` and a friendly notice that we moved to a better interface :wink:

timowest:
Looks good

Shredder121:
Actually, I only now see what you mean.
For that we would need to move the interface to Querydsl Codegen.
We can still do that, now that it's not yet merged.

timowest:
Great, this was what I had in mind

PR #1862, opened 2016-04-27
Improve aliasing

### LINKED ISSUE #1861: ORA-00918: column ambiguously defined when selecting two columns that differ only in case (on Oracle)
This is almost the same issue as #1445 - which was fixed.

The fix did not handle the case where there are two columns that have the same name except with different case (e.g. `Id` and `id`).

timowest:
Fixes #1861

### DIFF
--- a/querydsl-sql/src/main/java/com/querydsl/sql/SQLSerializer.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/SQLSerializer.java
@@ -176,7 +176,7 @@ public class SQLSerializer extends SerializerBase<SQLSerializer> {
             for (Expression<?> expr : expressions) {
                 if (expr instanceof Path) {
                     String name = ColumnMetadata.getName((Path<?>) expr);
-                    if (!names.add(name)) {
+                    if (!names.add(name.toLowerCase(Locale.ENGLISH))) {
                         expr = ExpressionUtils.as(expr, "col__" + name + replacements.size());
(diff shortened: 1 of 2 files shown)


querydsl-sql/src/main/java/com/querydsl/sql/SQLSerializer.java
```diff
@@ -176,7 +176,7 @@ public final SQLSerializer handleSelect(final String sep, final List<? extends E
             for (Expression<?> expr : expressions) {
                 if (expr instanceof Path) {
                     String name = ColumnMetadata.getName((Path<?>) expr);
-                    if (!names.add(name)) {
+                    if (!names.add(name.toLowerCase())) {
```
Shredder121: Locale dependent :(

Shredder121:
We still need a configurable Locale or something actually.

querydsl-sql/src/main/java/com/querydsl/sql/SQLSerializer.java
timowest: True, shall I use english until we have set up the Querydsl locale?

Shredder121: I think that's at least good.

Shredder121:
Well, at least it looks fine to me.
And the build succeeds. :wink:

PR #1871, opened 2016-05-04
Wildcard.all support for Tuple

### LINKED ISSUE #1870: WIldcard.all support for Tuple
https://groups.google.com/forum/#!topic/querydsl/iBoEe5SbnwI

This probably will need some changes to the way FactoryExpression works, since at creation time of `Tuple` you don't know how the bindings of `Wildcard.all` look like.

Shredder121:
Closes #1870

### DIFF
--- a/querydsl-sql/src/test/java/com/querydsl/sql/SelectBase.java
+++ b/querydsl-sql/src/test/java/com/querydsl/sql/SelectBase.java
@@ -1803,6 +1803,15 @@ public class SelectBase extends AbstractBaseTest {
     }
 
     @Test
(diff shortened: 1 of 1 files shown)


timowest:
Instead of providing support for mapping Wildcard.all into tuple we could directly support a `Map<String, Object>` mapping?

Projections.tuple(Wildcard.all)

timowest:
sorry, closed accidently

timowest:
Wildcard.all in QTuple should still binding all the columns of the result set into a single object array. The test should be changed to assert a single a tuple of size with a value of Object[] instead.

Issue #1869, opened 2016-05-04
PathBuilder is not threadsafe


robertandrewbain:
`private final Map<String, PathBuilder<?>> properties = Maps.newHashMap()` and `private final Map<Path<?>, Object> propertyMetadata = Maps.newHashMap()` should be `ConcurrentHashMap`s.

Furthermore, the check in `public <A> PathBuilder<A> get(String property, Class<A> type)`, `|| !type.isAssignableFrom(path.getType())` should be removed, as it does not make sense to overwrite existing values.

I believe all the superclass implementations to be thread safe.

Shredder121:
I _think_ we'll then at least also need a few test cases on what would need to happen if first you `.get("property")` and somewhere later you `.getString("property")` or something specific.
At least, we'll see.

PR #1872, opened 2016-05-04
Make PathBuilder thread safe

robertandrewbain:
#1869

### DIFF
--- a/querydsl-core/src/main/java/com/querydsl/core/types/dsl/PathBuilder.java
+++ b/querydsl-core/src/main/java/com/querydsl/core/types/dsl/PathBuilder.java
@@ -14,12 +14,13 @@
 package com.querydsl.core.types.dsl;
 
 import java.lang.reflect.Array;
+import java.util.AbstractMap.SimpleEntry;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 
-import com.google.common.collect.Maps;
 import com.querydsl.core.types.EntityPath;
 import com.querydsl.core.types.Path;
 import com.querydsl.core.types.PathMetadata;
@@ -44,9 +45,9 @@ public class PathBuilder<T> extends EntityPathBase<T> {
 
     private static final long serialVersionUID = -1666357914232685088L;
 
-    private final Map<String, PathBuilder<?>> properties = Maps.newHashMap();
+    private final ConcurrentHashMap<SimpleEntry<String, Class<?>>, PathBuilder<?>> properties = new ConcurrentHashMap<SimpleEntry<String, Class<?>>, PathBuilder<?>>();
 
-    private final Map<Path<?>, Object> propertyMetadata = Maps.newHashMap();
+    private final ConcurrentHashMap<Path<?>, Object> propertyMetadata = new ConcurrentHashMap<Path<?>, Object>();
 
     private final PathBuilderValidator validator;
 
@@ -96,15 +97,17 @@ public class PathBuilder<T> extends EntityPathBase<T> {
 
     private <P extends Path<?>> P  addMetadataOf(P newPath, Path<?> path) {
         if (path.getMetadata().getParent() instanceof EntityPath) {
-            EntityPath<?> parent = (EntityPath) path.getMetadata().getParent();
-            propertyMetadata.put(newPath, parent.getMetadata(path));
+            EntityPath<?> parent = (EntityPath<?>) path.getMetadata().getParent();
+            if (parent.getMetadata(path) != null) {
+                propertyMetadata.putIfAbsent(newPath, parent.getMetadata(path));
+            }
         }
         return newPath;
     }
 
     @SuppressWarnings("unchecked")
-    protected <T> Class<? extends T> validate(String property, Class<T> propertyType) {
-        Class<T> validatedType = (Class) validator.validate(getType(), property, propertyType);
+    protected <A> Class<? extends A> validate(String property, Class<A> propertyType) {
+        Class<A> validatedType = (Class<A>) validator.validate(getType(), property, propertyType);
         if (validatedType != null) {
             return validatedType;
         } else {
@@ -125,13 +128,15 @@ public class PathBuilder<T> extends EntityPathBase<T> {
      */
     @SuppressWarnings("unchecked")
     public PathBuilder<Object> get(String property) {
-        PathBuilder<Object> path = (PathBuilder) properties.get(property);
+        SimpleEntry<String, Class<?>> entry = new SimpleEntry<String, Class<?>>(property, Object.class);
+        PathBuilder<Object> path = (PathBuilder<Object>) properties.get(entry);
+        PathBuilder<?> existingPath = null;
         if (path == null) {
             Class<?> vtype = validate(property, Object.class);
             path = new PathBuilder<Object>(vtype, forProperty(property), validator);
-            properties.put(property, path);
+            existingPath = properties.putIfAbsent(entry, path);
         }
-        return path;
+        return existingPath == null ? path : (PathBuilder<Object>) existingPath;
     }
 
     /**
@@ -144,13 +149,15 @@ public class PathBuilder<T> extends EntityPathBase<T> {
      */
     @SuppressWarnings("unchecked")
     public <A> PathBuilder<A> get(String property, Class<A> type) {
-        PathBuilder<A> path = (PathBuilder<A>) properties.get(property);
-        if (path == null || !type.isAssignableFrom(path.getType())) {
+        SimpleEntry<String, Class<?>> entry = new SimpleEntry<String, Class<?>>(property, type);
+        PathBuilder<A> path = (PathBuilder<A>) properties.get(entry);
+        PathBuilder<?> existingPath = null;
+        if (path == null) {
             Class<? extends A> vtype = validate(property, type);
             path = new PathBuilder<A>(vtype, forProperty(property), validator);
-            properties.put(property, path);
+            existingPath = properties.putIfAbsent(entry, path);
         }
-        return path;
+        return existingPath == null ? path : (PathBuilder<A>) existingPath;
     }
 
     /**
@@ -240,7 +247,7 @@ public class PathBuilder<T> extends EntityPathBase<T> {
     @SuppressWarnings("unchecked")
     public <A extends Comparable<?>> ComparablePath<A> getComparable(String property, Class<A> type) {
         Class<? extends A> vtype = validate(property, type);
-        return super.createComparable(property, (Class) vtype);
+        return super.createComparable(property, (Class<? super A>) vtype);
     }
 
     /**
@@ -267,7 +274,7 @@ public class PathBuilder<T> extends EntityPathBase<T> {
     @SuppressWarnings("unchecked")
     public <A extends Comparable<?>> DatePath<A> getDate(String property, Class<A> type) {
         Class<? extends A> vtype = validate(property, type);
-        return super.createDate(property, (Class) vtype);
+        return super.createDate(property, (Class<? super A>) vtype);
     }
 
     /**
@@ -294,7 +301,7 @@ public class PathBuilder<T> extends EntityPathBase<T> {
     @SuppressWarnings("unchecked")
     public <A extends Comparable<?>> DateTimePath<A> getDateTime(String property, Class<A> type) {
         Class<? extends A> vtype = validate(property, type);
-        return super.createDateTime(property, (Class) vtype);
+        return super.createDateTime(property, (Class<? super A>) vtype);
     }
 
     /**
@@ -405,7 +412,7 @@ public class PathBuilder<T> extends EntityPathBase<T> {
     @SuppressWarnings("unchecked")
     public <A extends Number & Comparable<?>> NumberPath<A> getNumber(String property, Class<A> type) {
         Class<? extends A> vtype = validate(property, type);
-        return super.createNumber(property, (Class) vtype);
+        return super.createNumber(property, (Class<? super A>) vtype);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


querydsl-core/src/main/java/com/querydsl/core/types/dsl/PathBuilder.java
```diff
@@ -44,9 +44,9 @@
 
     private static final long serialVersionUID = -1666357914232685088L;
 
-    private final Map<String, PathBuilder<?>> properties = Maps.newHashMap();
+    private final ConcurrentHashMap<String, PathBuilder<?>> properties = new ConcurrentHashMap<String, PathBuilder<?>>();
```
Shredder121: You could also use `Maps.newConcurrentHashMap()`.

Shredder121: And maybe keep the declaration as `Map`?

robertandrewbain: `Maps.newConcurrentHashMap()` returns a different implementation, `ConcurrentHashMap` is what's used in the superclasses.

The declaration change is deliberate. I'm forcing the type, as we *always* want a `ConcurrentHashMap`.

querydsl-core/src/test/java/com/querydsl/core/types/dsl/PathBuilderFactoryTest.java
```diff
@@ -14,21 +14,23 @@
...
-
 import org.junit.Test;
 
+import static org.junit.Assert.assertEquals;
+
 public class PathBuilderFactoryTest {
 
     @Test
     public void create() {
         PathBuilderFactory factory = new PathBuilderFactory();
```
robertandrewbain: `PathBuilderFactory` seems like a pretty pointless class. Thoughts?

querydsl-core/src/test/java/com/querydsl/core/types/dsl/PathBuilderTest.java
```diff
@@ -112,4 +117,27 @@ public void various() {
...
+
+    @Test
+    public void attempting_to_overwrite_an_existing_path_with_one_of_a_different_type_throws_a_RuntimeException() {
+        expectedException.expect(RuntimeException.class);
+        expectedException.expectMessage(
+                "Property name [some_path] of type [class java.lang.Object] cannot be overwritten with type [class java.lang.Integer] due to thread safety");
+        PathBuilder<User> entity = new PathBuilder<User>(User.class, "entity");
+        String pathName = "some_path";
+        entity.get(pathName);
+        entity.get(pathName, Integer.class);
```
Shredder121: I'd really like that it doesn't throw actually.
We have generic filtering that uses `.get(path)` and sorting that uses `.getComparable(path)`.
So it's perfectly valid for it to remake path objects with a more specific type :(

Shredder121: Let me sleep on it.
I think it could work, but there might be cases where this might throw too much in comparison to before.

robertandrewbain: I would have hoped if that were the case, we'd see some breaking tests.

robertandrewbain: If we do need to support path objects with the same name but different types, perhaps we could change the map declaration to `Map<SimpleEntry<String,Class<?>>, PathBuilder<?>>` and pull out the correct types?

querydsl-core/src/main/java/com/querydsl/core/types/dsl/PathBuilder.java
```diff
@@ -96,15 +97,17 @@ public PathBuilder(Class<? extends T> type, String variable) {
...
+                propertyMetadata.putIfAbsent(newPath, parent.getMetadata(path));
+            }
         }
         return newPath;
     }
 
     @SuppressWarnings("unchecked")
-    protected <T> Class<? extends T> validate(String property, Class<T> propertyType) {
-        Class<T> validatedType = (Class) validator.validate(getType(), property, propertyType);
+    protected <A> Class<? extends A> validate(String property, Class<A> propertyType) {
```
robertandrewbain: Removed compiler warning about hiding class level generic T

querydsl-core/src/main/java/com/querydsl/core/types/dsl/PathBuilder.java
```diff
@@ -240,7 +245,7 @@ public BooleanPath getBoolean(String propertyName) {
     @SuppressWarnings("unchecked")
     public <A extends Comparable<?>> ComparablePath<A> getComparable(String property, Class<A> type) {
         Class<? extends A> vtype = validate(property, type);
-        return super.createComparable(property, (Class) vtype);
+        return super.createComparable(property, (Class<? super A>) vtype);
```
robertandrewbain: Fixed raw type

querydsl-core/src/main/java/com/querydsl/core/types/dsl/PathBuilder.java
```diff
@@ -267,7 +272,7 @@ public BooleanPath getBoolean(String propertyName) {
     @SuppressWarnings("unchecked")
     public <A extends Comparable<?>> DatePath<A> getDate(String property, Class<A> type) {
         Class<? extends A> vtype = validate(property, type);
-        return super.createDate(property, (Class) vtype);
+        return super.createDate(property, (Class<? super A>) vtype);
```
robertandrewbain: Fixed raw type

querydsl-core/src/main/java/com/querydsl/core/types/dsl/PathBuilder.java
```diff
@@ -294,7 +299,7 @@ public BooleanPath getBoolean(String propertyName) {
     @SuppressWarnings("unchecked")
     public <A extends Comparable<?>> DateTimePath<A> getDateTime(String property, Class<A> type) {
         Class<? extends A> vtype = validate(property, type);
-        return super.createDateTime(property, (Class) vtype);
+        return super.createDateTime(property, (Class<? super A>) vtype);
```
robertandrewbain: Fixed raw type

querydsl-core/src/main/java/com/querydsl/core/types/dsl/PathBuilder.java
```diff
@@ -405,7 +410,7 @@ public BooleanPath getBoolean(String propertyName) {
     @SuppressWarnings("unchecked")
     public <A extends Number & Comparable<?>> NumberPath<A> getNumber(String property, Class<A> type) {
         Class<? extends A> vtype = validate(property, type);
-        return super.createNumber(property, (Class) vtype);
+        return super.createNumber(property, (Class<? super A>) vtype);
```
robertandrewbain: Fixed raw type

querydsl-core/src/main/java/com/querydsl/core/types/dsl/PathBuilder.java
```diff
@@ -459,7 +464,7 @@ public BooleanPath getBoolean(String propertyName) {
     @SuppressWarnings("unchecked")
     public <A> SimplePath<A> getSimple(String property, Class<A> type) {
         Class<? extends A> vtype = validate(property, type);
-        return super.createSimple(property, (Class) vtype);
+        return super.createSimple(property, (Class<? super A>) vtype);
```
robertandrewbain: Fixed raw type

querydsl-core/src/main/java/com/querydsl/core/types/dsl/PathBuilder.java
```diff
@@ -508,7 +513,7 @@ public StringPath getString(String property) {
     @SuppressWarnings("unchecked")
     public <A extends Comparable<?>> TimePath<A> getTime(String property, Class<A> type) {
         Class<? extends A> vtype = validate(property, type);
-        return super.createTime(property, (Class) vtype);
+        return super.createTime(property, (Class<? super A>) vtype);
```
robertandrewbain: Fixed raw type

querydsl-core/src/main/java/com/querydsl/core/types/dsl/PathBuilder.java
```diff
@@ -44,9 +45,9 @@
 
     private static final long serialVersionUID = -1666357914232685088L;
 
-    private final Map<String, PathBuilder<?>> properties = Maps.newHashMap();
+    private final ConcurrentHashMap<SimpleEntry<String, Class<?>>, PathBuilder<?>> properties = new ConcurrentHashMap<SimpleEntry<String, Class<?>>, PathBuilder<?>>();
```
robertandrewbain: properties will now support multiples types for the same name

querydsl-core/src/main/java/com/querydsl/core/types/dsl/PathBuilder.java
```diff
@@ -125,11 +128,12 @@ public Object getMetadata(Path<?> property) {
...
     @SuppressWarnings("unchecked")
     public PathBuilder<Object> get(String property) {
-        PathBuilder<Object> path = (PathBuilder) properties.get(property);
+        SimpleEntry<String, Class<?>> entry = new SimpleEntry<String, Class<?>>(property, Object.class);
+        PathBuilder<Object> path = (PathBuilder<Object>) properties.get(entry);
         if (path == null) {
             Class<?> vtype = validate(property, Object.class);
             path = new PathBuilder<Object>(vtype, forProperty(property), validator);
-            properties.put(property, path);
+            properties.putIfAbsent(entry, path);
```
Shredder121: image

querydsl-core/src/main/java/com/querydsl/core/types/dsl/PathBuilder.java
```diff
@@ -144,11 +148,12 @@ public Object getMetadata(Path<?> property) {
...
     public <A> PathBuilder<A> get(String property, Class<A> type) {
-        PathBuilder<A> path = (PathBuilder<A>) properties.get(property);
-        if (path == null || !type.isAssignableFrom(path.getType())) {
+        SimpleEntry<String, Class<?>> entry = new SimpleEntry<String, Class<?>>(property, type);
+        PathBuilder<A> path = (PathBuilder<A>) properties.get(entry);
+        if (path == null) {
             Class<? extends A> vtype = validate(property, type);
             path = new PathBuilder<A>(vtype, forProperty(property), validator);
-            properties.put(property, path);
+            properties.putIfAbsent(entry, path);
```
Shredder121: Same for this one.

querydsl-core/src/main/java/com/querydsl/core/types/dsl/PathBuilder.java
timowest: putIfAbsent is Java 8, we should be compatible with Java 6/7 as well.

querydsl-core/src/main/java/com/querydsl/core/types/dsl/PathBuilder.java
timowest: maybe still Map as the field type, no strong opinion on this though

robertandrewbain: Deliberately forcing the type as we always want it to be a `ConcurrentHashMap`

querydsl-core/src/main/java/com/querydsl/core/types/dsl/PathBuilder.java
robertandrewbain: http://docs.oracle.com/javase/6/docs/api/java/util/concurrent/ConcurrentHashMap.html what am I missing?

Shredder121: In [`ConcurrentMap` there is `putIfAbsent`](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ConcurrentMap.html#putIfAbsent(K,%20V)).
Otherwise animal-sniffer would have complained.

querydsl-core/src/main/java/com/querydsl/core/types/dsl/PathBuilder.java
Shredder121: The highest we can go is `ConcurrentMap`, otherwise it will compile against `Map.putIfAbsent`, which is only Java 8.
But @robertandrewbain said that elsewhere we also have the `ConcurrentHashMap`.

querydsl-core/src/main/java/com/querydsl/core/types/dsl/PathBuilder.java
timowest: my bad, looked at the interface level

robertandrewbain: I guess this means the declaration stays, too.

Shredder121:
Well at least the builds now pass. :sweat_smile: 
My comment is indeed addressed now.
The only things I can think of is declaring `ConcurrentMap` instead. And using `Maps.newConcurrentMap()` (it doesn't use a different implementation, it only does so when you want special keys (soft, or weak)).
But other than that, I think that all points are addressed.

querydsl-core/src/main/java/com/querydsl/core/types/dsl/PathBuilder.java
```diff
@@ -125,13 +128,15 @@ public Object getMetadata(Path<?> property) {
...
     public PathBuilder<Object> get(String property) {
-        PathBuilder<Object> path = (PathBuilder) properties.get(property);
+        SimpleEntry<String, Class<?>> entry = new SimpleEntry<String, Class<?>>(property, Object.class);
+        PathBuilder<Object> path = (PathBuilder<Object>) properties.get(entry);
+        PathBuilder<?> existingPath = null;
         if (path == null) {
             Class<?> vtype = validate(property, Object.class);
             path = new PathBuilder<Object>(vtype, forProperty(property), validator);
-            properties.put(property, path);
+            existingPath = properties.putIfAbsent(entry, path);
```
Shredder121: Question: should we recreate the `entry` before `put()`ing, in case a validator returns a more specific `vtype`?

timowest:
Looks great. PathBuilder should indeed be threadsafe.

PR #1878, opened 2016-05-09
Ensure all date operations actually resolve to integer

### LINKED ISSUE #1876: direct cast in SQL query
Hello, I had a problem with typing "where part" query and I noticed that there is no way to cast some part of SQL query. DB - postgresql

Example:  - user anniversary >
`.where(USER.validFrom.year().subtract(LocalDate.now().getYear()).mod(5).eq(0))`
it will generate 
`MOD( EXTRACT( year from timestamp '1.1.2011' ) - 2016, 5) = 0`
but function `EXTRACT `return `DOUBLE `(every time) which I cant cast to Integer (in SQL) and it throw SQL syntax error (`function mod(double, int) doesnt exists`)

right it would be something like that >
`.where(USER.validFrom.year().sqlCast(Integer.class).subtract(LocalDate.now()).mod(5).eq(0))`
which could generate
```
MOD( EXTRACT( year from timestamp '1.1.2011' )::int - 2016, 5) = 0     OR
MOD( CAST( EXTRACT( year from timestamp '1.1.2011' ) AS integer) - 2016, 5) = 0
```

this would be great :)

Shredder121:
Databases like [Postgresql](http://www.postgresql.org/docs/current/static/functions-datetime.html#FUNCTIONS-DATETIME-EXTRACT) return double precision numbers with `extract(...)`.
We need to update the templates to include the cast.

Oh, fixes #1876

### DIFF
--- a/querydsl-sql/src/test/java/com/querydsl/sql/SelectBase.java
+++ b/querydsl-sql/src/test/java/com/querydsl/sql/SelectBase.java
@@ -1265,6 +1265,15 @@ public class SelectBase extends AbstractBaseTest {
     }
 
     @Test
+    public void num_date_operation() {
+        long result = query()
+                .select(employee.datefield.year().mod(1))
+                .from(employee)
+                .fetchFirst();
+        assertEquals(0, result);
+    }
+
+    @Test
     @ExcludeIn({DERBY, FIREBIRD, POSTGRESQL})
     public void number_as_boolean() {
         QNumberTest numberTest = QNumberTest.numberTest;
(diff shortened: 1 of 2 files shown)


querydsl-sql/src/test/java/com/querydsl/sql/SelectBase.java
```diff
@@ -1265,6 +1265,15 @@ public void num_cast2() {
     }
 
     @Test
+    public void num_cast_dateTime() { //TODO: think of good name
```
robertandrewbain: Naming seems consistent with the rest of the test class. I'd accept this change with the TODO removed.

Shredder121: But it doesn't actually use cast anymore (that was the workaround).
It's now actually more of a `numeric_date_operation`.

Shredder121:
Nice, green.

timowest:
Looks good

robertandrewbain:
Looks good

PR #1886, opened 2016-05-15
Querydsl 4.1.1

timowest:


querydsl-codegen/src/main/java/com/querydsl/codegen/DefaultVariableNameFunction.java
```diff
@@ -17,13 +17,14 @@
...
 
+import com.google.common.base.Function;
 import com.mysema.codegen.StringUtils;
 
 /**
  * Default variable name generation strategy which un-capitalizes the first letter of the class name.
  *
  */
-public final class DefaultVariableNameFunction implements NamingFunction {
+public final class DefaultVariableNameFunction implements Function<EntityType, String> {
```
Shredder121: You're going with the original instead?

PR #1887, opened 2016-05-15
Querydsl 3.7.3

timowest:


querydsl-codegen/src/main/java/com/mysema/query/codegen/DefaultVariableNameFunction.java
```diff
@@ -15,14 +15,15 @@
...
+import com.google.common.base.Function;
 import com.mysema.codegen.StringUtils;
 import com.mysema.util.JavaSyntaxUtils;
 
 /**
  * Default variable name generation strategy which un-capitalizes the first letter of the class name.
  *
  */
-public final class DefaultVariableNameFunction implements NamingFunction {
+public final class DefaultVariableNameFunction implements Function<EntityType, String> {
```
Shredder121: You're going with the original instead?

timowest: the backwards compatibility checker didn't like it, although I think it's compatible

Shredder121: Oh I see, yes I'd like to think it's compatible as well.
But this is also fine.

PR #1893, opened 2016-05-20
Add optional suffix to PathBuilderFactory

### LINKED ISSUE #1891: Using alternative entity alias when it is looks like a part of HQL keywords
For example QueryDSL generates HQL query for delete entity Group with Q-Class QGroup as:
```
delete from Group group
where group = ?1
```
And Hibernate throws an exception:

```
org.hibernate.hql.internal.ast.QuerySyntaxException: unexpected token: group near line 1, column 55 [delete from com.lukoil.wellinfo.dao.model.admin.Group group
where group = ?1]
```

Looks like Hibernate decide that "group" is part of  keyword "group by" instead alias.
May be QueryDSL should be generate another alias name for entity if it looks like a part of HQL keywords?

timowest:
Fixes #1891

### DIFF
--- a/querydsl-core/src/main/java/com/querydsl/core/types/dsl/PathBuilderFactory.java
+++ b/querydsl-core/src/main/java/com/querydsl/core/types/dsl/PathBuilderFactory.java
@@ -13,10 +13,10 @@
  */
 package com.querydsl.core.types.dsl;
 
-import java.util.HashMap;
 import java.util.Map;
 
 import com.google.common.base.CaseFormat;
+import com.google.common.collect.Maps;
 
 /**
  * {@code PathBuilderFactory} is a factory class for PathBuilder creation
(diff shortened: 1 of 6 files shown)


Shredder121:
Hm, seems to work.
Is there anything else you want to add?

timowest:
I'll comment in the issue.

querydsl-core/src/main/java/com/querydsl/core/types/dsl/PathBuilderFactory.java
```diff
@@ -38,10 +48,14 @@
...
         PathBuilder<T> rv = (PathBuilder<T>) paths.get(type);
         if (rv == null) {
-            rv = new PathBuilder<T>(type, CaseFormat.UPPER_CAMEL.to(CaseFormat.LOWER_CAMEL, type.getSimpleName()));
+            rv = new PathBuilder<T>(type, variableName(type));
             paths.put(type, rv);
         }
         return rv;
     }
 
+    private <T> String variableName(Class<T> type) {
```
Shredder121: I don't think this type parameter is necessary.

timowest: good point

Shredder121:
Looks fine to me.

PR #1899, opened 2016-05-26
Add support for bulk insert

### LINKED ISSUE #1888: Add support for bulk insert statements
At the moment the only way to insert multiple rows with one insert clause is to use the batch insert. But this is a very slow operation compared to a bulk insert especially if you have to insert thousands of rows at once.
So it would be nice if there would be a way to generate insert statements like this using the querydsl:
```
INSERT INTO table (column1, column2) VALUES (?, ?), (?, ?) ...
```

timowest:
Fixes #1888

### DIFF
--- a/querydsl-sql/src/main/java/com/querydsl/sql/dml/SQLInsertClause.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/dml/SQLInsertClause.java
@@ -150,7 +150,7 @@ public class SQLInsertClause extends AbstractSQLClause<SQLInsertClause> implemen
      * Will revert to batches, if bulk is not supported
      */
     public void setBatchToBulk(boolean b) {
-        this.batchToBulk = b;
+        this.batchToBulk = b && configuration.getTemplates().isBatchToBulkSupported();
     }
 
     @Override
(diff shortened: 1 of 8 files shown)


Shredder121:
What's the memory usage like with ~100 or more batches?
Is it more or less the same as with JDBC batching?
Just wondering.

timowest:
> What's the memory usage like with ~100 or more batches?
That depends on the values being bound and the JDBC driver implementation.

The batch handling in Querydsl is currently more sophisticated and safe, but since this is an explicit override internal optimization shouldn't be expected.

Shredder121:
Could you also add a verification test for `useLiterals`?

timowest:
What do you mean by verification test?

Shredder121:
A test case with `useLiterals` enabled, just so that also has coverage.

querydsl-sql/src/main/java/com/querydsl/sql/dml/SQLInsertClause.java
```diff
@@ -150,7 +150,7 @@ public SQLInsertClause addBatch() {
      * Will revert to batches, if bulk is not supported
      */
     public void setBatchToBulk(boolean b) {
-        this.batchToBulk = b;
+        this.batchToBulk = b && configuration.getTemplates().isBatchToBulkSupported();
```
Shredder121: Hmm, would it make sense to silently revert to batches?

Issue #1898, opened 2016-05-25
Coalesce using UNION keyword instead of a comma


robertandrewbain:
The following code `new Coalesce<>(some SQLQuery<BigDecimal>, Expressions.constant(BigDecimal.ZERO)).as(SOME_COLUMN_ALIAS)` will generate the following SQL `coalesce(<some query> union 0)` when it should generate `coalesce(<some query, 0)`. This bug has been introduced between 4.0.4 and 4.1.1.

**Update:** tested in 4.0.9 and bug not present. Tested in 4.1.0 and bug is present. Bug introduced in 4.1.0.

**Update:** I think this is caused by #1805 changes to `querydsl-sql/src/main/java/com/querydsl/sql/SQLSerializer.java`. https://github.com/querydsl/querydsl/pull/1805/files#diff-abdf9237fdabf083adb006f87913f4deR893

Shredder121:
Fixed in #1900.

PR #1900, opened 2016-05-26
Issue 1898: Coalesce using UNION keyword instead of a comma

robertandrewbain:
#1898

### DIFF
--- a/querydsl-sql/src/main/java/com/querydsl/sql/SQLSerializer.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/SQLSerializer.java
@@ -912,7 +912,13 @@ public class SQLSerializer extends SerializerBase<SQLSerializer> {
             }
         }
 
-        if (operator == SQLOps.UNION || operator == SQLOps.UNION_ALL) {
+        if (operator == Ops.SET && args.get(0) instanceof SubQueryExpression) {
+            boolean oldUnion = inUnion;
+            inUnion = true;
+            super.visitOperation(type, SQLOps.UNION, args);
+            inUnion = oldUnion;
+
+        } else if (operator == SQLOps.UNION || operator == SQLOps.UNION_ALL) {
             boolean oldUnion = inUnion;
             inUnion = true;
             super.visitOperation(type, operator, args);
--- a/querydsl-sql/src/main/java/com/querydsl/sql/SQLSerializer.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/SQLSerializer.java
@@ -912,13 +912,7 @@ public class SQLSerializer extends SerializerBase<SQLSerializer> {
             }
         }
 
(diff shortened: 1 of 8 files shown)


querydsl-sql/src/test/java/com/querydsl/sql/CoalesceTest.java
```diff
@@ -0,0 +1,21 @@
...
+
+import org.junit.Test;
+
+import com.querydsl.core.types.dsl.Coalesce;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.is;
+
+public class CoalesceTest {
```
robertandrewbain: I couldn't think where to put this... what do we think about the location of this test? Would the one individual test be better included in an existing test class somewhere?

querydsl-sql/src/main/java/com/querydsl/sql/SQLSerializer.java
```diff
@@ -895,13 +895,7 @@ protected void visitOperation(Class<?> type, Operator operator, List<? extends E
...
         }
 
-        if (operator == Ops.LIST && args.get(0) instanceof SubQueryExpression) {
-            boolean oldUnion = inUnion;
-            inUnion = true;
-            super.visitOperation(type, SQLOps.UNION, args);
-            inUnion = oldUnion;
-
-        } else if (operator == SQLOps.UNION || operator == SQLOps.UNION_ALL) {
+        if (operator == SQLOps.UNION || operator == SQLOps.UNION_ALL) {
```
robertandrewbain: The code I've removed looks wrong to me, and breaks coalesce for subqueries. Was it doing anything useful?

timowest: The problem is that the LIST operator is too general. It was used to turn

ex in (subquery1, subquery2, ...)

into

ex in (subquery1 union subquery2 union ...)

One way to fix it would be to introduce a SET operator and use that in ex in args expressions.

querydsl-sql/src/test/java/com/querydsl/sql/CoalesceTest.java
timowest: to me this is fine

robertandrewbain:
Looks good to me.

querydsl-sql/src/test/java/com/querydsl/sql/CoalesceTest.java
```diff
@@ -0,0 +1,21 @@
+package com.querydsl.sql;
+
+import org.junit.Test;
+
+import com.querydsl.core.types.dsl.Coalesce;
+
+import static org.hamcrest.MatcherAssert.assertThat;
```
Shredder121: Shouldn't this be`org.junit.Assert.assertThat`?

Shredder121:
Looks good.
I'll merge it then.

PR #1906, opened 2016-06-01
Improve release script

timowest:


release.sh
```diff
@@ -18,18 +19,18 @@ doit() {
   git pull
   mvn clean deploy -DskipTests -Dgpg.skip=false
   ./dist.sh
-  # TODO upload
+  ssh root@querydsl.com "mkdir /var/www/html/static/querydsl/$VERSION"
+  scp -r target/dist/* root@querydsl.com:/var/www/html/static/querydsl/$VERSION/
```
Shredder121: Could you add this command as well?
`cd /var/www/html/static/querydsl && unlink latest && ln -sT $VERSION latest`.

timowest: done

release.sh
```diff
@@ -18,18 +19,18 @@ doit() {
   git pull
```
Shredder121: You can also make this `git pull --ff-only` just in case, but not really necessary.

Shredder121:
Creating a pull request is not quite hard, we can do that as well.
https://developer.github.com/v3/pulls/#create-a-pull-request
The only thing that differs is the target (if the version is 4, master, else QUERYDSL_3).

timowest:
Let's do the other github integrations in another PR.

PR #1910, opened 2016-06-05
Make method handling compatible with APT logic

timowest:


Shredder121:
Could you also add a few test cases to `GenericExperterTest`, configured with `propertyHandling` directly?
Also, there is no propertyHandling equivalent for `export_handle_no_methods_nor_fields()` , is that right?

Shredder121:
If we also deprecate the fields, we'll notice faster that we have to remove it.
Or do you think it's fine like this?

Other than that, looks good to me.

PR #1918, opened 2016-06-10
Replace concrete ImmutableList with List interface in public api

### LINKED ISSUE #1908: Confusing Expressions/Template API
I find that the method overloading in a series of template factory methods (StringTemplate, BooleanTemplate, ...) defies the principle of least surprise. Example from [Expressions](https://github.com/querydsl/querydsl/blob/6bb7a28/querydsl-core/src/main/java/com/querydsl/core/types/dsl/Expressions.java#L622):

```java
public static StringTemplate stringTemplate(String template, Object... args) {
    return stringTemplate(createTemplate(template), ImmutableList.copyOf(args));
}

public static StringTemplate stringTemplate(String template, ImmutableList<?> args) {
    return stringTemplate(createTemplate(template), args);
}
```

Having method overloads for `Object... args` and `List<?> args` and even `Object firstArgs, Object... otherArgs` are common practices. However using a concrete implementation of a collection (`ImmutableList` here) is not. 
I was refactoring some code that interacts directly with queryDsl and wanted to change from using arrays of args to using collections. As I read the queryDsl API my eyes totally ignored the `Immutable` prefix because they are just so damn trained to expect an interface from the collections API. My mutable list ended up being passed to the `Object...` overload and stuff blew up. Quite frustrating :)

I'd suggest deprecating the `ImmutableList` overload and adding method overloads that take a `List` from the collections API. 

Let me know if you're OK with this and I'll gladly prepare a PR.

ruisalgado:
Avoids mistakes with other list implementations being dispatched
to a different method overload (intended for  #varargs).
Resolves #1908

### DIFF
--- a/querydsl-core/src/main/java/com/querydsl/core/types/ExpressionUtils.java
+++ b/querydsl-core/src/main/java/com/querydsl/core/types/ExpressionUtils.java
@@ -144,11 +144,25 @@ public final class ExpressionUtils {
     /**
      * Create a new Template expression
      *
+     * @deprecated Use {@link #predicateTemplate(String, List)} instead.
+     *
      * @param template template
      * @param args template parameters
      * @return template expression
      */
+    @Deprecated
     public static PredicateTemplate predicateTemplate(String template, ImmutableList<?> args) {
+        return predicateTemplate(TemplateFactory.DEFAULT.create(template), args);
+    }
+
+    /**
+     * Create a new Template expression
+     *
+     * @param template template
+     * @param args template parameters
+     * @return template expression
+     */
+    public static PredicateTemplate predicateTemplate(String template, List<?> args) {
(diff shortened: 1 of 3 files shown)


timowest:
This change makes sense in public factory methods, but not so much in protected methods. Also clean up the imports and make sure that the change is backwards compatible. In public methods the old variant still needs to be available, it can be deprecated though.

ruisalgado:
tks for the feedback. I'll make the changes.

I changed the protected methods because I didn't see an obvious reason to me why you'd want to commit to a concrete ImmutableList vs a more flexible interface. But I guess it makes sense if you want to enforce immutability as much as possible

timowest:
Looks good

Shredder121:
Yes, looks good to me as well.
I'll merge it then.

Shredder121:
@ruisalgado Thanks again for the PR!

PR #1925, opened 2016-06-16
Fix docs

### LINKED ISSUE #1923: Wrong code in result handling/constructor usage example
Chapter 3.2.3 Constructor usage uses [code](https://github.com/querydsl/querydsl/blob/master/querydsl-docs/src/main/docbook/en-US/content/general/result-handling.xml#L74) from the previous example, 3.2.2 Bean population

> 3.2.3. Constructor usage
> 
> Constructor based row transformation can be used like this
> 
> List<UserDTO> dtos = query.select(
>     Projections.bean(UserDTO.class, user.firstName, user.lastName)).fetch();

Should use `Projections.constructor(…)`

timowest:
Fixes #1923


Shredder121:
Looks good to me already :+1:

PR #1930, opened 2016-06-21
Fix template handling

### LINKED ISSUE #1929: PostgreSQL exception on SUBSTR_2ARGS
Using 4.1.2:
I'm passing a string path and two integer constants to substring, and get the following error on PostgreSQL 9.3. It seems the first integer parameter is converted to a numeric by the addition, but the second is not.

```
ERROR:  function substr(text, numeric, integer) does not exist at character 93
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
STATEMENT:  select "e1"."ID", "e1"."DESCRIPTION", "e1"."PROPERTIES"
	from "THINGS" "e1"
	where substr("e1"."DESCRIPTION",($1 + $2),($3 - $4)) = $5
```

I'm guessing it's because of the template for substring: `"substr({0},{1+'1's},{2-1s})"` 
Explicitly casting my integers to integers won't work of course, because the addition and subtraction happen after my manual casts...
A bit of debugging showed that the first `'1's` is converted to a BigDecimal, while the second `1s` is simply an Integer.

I've tried playing around with the template but I've not manged to compile querydsl myself due to https://github.com/querydsl/querydsl/issues/1788

timowest:
Fixes #1929

### DIFF
--- a/querydsl-core/src/main/java/com/querydsl/core/types/dsl/StringExpression.java
+++ b/querydsl-core/src/main/java/com/querydsl/core/types/dsl/StringExpression.java
@@ -483,6 +483,19 @@ public abstract class StringExpression extends LiteralExpression<String> {
     }
 
     /**
+     * Create a {@code locate(str, this, start)} expression
+     *
+     * <p>Get the position of the given String in this String, the first position is 1</p>
+     *
+     * @param str string
+     * @param start start
+     * @return locate(str, this, start)
+     */
+    public NumberExpression<Integer> locate(String str, Expression<Integer> start) {
+        return Expressions.numberOperation(Integer.class, Ops.StringOps.LOCATE2, ConstantImpl.create(str), mixin, start);
+    }
+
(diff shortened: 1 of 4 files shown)


querydsl-core/src/main/java/com/querydsl/core/types/dsl/StringExpression.java
```diff
@@ -483,6 +483,19 @@ public BooleanExpression likeIgnoreCase(Expression<String> str, char escape) {
     }
 
     /**
+     * Create a {@code locate(str, this, start)} expression
```
Shredder121: Either the signature, or this comment is wrong.

Shredder121: Oh nevermind, it's function notation

Shredder121:
Looks good.

PR #1935, opened 2016-06-23
Use default locale instead of ENGLISH

### LINKED ISSUE #1933: StringExpression.containsIgnoreCase utf-8 char problem
Hi i have this issue and can't find solution on the web.

when i send turkish char 'İ' as a param to containsIgnoreCase function doesn't find 'i' in database. when i look at logs i see that it lowers query parameter as  'i ̇ '

timowest:
Fixes #1933

### DIFF
--- a/querydsl-core/src/main/java/com/querydsl/core/types/TemplateFactory.java
+++ b/querydsl-core/src/main/java/com/querydsl/core/types/TemplateFactory.java
@@ -60,7 +60,7 @@ public class TemplateFactory {
                     } else if (arg instanceof Expression) {
                         return ExpressionUtils.operation(String.class, Ops.LOWER, (Expression) arg);
                     } else {
-                        return String.valueOf(arg).toLowerCase(Locale.ENGLISH);
+                        return String.valueOf(arg).toLowerCase();
                     }
                 }
             };
@@ -74,7 +74,7 @@ public class TemplateFactory {
                     } else if (arg instanceof Expression) {
                         return ExpressionUtils.operation(String.class, Ops.UPPER, (Expression) arg);
                     } else {
-                        return String.valueOf(arg).toUpperCase(Locale.ENGLISH);
+                        return String.valueOf(arg).toUpperCase();
                     }
                 }
             };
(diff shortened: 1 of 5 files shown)


querydsl-core/src/main/java/com/querydsl/core/types/TemplateFactory.java
```diff
@@ -183,9 +182,9 @@ public Template create(String template) {
                 if (m.start() > end) {
                     elements.add(new Template.StaticText(template.substring(end, m.start())));
                 }
-                String premodifiers = m.group(1).toLowerCase(Locale.ENGLISH);
+                String premodifiers = m.group(1).toLowerCase();
                 int index = Integer.parseInt(m.group(2));
-                String postmodifiers = m.group(6).toLowerCase(Locale.ENGLISH);
+                String postmodifiers = m.group(6).toLowerCase();
```
Shredder121: These aren't necessary, these only contain `[sl%]` I believe.
Those are safely doable with `Locale.ENGLISH`.

Shredder121: Not that we will, but if we would add`i`, that would then fail in a Turkish `Locale`.

Shredder121:
This looks good, but we do need a few test that uses the Turkey Locale.
Because `i` behaves differently, and we want to make sure that the code works.

querydsl-sql-codegen/src/main/java/com/querydsl/sql/codegen/DefaultNamingStrategy.java
```diff
@@ -33,7 +31,7 @@ public DefaultNamingStrategy() {
 
     @Override
     public String getClassName(String tableName) {
-        String className = tableName.substring(0, 1).toUpperCase(Locale.ENGLISH) +
+        String className = tableName.substring(0, 1).toUpperCase() +
```
utkanozyurek: upercase tableName with system locale is not a good idea, java uses ENGLISH locale.
in example making className named "something" to "SOMETHİNG" in turkish doesn't work with java. 
query parameters should be loweCased and upperCased with system locale but field names, class names should use Locale.ENGLISH .

PR #1937, opened 2016-06-24
Make cross join overridable

### LINKED ISSUE #1932: SQLite CROSS JOIN possibility via JoinType.DEFAULT
SQLite has a separate `CROSS JOIN` (that can be used to disable the optimizer) and I noticed that the default comma join is documented as cross join so naturally I tried to override this functionality for my particular use case via `SQLTemplate`.

To my surprise this was hardcoded so I patched it customizable using the snipped below.

``` patch
@@ -196,6 +196,8 @@
 
     private String countStar = "count(*)";
 
+    private String crossJoin = ", ";
+
     private String delete = "delete ";
 
     private String desc = " desc";
@@ -571,6 +573,10 @@
         return countStar;
     }
 
+    public String getCrossJoin() {
+        return crossJoin;
+    }
+
     public final String getDelete() {
         return delete;
     }
@@ -626,7 +632,7 @@
             case FULLJOIN:  return fullJoin;
             case LEFTJOIN:  return leftJoin;
             case RIGHTJOIN: return rightJoin;
-            default:       return ", ";
+            default:       return crossJoin;
         }
     }
 
@@ -999,6 +1005,10 @@
         this.countStar = countStar;
     }
 
+    protected void setCrossJoin(String crossJoin) {
+        this.crossJoin = crossJoin;
+    }
+
     protected void setDelete(String delete) {
         this.delete = delete;
     }

```

timowest:
Fixes #1932

### DIFF
--- a/querydsl-sql/src/main/java/com/querydsl/sql/SQLTemplates.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/SQLTemplates.java
@@ -140,6 +140,8 @@ public class SQLTemplates extends Templates {
 
     private String countStar = "count(*)";
 
+    private String crossJoin = ", ";
+
     private String delete = "delete ";
 
     private String desc = " desc";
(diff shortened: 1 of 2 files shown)


Shredder121:
An added test that overrides `crossJoin`?
Just to see if it is inserted in the right place.
But looks good to me. :+1:

PR #1940, opened 2016-06-29
Join for dynamic entity path alias, querydsl-sql needs implementation

### LINKED ISSUE #1936: Join for dynamic entity path alias, querydsl-sql needs implementation for EntityPath and alias.
In case of querydsl ProjectableSQLQuery of querydsl-sql, we have following implentation:

```
  @Override
    public Q innerJoin(EntityPath<?> target) {
        return queryMixin.innerJoin(target);
    }

    @Override
    public <E> Q innerJoin(RelationalFunctionCall<E> target, Path<E> alias) {
        return queryMixin.innerJoin(target, alias);
    }

    @Override
    public Q innerJoin(SubQueryExpression<?> target, Path<?> alias) {
        return queryMixin.innerJoin(target, alias);
    }

    @Override
    public <E> Q innerJoin(ForeignKey<E> key, RelationalPath<E> entity) {
        return queryMixin.innerJoin(entity).on(key.on(entity));
    }
```

It should also have joins with:

```
@Override
public Q innerJoin(EntityPath<T> target, Path<T> alias) {
	return queryMixin.innerJoin(target, alias);
}

```
This is present in querydsl-jpa but not present for joins in querydsl-sql.

tan31989:
Closes #1936 
1. Added 4 join related method implementations: join, innerJoin, fullJoin and rightJoin for dynamic query building, which would be without code generation.
2. The fix will solve the implementation required to build a join clause

What it solves:

```java
SQLTemplates templates = MySQLTemplates.builder().printSchema().build();
Configuration configuration = new Configuration(templates);
configuration.setUseLiterals(true);
SQLQueryFactory queryFactory = new SQLQueryFactory(configuration, sqlDataSource); //DataSource sqlDataSource

Path<Object> customerPath = new PathBuilder<Object>(Object.class, "customer");
SQLQuery<?> sqlQuery = queryFactory.from(customerPath.as("c"));

PathBuilder<Object> orderPath = new PathBuilder<Object>(Object.class, "order");
Path<Object> aliasPath = new PathBuilder<Object>(Object.class, "o");

sqlQuery = sqlQuery.innerJoin(orderPath, aliasPath); // new innerJoin method implemented

sqlQuery = sqlQuery.select(SQLExpressions.all);
System.out.println(sqlQuery.getSQL().getSQL());
```

Will print: `select * from customer as c inner join order as o`

### DIFF
--- a/querydsl-sql/src/main/java/com/querydsl/sql/SQLCommonQuery.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/SQLCommonQuery.java
@@ -106,6 +106,16 @@ public interface SQLCommonQuery<Q extends SQLCommonQuery<Q>> extends Query<Q> {
      * @param alias alias
      * @return the current object
      */
+    <E> Q fullJoin(EntityPath<E> o, Path<E> alias);
+
+    /**
+     * Adds a full join to the given target
+     *
+     * @param <E>
+     * @param o full join target
+     * @param alias alias
+     * @return the current object
+     */
     <E> Q fullJoin(RelationalFunctionCall<E> o, Path<E> alias);
 
     /**
@@ -134,6 +144,16 @@ public interface SQLCommonQuery<Q extends SQLCommonQuery<Q>> extends Query<Q> {
      * @return the current object
      */
     Q innerJoin(EntityPath<?> o);
+
+    /**
+     * Adds an inner join to the given target
+     *
+     * @param <E>
+     * @param o
+     * @param alias alias
+     * @return the current object
+     */
+    <E> Q innerJoin(EntityPath<E> o, Path<E> alias);
 
     /**
      * Adds a full join to the given target
@@ -180,6 +200,16 @@ public interface SQLCommonQuery<Q extends SQLCommonQuery<Q>> extends Query<Q> {
      * @param alias alias
      * @return the current object
      */
+    <E> Q join(EntityPath<E> o, Path<E> alias);
+
+    /**
+     * Adds a full join to the given target
+     *
+     * @param <E>
+     * @param o join target
+     * @param alias alias
+     * @return the current object
+     */
     <E> Q join(RelationalFunctionCall<E> o, Path<E> alias);
 
     /**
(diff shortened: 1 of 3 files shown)


querydsl-sql/src/main/java/com/querydsl/sql/SQLCommonQuery.java
```diff
@@ -180,6 +200,16 @@
      * @param alias alias
      * @return the current object
      */
+    <E> Q join(EntityPath<E> o, Path<E> alias);
```
Shredder121: Javadoc of this method is not right yet (mentions full join)

querydsl-sql/src/main/java/com/querydsl/sql/SQLCommonQuery.java
```diff
@@ -213,6 +243,16 @@
      * Adds a full join to the given target
      *
      * @param <E>
+     * @param o left join target
+     * @param alias alias
+     * @return the current object
+     */
+    <E> Q leftJoin(EntityPath<E> o, Path<E> alias);
```
Shredder121: Javadoc of this method is not right yet (mentions full join)

querydsl-sql/src/main/java/com/querydsl/sql/SQLCommonQuery.java
```diff
@@ -258,6 +298,16 @@
      * Adds a full join to the given target
      *
      * @param <E>
+     * @param o right join target
+     * @param alias alias
+     * @return the current object
+     */
+    <E> Q rightJoin(EntityPath<E> o, Path<E> alias);
```
Shredder121: Javadoc of this method is not right yet (mentions full join)

Shredder121:
I added some comments on the interface methods.

Right now, only the new methods are added, which compiles and verifies fine.
We also need a test (preferably a few), using those new methods, to actually have some coverage.

If you add a few in [SelectBase, like these](https://github.com/querydsl/querydsl/blob/master/querydsl-sql/src/test/java/com/querydsl/sql/SelectBase.java#L921-L960) They will be picked up in the various suites.

Something close to your snippet in the Pull Request sounds fine.

tan31989:
@Shredder121 Have added several test cases, which I thought would best fit in already existing implementations of test cases which point for dynamic path along side other things. (`SQLSerializerTest`)

Also, thanks for letting me know about the `Closes` key.

PR #1942, opened 2016-06-29
Remove dots from MDC log fields.

### LINKED ISSUE #1941: QueryDSL MDC logging fields contain dots
QueryDSL MDC logging fields contain dots in the field names. All logging infrastructure based on ELK (Elasticsearch, Logstash, Kibana) will require the dots in fields to be removed when upgrading to 2.0 (see: https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking_20_mapping_changes.html#_field_names_may_not_contain_dots).

utwyko:
Fixes #1941

### DIFF
--- a/querydsl-core/src/main/java/com/querydsl/core/support/QueryBase.java
+++ b/querydsl-core/src/main/java/com/querydsl/core/support/QueryBase.java
@@ -13,13 +13,13 @@
  */
 package com.querydsl.core.support;
 
-import javax.annotation.Nonnegative;
-
 import com.querydsl.core.QueryModifiers;
 import com.querydsl.core.types.Expression;
 import com.querydsl.core.types.OrderSpecifier;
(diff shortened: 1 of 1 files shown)


timowest:
Maybe the backwards compatibility check is too strict here?

Shredder121:
Why is it too strict? You shouldn't change stuff in public members that are not backwards compatible.
So, the best we can do I think is push this back to a minor release.

Shredder121:
So, about the minor release.
@querydsl/developers Do you think we should include this is 4.2?

Shredder121:
The easiest would be to do this in Querydsl 5.
Just so we can be sure we're not breaking things accidentally.

Then we'll break things on purpose, but then it will at least be expected with a major version bump.

jwgmeligmeyling:
Applied to master.

PR #1946, opened 2016-06-30
Fix docs

### LINKED ISSUE #1938: docs - section 2.3.8 - joins - possible error in example
it says alternative way to write join is

    query.from(customer)
         .leftJoin(company).on(customer.company.eq(company.id))
         .list(customer.firstName, customer.lastName, company.name)

shouldn't predicate condition be like customer.company.**id**.eq(company.id) ?

timowest:
Fixes #1938


Shredder121:
Ah, looks good. :+1:

PR #1948, opened 2016-06-30
Querydsl 4.1.3

timowest:


Shredder121:
`querydsl-docs` versions are not updated here.

Shredder121:
I think.

timowest:
no, it was only on the 3.7.4 side

PR #1883, opened 2016-05-12
Added note on requirement of columnAnnotation for ExtendedBeanSerializer

robertandrewbain:
closes #1873
Related to #1868

PR #1836, opened 2016-03-31
Make asDBKey and asDBValue protected

timowest:
Fixes #1832

PR #1820, opened 2016-03-14
Add test

timowest:
Related to #1818

PR #1804, opened 2016-02-25
Add test case

timowest:
Related to #1803

PR #1950, opened 2016-07-01
Add PullApprove config

timowest:


Shredder121:
How about adding this one, so we can't merge `progress` PRs?
[docs](http://docs.pullapprove.com/yaml/always_pending/)
```yaml
always_pending:
  labels: progress
```

.pullaprove.yml
```diff
@@ -0,0 +1,14 @@
+approve_by_comment: true
+approve_regex: (?i)^Approved
```
Shredder121: Maybe something close to the one on the [docs website](http://docs.pullapprove.com/yaml/approve_regex/)?
```yaml
approve_regex: '(?i)^(approved|looks good|:shipit:|:\+1:)'
```
Then "Looks good to me." would for example match as well.

Shredder121:
Can/will you also turn off [`author_approval`](http://docs.pullapprove.com/yaml/author_approval/)?
Then the dashboard looks better with the two lists (Your PRs, PRs requiring approval).

Shredder121:
A little annoying that you can't test out the configuration.
I suppose we'll have to see what happens if it can't read the configuration correctly.

Shredder121:
How many approvals shall we do? 2?
Or just 1, and preferably someone else merging it?
We don't want to lock up the repo when most of us are on vacation I suppose.

timowest:
One approval is enough, but it needs to be someone else.

Shredder121:
I won't go over the top with the regex, but what do you think of this:
```
(?i)^(approved|(this (also )?)?looks (good|fine|okay)|:shipit:|:\+1:)
```
This means that these would match:
https://github.com/querydsl/querydsl/pull/1816#issuecomment-203996318
https://github.com/querydsl/querydsl/pull/1883#issuecomment-219216111
https://github.com/querydsl/querydsl/pull/1883#issuecomment-219217235
https://github.com/querydsl/querydsl/pull/1804#issuecomment-216073271
https://github.com/querydsl/querydsl/pull/1900#issuecomment-226970951
https://github.com/querydsl/querydsl/pull/1900#issuecomment-229086241
https://github.com/querydsl/querydsl/pull/1820#issuecomment-201258515

The only few so far encountered that wouldn't match are somewhat like these:
https://github.com/querydsl/querydsl/pull/1841#issuecomment-216337601
https://github.com/querydsl/querydsl/pull/1836#issuecomment-203997474
But I'll just have to go easy on the responses.

timowest:
I prefer the current state of the regex, let's use a consistent style

Shredder121:
Fair enough.

Shredder121:
It seems that the configuration in `master` is used, so I can not yet approve this.
Is there anything you want to change?

timowest:
Looks good

Shredder121:
I'm not _sure_ whether they already implemented the new issue webhook events (edits).
So I think, for now, you'll have to make a new comment.

Shredder121:
Also, what does your 'merge pull request' button look like?
Is it grey, or red because you are an admin?

timowest:
Approved

Issue #1955, opened 2016-07-04
Foreign key is generated even NamingStrategy.shouldGenerateForeignKey returns false


balazs-zsoldos:
If there are multiple foreign keys in a class and some should be generated, all will be generated.

Reason in MetadataExporter class:

```
if (!foreignKeyData.isEmpty()) {
    for (ForeignKeyData fkd : foreignKeyData.values()) {
        if (namingStrategy.shouldGenerateForeignKey(schemaAndTable, fkd)) {
            classModel.getData().put(ForeignKeyData.class, foreignKeyData.values());
        }
    }
}
```

All entries in the foreignKeyData is passed to the data map.

PR #1956, opened 2016-07-04
 #1955 Fix foreign key generation in MetadataExp

balazs-zsoldos:
I collect the foreign keys that should be generated
into a set and only add them to the classModel if
there is any foreign key.

### DIFF
--- a/querydsl-sql-codegen/src/main/java/com/querydsl/sql/codegen/MetaDataExporter.java
+++ b/querydsl-sql-codegen/src/main/java/com/querydsl/sql/codegen/MetaDataExporter.java
@@ -356,10 +356,15 @@ public class MetaDataExporter {
                 Map<String,ForeignKeyData> foreignKeyData = keyDataFactory
                         .getImportedKeys(md, catalog, schema, tableName);
                 if (!foreignKeyData.isEmpty()) {
+                    Collection<ForeignKeyData> foreignKeysToGenerate = new HashSet<ForeignKeyData>();
                     for (ForeignKeyData fkd : foreignKeyData.values()) {
                         if (namingStrategy.shouldGenerateForeignKey(schemaAndTable, fkd)) {
-                            classModel.getData().put(ForeignKeyData.class, foreignKeyData.values());
+                            foreignKeysToGenerate.add(fkd);
                         }
+                    }
+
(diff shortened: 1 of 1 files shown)


querydsl-sql-codegen/src/main/java/com/querydsl/sql/codegen/MetaDataExporter.java
```diff
@@ -356,11 +356,16 @@ private void handleTable(DatabaseMetaData md, ResultSet tables) throws SQLExcept
...
                 if (!foreignKeyData.isEmpty()) {
+                    Collection<ForeignKeyData> foreignKeysToGenerate = new HashSet<ForeignKeyData>();
                     for (ForeignKeyData fkd : foreignKeyData.values()) {
                         if (namingStrategy.shouldGenerateForeignKey(schemaAndTable, fkd)) {
-                            classModel.getData().put(ForeignKeyData.class, foreignKeyData.values());
+                            foreignKeysToGenerate.add(fkd);
                         }
                     }
+
+                    if (foreignKeysToGenerate.size() > 0) {
```
Shredder121: Could you change this to:
```java
if (!foreignKeysToGenerate.isEmpty()) 
```

Shredder121:
Yeah, this is indeed an issue.
Thanks for opening up a PR.
I left a comment in the diff, but I don't know whether others have comments or not.

querydsl-sql-codegen/src/main/java/com/querydsl/sql/codegen/MetaDataExporter.java
balazs-zsoldos: @Shredder121 Changed it.

timowest:
Maybe still squash the commits

balazs-zsoldos:
If I look at the _Files changed_ tab, I see that the isEmpty function is used now.

timowest:
I meant to combine the two commits into one. The end result is fine.

balazs-zsoldos:
@timowest  I see. I have never done it before, but I think I managed to merge the two commits.

timowest:
Looks good, thanks a lot for the contribution @balazs-zsoldos!

Shredder121:
Look good to me as well.
I'll merge it then.
@balazs-zsoldos Thanks for the contribution!

Issue #1958, opened 2016-07-05
NPE: CreateExpression is not thread safe, yet is being shared in multithreaded environment


oxygenecore:
The following line occasionally throws NPEs:
https://github.com/querydsl/querydsl/blob/master/querydsl-core/src/main/java/com/querydsl/core/types/ConstructorExpression.java#L135
This is caused by the fact that the very same instance is being shared across different threads while in multithreaded environment. Supposedly, one of these threads succeeds on executing line 132 on uninitialized instance, while the other thread bypasses the switch on line 131 and directly goes to line 135. In this scenario `transformers` field is not yet initialized, and for-loop throws an NPE.

The testing conditions could be recreated here: https://gist.github.com/oxygenecore/60c909b9a3814301e4b2e287b4f872b5. I haven't come up with a better solution to expose that the single instance is being initialized concurrently other than just adding debug logging from IntelliJ Idea (ConstructorExpression:132, breakpoint: evaluate and display message 'hashCode()' on pass): while running tests you see multiple repeating lines with the same hashCode.

PR for this issue: https://github.com/querydsl/querydsl/pull/1959

oxygenecore:
@timowest do you have any estimations on when this fix will be released?

PR #1959, opened 2016-07-05
Bypassing NPE in multithreaded environment

oxygenecore:
This PR avoids NPEs thrown by ConstructorExpression in multithreaded environment. This PR solves this related issue: https://github.com/querydsl/querydsl/issues/1958

### DIFF
--- a/querydsl-core/src/main/java/com/querydsl/core/types/ConstructorExpression.java
+++ b/querydsl-core/src/main/java/com/querydsl/core/types/ConstructorExpression.java
@@ -128,9 +128,11 @@ public class ConstructorExpression<T> extends FactoryExpressionBase<T> {
     @SuppressWarnings("unchecked")
     public T newInstance(Object... args) {
         try {
-            if (constructor == null) {
-                constructor = getConstructor(getType(), parameterTypes);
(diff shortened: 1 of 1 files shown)


timowest:
Alternatively directly initialize it in the constructor, that would be the simplest

oxygenecore:
@timowest probably it is, unless there is implicit dependency which forces this initialization be outside of a constructor.

timowest:
No dependency, also the lazy init doesn't improve anything here. @oxygenecore Thanks for the PR!

oxygenecore:
We are really looking into solving this issue.@timowest, do you think it's better for me to create new PR with changes proposed by you, or you could pick this up?

Shredder121:
> create new PR with changes proposed by you

It's not necessary to create a new PR.
You can force push the new changes, then the PR's diff will be updated.
Then we can keep the discussion here in this PR.

timowest:
Here is the version with eager init https://github.com/querydsl/querydsl/pull/1970

oxygenecore:
@timowest, thank you for taking care of it, I hung with my current work and haven't been able to progress on this.

PR #1970, opened 2016-07-16
Initialize ConstructorExpression eagerly

### LINKED ISSUE #1958: NPE: CreateExpression is not thread safe, yet is being shared in multithreaded environment

timowest:
Fixes #1958

### DIFF
--- a/querydsl-core/src/main/java/com/querydsl/core/types/ConstructorExpression.java
+++ b/querydsl-core/src/main/java/com/querydsl/core/types/ConstructorExpression.java
@@ -15,7 +15,10 @@ package com.querydsl.core.types;
 
 import static com.querydsl.core.util.ConstructorUtils.*;
 
+import java.io.IOException;
+import java.io.ObjectInputStream;
 import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
 import java.util.Arrays;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


Shredder121:
Unfortunately, this break serialization, since the field is then not set.
I recommend not using final fields, and then also setting them during deserialization.
Or what did you have in mind?

timowest:
Good catch, I'll try to fix the serialization.

Shredder121:
Looks good now. :smile:

PR #1971, opened 2016-07-16
Fix null handling

### LINKED ISSUE #1969: Fetch result with ListPath.any().{field} cause NullPointerException
I have a model defined like this:

```
public class User {
	public List<Address> addresses;
}
```

And then, fetch with this command
`CollQueryFactory.from(QUser.user, users).where(predicate).fetchResults().getResults()` which `predicate` is defined as `QUser.user.addresses.any().city.eq("SomeCity")`

Also, `users` variable contains list of user and there is one user that has value of `addresses` field as `null` (it doesn't be an empty list).

It throws exception like this
```
java.lang.IllegalArgumentException: java.lang.reflect.InvocationTargetException

	at com.mysema.codegen.MethodEvaluator.evaluate(MethodEvaluator.java:52)
	at com.querydsl.collections.DefaultQueryEngine.evaluateMultipleSources(DefaultQueryEngine.java:116)
	at com.querydsl.collections.DefaultQueryEngine.count(DefaultQueryEngine.java:57)
	at com.querydsl.collections.AbstractCollQuery.fetchResults(AbstractCollQuery.java:189)
	...
Caused by: java.lang.reflect.InvocationTargetException
	at com.mysema.codegen.MethodEvaluator.evaluate(MethodEvaluator.java:48)
	... 32 more
Caused by: java.lang.NullPointerException
	at com.querydsl.collections.CollQueryFunctions.leftJoin(CollQueryFunctions.java:198)
	at Q_715591299_65821278_1275614662_1195259493.eval(Unknown Source)
	... 37 more
```

PS. I'm using querydsl 4.1.2

timowest:
Fixes #1969

### DIFF
--- a/querydsl-collections/src/main/java/com/querydsl/collections/CollQueryFunctions.java
+++ b/querydsl-collections/src/main/java/com/querydsl/collections/CollQueryFunctions.java
@@ -195,7 +195,7 @@ public final class CollQueryFunctions {
     }
 
     public static <T> Collection<T> leftJoin(Collection<T> coll) {
-        if (coll.isEmpty()) {
+        if (coll == null || coll.isEmpty()) {
             @SuppressWarnings("unchecked") // List only contains null
             Collection<T> rv = (Collection<T>) nullList;
             return rv;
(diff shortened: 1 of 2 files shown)


Shredder121:
Looks good to me.

PR #1972, opened 2016-07-18
Fixed map grouping by, was generating an invalid jpql query

velo:
Query dsl was generating a query like this:
```
select show_acts_0
from Show show
  left join show.acts as show_acts_0 with key(show_acts_0) = ?1
```

So I included convert to groupBy as well, same as setProjection()


timowest:
Looks good
Thanks for the contribution @velo!

Issue #1961, opened 2016-07-06
HsqlDb creates parentheses for Union


robertcsakany:
With HSQLTemplate the folowwing query generates syntactiically incorrect Query

'''

                QUser u = new QUser("u");
                QUserRole ur = new QUserRole("ur");
                QRoleRole rr = new QRoleRole("rr");
                QPermission p = new QPermission("p");
                QRolePerm rp = new QRolePerm("rp");

                PathBuilder<Tuple> sub = new PathBuilder(Tuple.class, "sub");

                return query
                        .withRecursive(sub, new Path[] {sub.get(C_ROLES_ID)}).as(
                        unionAll(
                                select(ur.roles.as(C_ROLES_ID)).from(u).join(ur).on(ur.users.eq(u.id)).where(u.id.eq(uid)),
                                select(rr.children.as(C_ROLES_ID)).from(rr).join(sub).on(rr.parents.eq(sub.get(C_ROLES_ID)))
                        )
                ).select(bean(Permission.class, p.all()))
                        .from(sub).join(rp).on(rp.roles.eq(sub.get(C_ROLES_ID))).join(p).on(p.id.eq(rp.permissions)).fetch();

'''

Generates the folowwing HSQL code:

'''
with recursive sub (C_ROLES_ID) as ((select ur.C_ROLES_ID as C_ROLES_ID
from V_FRW_USER u
join T_USER_ROLE_ED3759 ur
on ur.C_USERS_ID = u.id
where u.id = 1)
union all
(select rr.C_CHILDREN_ID as C_ROLES_ID
from T_ROLE_ROLE_BC67D7 rr
join sub
on rr.C_PARENTS_ID = sub.C_ROLES_ID))
select p.id, p.C_XMIID, p.C_RESOURCE, p.C_PERMID, p.C_GUID, p.C_AUU, p.C_AUT, p.C_ACU, p.C_ACT, p.C_ACTION
from sub
join T_ROLE_PERM_446F1D rp
on rp.C_ROLES_ID = sub.C_ROLES_ID
join V_FRW_PERM p
on p.id = rp.C_PERMS_ID
'''

Which is incorrect. 

In HSQL DB the folowwing sytax is correct:

'''
with recursive sub (C_ROLES_ID) as (select ur.C_ROLES_ID as C_ROLES_ID
from V_FRW_USER u
join T_USER_ROLE_ED3759 ur
on ur.C_USERS_ID = u.id
where u.id = 1
union all
[...]

PR #1973, opened 2016-07-18
#1961 HSQLDB dont use parenthes in Union.

robertcsakany:
HsqldbSuteTest and HsqldbLiteralsSuiteTest are OK.


querydsl-sql/src/test/java/com/querydsl/sql/AbstractSQLTemplatesTest.java
```diff
@@ -70,19 +70,19 @@ public void union() {
 
         if (templates.getDummyTable() == null) {
             assertEquals(
-                    "(select 1 as col1)\n" +
```
timowest: this should be changed in the subclass, look at the test failures
alternatively parametrize it better

robertcsakany: Is it a good idea to reuse the wrapping parameter when asserting the SQL?  Means the parentheses is built in asserted SQL string when the wrapping is true?

timowest: Yes, you can do it like that, alternatively just override the method in the subclass.

timowest:
Looks good

Shredder121:
Looks good to me as well.
I'll merge it.

Shredder121:
@robertcsakany thanks for the contribution!

PR #1974, opened 2016-07-19
Normalization and quoted things

Shredder121:
It's not getting any prettier.
One thing we could maybe do is split on quotes, and only process all odd numbered indexes. (string mode off)


Shredder121:
Repushing, one commit isn't being built.

Shredder121:
Forgot my `progress` label.
There are still some cases that fail.

jwgmeligmeyling:
I don't think we intend maintaining QueryDSL 3 anymore and as far as I am aware,  normalisation and quotation in queries was actually improved in QueryDSL by a fair bit. Please reopen if I am mistaken.

PR #1979, opened 2016-07-28
Add extra parenthesis to fix precedence issues

### LINKED ISSUE #1978: IN clause splitting produces incorrect query for Oracle
The following code

```
db.from(card)
  .select(card.id, card.accepted)
  .where(card.accepted.isTrue())
  .where(card.id.in(list))
  .fetch()
```

produces the following SQL:

```
SELECT ID, ACCEPTED FROM CARD WHERE ACCEPTED = 1 AND ID IN (20378,20379,...) OR ID IN (20380,20381)
```

According to the Oracle documentation (https://docs.oracle.com/cd/B19306_01/server.102/b14200/conditions001.htm#i1034834) this should be correct, but it isn't. AND and OR behave as having the same precedence.

The result is that the first clause is only applied to the first 1000 items.

Tested on QueryDSL 4.1.3. Manually overriding OracleTemplates and changing the precedence of the OR operator does not produce a different result.

ntkoopman:
Attempt to fix #1978

### DIFF
--- a/querydsl-sql/src/main/java/com/querydsl/sql/SQLSerializer.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/SQLSerializer.java
@@ -986,7 +986,9 @@ public class SQLSerializer extends SerializerBase<SQLSerializer> {
                     } else {
                         result = ExpressionUtils.notInAny(path, partitioned);
                     }
+                    append("(");
                     result.accept(this, null);
+                    append(")");
                 }
             }
 


Shredder121:
In `OracleTemplates` you can `setPrecedence` for `Ops.AND`/`Ops.OR`.
Because mostly we already assume that the precedences are clean.

_NOTE: I am however out for today, so this is mostly me just writing up something to come back to._

ntkoopman:
Changing the precedence in OracleTemplates does not change the behaviour. The current precedence is also correct.

timowest:
The issue is not the precedence, but the IN operation handling, closing in favor of https://github.com/querydsl/querydsl/pull/1988

PR #1983, opened 2016-08-04
insertion order for JPA query hints

encircled:
Some JPA hints are order-dependent

### DIFF
--- a/querydsl-jpa/src/main/java/com/querydsl/jpa/impl/AbstractJPAQuery.java
+++ b/querydsl-jpa/src/main/java/com/querydsl/jpa/impl/AbstractJPAQuery.java
@@ -27,7 +27,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.slf4j.MDC;
 
-import com.google.common.collect.HashMultimap;
+import com.google.common.collect.LinkedHashMultimap;
 import com.google.common.collect.Multimap;
 import com.mysema.commons.lang.CloseableIterator;
 import com.querydsl.core.*;
@@ -50,7 +50,7 @@ public abstract class AbstractJPAQuery<T, Q extends AbstractJPAQuery<T, Q>> exte
 
     private static final Logger logger = LoggerFactory.getLogger(JPAQuery.class);
 
-    protected final Multimap<String,Object> hints = HashMultimap.create();
+    protected final Multimap<String,Object> hints = LinkedHashMultimap.create();
 
     protected final EntityManager entityManager;
 


timowest:
Are you sure ArrayListMultimap does what you want? It uses a HashMap internally for key -> values mapping. The only difference is that the values for a given key are iterated in insertion order.

encircled:
Insertion order of values is exactly what we need, PR description is a bit misleading.
The initial problem is that eclipselink requires hints for nested fetch joins and there is a bug (or feature), that `parent.child` fetch hint must be before `child.nestedChild` one.

timowest:
Look at the javadoc for ArrayListMultimap. Iteration order of keys comes from HashMap, iteration order of values for a given key comes from ArrayList.

encircled:
I know, for fetch joins in eclipse there is one hint (key) and multiple values

timowest:
Ok, sorry, I misunderstood. parent.child and child.nestedChid are the values?

encircled:
Yes

Shredder121:
A `String` hashcode is probably still very good, so maybe a `LinkedHashMultimap` is a better pick?

Shredder121:
Looks good to me as well, other than that.
Just want to hear an opinion.

encircled:
LinkedHashMultimap uses Set for hint values, it will break the order

Shredder121:
There are Set implementations that preserve insertion order.

@timowest What is your view?

encircled:
Yes, but you can't change used implementation in LinkedHashMultimap, can you?

Shredder121:
The used implementation preserves insertion order.
The Javadoc is very specific about this:

> Implementation of Multimap that does not allow duplicate key-value entries and that returns collections **whose iterators follow the ordering in which the data was added to the multimap**.

Emphasis mine.
This means insertion order.

Or is there something I am missing?

http://google.github.io/guava/releases/snapshot/api/docs/com/google/common/collect/LinkedHashMultimap.html

encircled:
Insertion order of root keys is preserved indeed, but what do I need is insertion order of associated values  (nested collections)

Shredder121:
But that is preserved:
https://github.com/google/guava/wiki/NewCollectionTypesExplained#implementations-1

All documentation I find points to that, and a quick test also confirms that.

@timowest you can merge this if you want.

image

encircled:
Sorry, you are right. That Set has name WrappedSet, didn't mention it uses LinkedHashSet as a delegate.

timowest:
LinkedHashMultimap might be better.

encircled:
Indeed, I will change it

timowest:
Could you squash the commits, then we're good

encircled:
done

timowest:
@encircled Thanks a lot for the contribution!

Issue #1982, opened 2016-08-04
null check for joined entity not working


Kernle32DLL:
Hi there,

we encountered a strange problem while performing null checks with QueryDSL 3.7.4. Boiled down to its simplest form, we build a query like this:

```Java
JPAQuery testQuery = new JPAQuery(em);
testQuery.from(QPerson.person);
testQuery.where(QPerson.person.radiationExposedPerson.isNotNull());
```

Which yields the following JPQL query:

```SQL
select person
from person person
where person.radiationExposedPerson is not null
```

The Java classes are (condensed) like this:

```Java
@Entity(name = "person")
public class Person {

...

@OneToOne(mappedBy= "person", optional = true)
private RadiationExposedPerson radiationExposedPerson;

...

}

@Entity(name = "radiationexposedperson")
public class RadiationExposedPerson  {

...

@OneToOne
private Person person;

...

}
```

radiationExposedPerson is a mapped attribute in Person, and can be null. RadiationExposedPerson "owns" the one-to-one association to a Person (this makes sense in our case - keep in mind that this is a condensed version for the problem at hand). We were however not able to verify if this impacts the issue.

The expected output of the query above would be to show all persons, in which the radiationExposedPerson is not null. However, this does not work, and returns **all elements**. When the same query is modified to a "isNull" check, an **empty result** is returned. We verified that our database contains persons where radiationExposedPerson is null, and others where it is set.

Another thing. If the null check is used as an expression for ordering, we run into an AST Exception from hibernate:

```Java
JPAQuery testQuery = new JPAQuery(em);
testQuery.from(QPerson.person);
testQuery.orderBy(new OrderSpecifier<>(Order.ASC, QPerson.person.radiationExposedPerson.isNotNull()));
```
```SQL
select person
from person person
order by person.radiationExposedPerson is not null asc
```
```
org.hibernate.hql.internal.ast.QuerySyntaxException: unexpected AST node: is not null [select person
from de.com.steep.radoclid2.model.entities.Person person
order by person.radiationExposedPerson is not null asc]
	at org.hibernate.hql.internal.ast.QuerySyntaxException.convert(QuerySyntaxException.java:91)
	at org.hibernate.hql.internal.ast.ErrorCounter.throwQueryException(ErrorCounter.java:109)
	at org.hibernate.hql.internal.ast.QueryTranslatorImpl.analyze(QueryTranslatorImpl.java:284)
[...]

PR #1985, opened 2016-08-06
Add test

timowest:
Related to #1982


timowest:
@Shredder121 @johnktims @robertandrewbain Could I get a review?

Shredder121:
What are the semantics of `order by xxx is (not) null`? Is that even valid JPQL?

timowest:
It's ordering by an boolean expression, it's not valid JPQL, but supported by the JPA providers.

Shredder121:
So would it make sense to add that as a test case?

Shredder121:
:+1: I mean, it looks fine, but doesn't really reflect the issue.

timowest:
You are right, it's misleading.

PR #1986, opened 2016-08-06
Unify dependency versions

### LINKED ISSUE #1981: AbstractMethodError thrown by org.hibernate.internal.SessionFactoryImpl.<init> when upgrading from Spring Boot 1.3 to 1.4 RC1
Hi,

I tried to upgrade springboot from 1.3 to 1.4 and I get this same error. As someone pointed out there's a problem with the hibernate dependencies.
I thought that the same is happening with queryDsl as it uses different hibernate versions in different dependencies :
querydsl-apt -> 4.2.0.Final
querydsl-jpa -> 4.3.11.Final

How can I track this? 
Should querydDSL upgrade its dependencies to the latest?

reference link with same issue but difference depency.
[http://stackoverflow.com/questions/38596003/abstractmethoderror-thrown-by-org-hibernate-internal-sessionfactoryimpl-init-w/38599681#38599681](url)

And the issue I opened in springboot:

[url](https://github.com/spring-projects/spring-boot/issues/6525)

Thanks

timowest:
Fixes #1981

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -65,6 +65,9 @@
     <slf4j.version>1.6.1</slf4j.version>
     <surefire.version>2.18</surefire.version>
     <animal-sniffer.version>1.14</animal-sniffer.version>
+
+    <jdo.version>3.0.1</jdo.version>
+    <morphia.version>0.105</morphia.version>
     
     <!-- Import-Package definitions for maven-bundle-plugin -->
     <osgi.import.package.root>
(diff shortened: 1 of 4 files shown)


Shredder121:
How about slowly moving all occurrences of dependencies to `dependencyManagement`?
Then we can just touch the root pom to bump versions, and we are less likely to miss occurrences.

Just an idea though.

timowest:
Good idea, can we do it in an additional PR though?

Shredder121:
Looks good anyway. :+1:

And that is fine with me.

rmordillo-witbooking:
:+1: thanks @timowest

Shredder121:
Do you think we can we backport this @timowest?

PR #1988, opened 2016-08-06
Fix serialization

### LINKED ISSUE #1978: IN clause splitting produces incorrect query for Oracle

timowest:
Fixes #1978

### DIFF
--- a/querydsl-sql/src/main/java/com/querydsl/sql/SQLSerializer.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/SQLSerializer.java
@@ -986,7 +986,9 @@ public class SQLSerializer extends SerializerBase<SQLSerializer> {
                     } else {
                         result = ExpressionUtils.notInAny(path, partitioned);
                     }
+                    append("(");
                     result.accept(this, null);
+                    append(")");
                 }
             }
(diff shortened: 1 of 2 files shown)


timowest:
@Shredder121 @johnktims @robertandrewbain Could I get a review?

Shredder121:
:+1: Sorry, looks fine to me.

Shredder121:
Oh, we could also backport this I suppose.

PR #1990, opened 2016-08-06
Generate constructor with variable and schema parameters.

### LINKED ISSUE #1877: Generate constructor with var and schema parameters
I have a scenario where the database (mysql) is composed of a master schema and a schema for each client. The tables in each schema have the same structure.

```
mysql
  |
  +- master (db)
  |
  +- client_1 (db)
  |       |
  |       +- user (table)
  |       .
  |       .
  |
  +- client_2 (db)
  |       |
  |       +- user (table)
  |       .
  |       .
  .
  .
  .
  +- client_n (db)
```

I guess the solution is to construct the Q types with the schema 

```java
QUser u1 = new QUser("u", "client_1", "user");
QUser u2 = new QUser("u", "client_2", "user");
```
My work would be simplified if I had a constructor with only the variable and the schema:

```java
    public QUser(String variable, String schema) {
        super(User.class, forVariable(variable), schema, "USER");
        addMetadata();
    }
```

Using QueryDSL SQL 4.1.0

natros:
This is for the case where each customer has its own database with the same strucure using a single connection.

Fixes #1877

### DIFF
--- a/querydsl-sql-codegen/src/main/java/com/querydsl/sql/codegen/MetaDataSerializer.java
+++ b/querydsl-sql-codegen/src/main/java/com/querydsl/sql/codegen/MetaDataSerializer.java
@@ -25,6 +25,7 @@ import javax.inject.Named;
 
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
+
 import com.mysema.codegen.CodeWriter;
 import com.mysema.codegen.model.*;
 import com.querydsl.codegen.*;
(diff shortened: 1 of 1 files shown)


querydsl-sql-codegen/src/main/java/com/querydsl/sql/codegen/MetaDataSerializer.java
```diff
@@ -110,6 +111,13 @@ protected void constructorsForVariables(CodeWriter writer, EntityType model) thr
                 + "forVariable(variable), schema, table);");
         constructorContent(writer, model);
         writer.end();
+
+        writer.beginConstructor(new Parameter("variable", Types.STRING),
+                                new Parameter("schema", Types.STRING));
+        writer.line(SUPER, "(", writer.getClassConstant(localName) + COMMA
+                + "forVariable(variable), schema, \"", model.getData().get("table") + "\");");
```
Shredder121: Could you maybe change it so it doesn't concatenate non-constants?

Shredder121:
Sorry for the wait!
Added a comment.

querydsl-sql-codegen/src/main/java/com/querydsl/sql/codegen/MetaDataSerializer.java
natros: I'm under vacations for the next two weeks.

Shredder121: That's fine, do you want us to wait, or shall we take it from here?

natros: Take it please.

Thank you

Issue #1429, opened 2015-07-07
Reusing of constants in JPQL generation causes issues with hibernate query caching


strguntbr:
I know it's actually an hibernate issue (I also created an issue for hibernate https://hibernate.atlassian.net/browse/HHH-9900) but maybe there is some workaround to avoid the problem I'm not aware of.

QueryDSL reuses constants when generating JPQL in com.mysema.query.jpa.JPQLSerializer.visitConstant(Object constant) thus the following JPAQuery:

```
new JPAQuery(entityManager).from(QCacheTes.cacheTest).where(QCacheTest.cacheTest.boola.eq(true).and(QCacheTest.cacheTest.boolb.eq(false)).and(QCacheTest.cacheTest.boolc.eq(true)))
```

results in the following JPQL:

```
select cacheTest from at.bluesource.CacheTest cacheTest WHERE cacheTest.boola = ?1 AND cacheTest.boolb = ?2 AND cacheTest.boolc = ?1
```

using the paramters:

```
{ 1=true, 2=false }
```

which is problematic when using hibernate query caching as described in the hibernate issue.

Is there any workaround I'm not aware of to avoid those problems? I thought about using Params instead of the raw values (which become constants in the JPAQuery) to avoid the reusing of constants but on the other hand it's not practical because I can't make sure that everybody always uses Params in out project. Maybe there's a way to configure QueryDSL to not reuse constants (after having a look at the code of QueryDSL I'm pretty sure there is no such way - at least in 3.6.4 which is the version we are currently using)? Or maybe there is a way to somehow magically 'convert' all constants to params just before generation of the JPQL query?

Shredder121:
Hi.
This is related to #1413, you can use template expressions so that the underlying literal gets serialized *into* the query instead of being parameterized.

strguntbr:
Thanks for the suggestion, but I decided to try another aproach (because I'm, pretty sure that I would forget to use template expressions everywhere and I'm not sure if the template expression approach works with date/time objects)

I just implemented a workaround that treats constants as param expressions and seems to work (at least our tests are still passing and I have not yet found a problem. Maybe the better way would be to treat them like literals? But I think this would not work/create issues with date/time objects...

We are now just using our CustomJPAQuery instead of the JPAQuery globally. I know this workaround is not nice (at least the removed reflection part to get the inCaseOperation flag from the super class is pretty ugly)

```java
public class CustomJPAQuery extends JPAQuery {

[...]

Issue #1413, opened 2015-06-21
Incorrect parameter values with Hibernate custom types.


kwon37xi:
We are using querydsl 3.x, and there lots of legacy columns in our JPA Entity mappings, because of this we had to use many hibernate custom types.

When there are 
- the same java type(like Boolean) but different custom types'(like yes_no, true_false) columns related conditions 
- and their  parameter values are the same (like Boolean.TRUE)
- and parameters' position number is the same in a query(like ?1),

Hibernate maps parameter values incorrectly.
This was reported already - https://github.com/querydsl/querydsl/issues/1168
Actually I think this is a bug of hibernate - https://hibernate.atlassian.net/browse/HHH-9871 .
But QueryDSL can work around this bug.

For exampe, when there are two or more boolean types(same java type) but they're custom types like the following(it doesn't matter that the colums are in the same entity or two or more entities),

``` java
@Type(type = "true_false")
@Column(name = "employee", columnDefinition = "char(1)")
 private Boolean employee;

@Type(type = "yes_no")
@Column(name = "male", columnDefinition = "char(1)")
private Boolean male;

@Column(name = "old")
private Boolean old;
```

And in a QueryDsl query has conditions with same java type and same value(Boolean.TRUE), QueryDsl generates JPQL like the follwing,

```
// employee == Booealn,TRUE, male == Boolean.TRUE, old == Boolean.FALSE
from User user where employee = ?1 and male = ?1 and old = ?2
```

**QueryDSL maps same java type value into same positional parameter.**

real parameter value mappings in hibernate is like the following

```
employee = 'T' and male = 'T' and old = false
```

_male parameter must have been 'Y' but it follows the first parameter value - 'T'._

employee and male columns have same java type but different custom types, QueryDsl maps same positional parameter(?1)  to these two columns.

**I wonder QueryDSL could map different poisitional parameters for every different condition even though they have same java type value(like Boolean.TRUE in the example).**

timowest:
I'd rather not add a fix on the Querydsl level for this Hibernate bug, since maintaining those is quite time consuming.

But a valid workaround should be literals for these cases, e.g. BooleanTemplate.TRUE/FALSE

PR #2000, opened 2016-08-25
#1429 JPQL generation causes issues with hibernate query caching

l30thelion:
Please review potential solution for #1429 and #1413


timowest:
There seem to be a few backwards incompatible signature changes in the code. We can do those only in major releases.

l30thelion:
My proposed solution required the change of the method getConstantToLabel to getLabelToConstant to ensure that the constant labels are unique thereby avoiding the Hibernate bug.  I'm using the fix on my project at work.........this solves the issue, but is it a proper solution in your opinion?  Thanks for taking the time to reply!

timowest:
Based on my short review it looks like a proper solution, but for backwards compatible reason it can't be applied to a minor or patch release.

jwgmeligmeyling:
The PR is outdated and QueryDSL 3.x is no longer maintained. This fix would need to be updated.

PR #2008, opened 2016-08-31
add support for MongoDB GeoJSON Points

jmoghisi:
extends initial work done as part of '[Support for geospatial queries in Mongodb](https://bugs.launchpad.net/querydsl/+bug/727942)' to add support for querying using a GeoJSON point instead of legacy coordinates

upgrade MongoDB driver and Morphia versions

### DIFF
--- a/querydsl-mongodb/src/main/java/com/querydsl/mongodb/morphia/MorphiaSerializer.java
+++ b/querydsl-mongodb/src/main/java/com/querydsl/mongodb/morphia/MorphiaSerializer.java
@@ -20,9 +20,12 @@ import org.mongodb.morphia.Morphia;
 import org.mongodb.morphia.annotations.Id;
 import org.mongodb.morphia.annotations.Property;
 import org.mongodb.morphia.annotations.Reference;
+import org.mongodb.morphia.geo.Geometry;
+import org.mongodb.morphia.geo.GeometryQueryConverter;
 import org.mongodb.morphia.mapping.Mapper;
 
 import com.mongodb.DBRef;
+import com.querydsl.core.types.Constant;
 import com.querydsl.core.types.Path;
 import com.querydsl.core.types.PathMetadata;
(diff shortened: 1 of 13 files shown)


querydsl-mongodb/src/main/java/com/querydsl/mongodb/morphia/MorphiaSerializer.java
```diff
@@ -36,9 +39,20 @@
 public class MorphiaSerializer extends MongodbSerializer {
 
     private final Morphia morphia;
+    private final GeometryQueryConverter geometryQueryConverter;
 
     public MorphiaSerializer(Morphia morphia) {
-        this.morphia = morphia;
+        this.morphia = morphia == null ? new Morphia() : morphia;
```
timowest: this implies that null would be a valid argument to the constructor, which it shouldn't be

querydsl-mongodb/src/main/java/org/mongodb/morphia/geo/QPoint.java
```diff
@@ -0,0 +1,38 @@
...
+import com.querydsl.core.types.dsl.PathInits;
+import com.querydsl.mongodb.morphia.MorphiaExpressions;
+
+/**
+ * {@code QPoint} is an adapter type for Point to use geo spatial querying features of Mongodb
+ *
+ * @author jmoghisi
+ *
+ */
+public class QPoint extends EntityPathBase<Point> {
```
timowest: EntityPathBase is probably not the right superclass here, better something like
`extends SimpleExpression<Point> implements Path<Point>`

Look at StringPath for details

jmoghisi: I didn't fully understand this point, when is EntityPathBase intended to be used?

jmoghisi:
Hi, Please can you review the latest changes?

Issue #2007, opened 2016-08-31
[querydsl-sql] Possibility of specifying NameMapping programmatically


balazs-zsoldos:
There is a possibility in the Configuration class to override table and column names. This is a great feature. However, it would be really nice, implement a renaming algorithm in a custom class.

Use-cases:

Atlassian Jira

For example in Atlassian Jira, some tables are created with no quotes and tables beginning with AO prefix are created with upper-case. The implementation would check this and

if the table starts with AO, it would replace always the table and column names to be upper-case
otherwise it would replace the table name to have the same case as it is specified by the database (e.g.: upper-case in Oracle, lower-case in postgres, ...)
Resolving table names from database metadata

We get all table and column names from the database connection. We expect that there are no two tables with the same name but different mixture of upper- and lower-case letters. When the name-mapping implementation is called, it always replaces the table and column names to the ones that exist in the database.

**Conclusion**

We use quoting in our products with upper-case letters only to make querydsl work in every database. E.g.:

Oracle does not work with lower-case letters in special queries even if quoting is used
Postgresql JDBC driver quotes table names always if insert-with-key is used. Therefore we must provide exactly the same name as it is used in the database.
We cannot use our naming strategy in case of 3rd products (like Jira, Liferay, Alfresco, ...) as some of them do not use quoting. However, name-mapping could be solved very easily with a primitive implementation of an interface in these situations.

See discussion at https://groups.google.com/forum/#!topic/querydsl/E-Uny6TnDFY

balazs-zsoldos:
Hi,

I am creating a patch at https://github.com/balazs-zsoldos/querydsl/tree/i2007_Programmatic_Name_Mapping

See the commit: https://github.com/balazs-zsoldos/querydsl/commit/d9a1105a537843e69964db7a09c1ac70841b4a54

Please check if the change is not too "complex" for you. In summary:

 - I created a new package called namemapping
 - I created a new interface called NameMapping that has similar methods to Configuration. However, these methods return Optional values. If there is no mapping, it returns Optional.absent();

NameMapping has three implementations:

 - __PreConfiguredNameMapping__: This class contains the logic that was available till now
 - __CaseChangeNameMapping__: Changes all identifiers to upper-case or lower-case letters based on the parameter that is passed via its constructor. This mapping will be often used in my opinion
 - __ChainedNameMapping__: This implementation can wrap multiple mappings and returns the first non-null overridden value

[...]

PR #2016, opened 2016-09-13
Adding dynamic name mapping functionality (#2007)

balazs-zsoldos:
Forgot to send the pull request. Please let me know if I should change anything or if you see issues in the concept.

### DIFF
--- a/querydsl-sql/src/main/java/com/querydsl/sql/Configuration.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/Configuration.java
@@ -26,7 +26,6 @@ import javax.annotation.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.google.common.base.MoreObjects;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Maps;
 import com.google.common.primitives.Primitives;
--- a/querydsl-sql/src/main/java/com/querydsl/sql/Configuration.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/Configuration.java
@@ -26,10 +26,14 @@ import javax.annotation.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.base.MoreObjects;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Maps;
 import com.google.common.primitives.Primitives;
 import com.querydsl.core.types.Path;
+import com.querydsl.sql.namemapping.ChainedNameMapping;
+import com.querydsl.sql.namemapping.NameMapping;
+import com.querydsl.sql.namemapping.PreConfiguredNameMapping;
 import com.querydsl.sql.types.ArrayType;
 import com.querydsl.sql.types.Null;
 import com.querydsl.sql.types.Type;
@@ -50,7 +54,11 @@ public final class Configuration {
 
     private final JavaTypeMapping javaTypeMapping = new JavaTypeMapping();
 
-    private final NameMapping nameMapping = new NameMapping();
+    private final PreConfiguredNameMapping internalNameMapping = new PreConfiguredNameMapping();
+
+    private NameMapping nameMapping = internalNameMapping;
+
+    private final Map<String, String> schemaMapping = Maps.newHashMap();
 
     private final Map<String, Class<?>> typeToName = Maps.newHashMap();
 
@@ -196,7 +204,11 @@ public final class Configuration {
      */
     @Nullable
     public SchemaAndTable getOverride(SchemaAndTable key) {
-        return nameMapping.getOverride(key);
+        SchemaAndTable result = nameMapping.getOverride(key).or(key);
+        if (schemaMapping.containsKey(key.getSchema())) {
+            result = new SchemaAndTable(schemaMapping.get(key.getSchema()), result.getTable());
+        }
+        return result;
     }
 
     /**
@@ -207,7 +219,24 @@ public final class Configuration {
      * @return overridden column
      */
     public String getColumnOverride(SchemaAndTable key, String column) {
-        return nameMapping.getColumnOverride(key, column);
+        return nameMapping.getColumnOverride(key, column).or(column);
+    }
+
+    /**
+     * Programmers can specify name mappings by implementing the
+     * {@link NameMapping} interface. The mapping rules that are specified by
+     * this property are checked if no mapping is specified by any of the
+     * <code>register*Override</code> functions.
+     *
+     * @param nameMapping
+     *            The name mapping that is implemented by the user.
+     */
+    public void setDynamicNameMapping(NameMapping nameMapping) {
+        if (nameMapping == null) {
+            this.nameMapping = this.internalNameMapping;
+        } else {
+            this.nameMapping = new ChainedNameMapping(this.nameMapping, nameMapping);
+        }
     }
 
     /**
@@ -290,7 +319,7 @@ public final class Configuration {
      * @return previous override value
      */
     public String registerSchemaOverride(String oldSchema, String newSchema) {
-        return nameMapping.registerSchemaOverride(oldSchema, newSchema);
+        return schemaMapping.put(oldSchema, newSchema);
     }
 
     /**
@@ -301,7 +330,7 @@ public final class Configuration {
      * @return previous override value
      */
     public String registerTableOverride(String oldTable, String newTable) {
-        return nameMapping.registerTableOverride(oldTable, newTable);
+        return internalNameMapping.registerTableOverride(oldTable, newTable);
     }
 
     /**
@@ -338,7 +367,7 @@ public final class Configuration {
      * @return previous override
      */
     public SchemaAndTable registerTableOverride(SchemaAndTable from, SchemaAndTable to) {
-        return nameMapping.registerTableOverride(from, to);
+        return internalNameMapping.registerTableOverride(from, to);
     }
 
     /**
@@ -351,7 +380,7 @@ public final class Configuration {
      * @return previous override
      */
     public String registerColumnOverride(String schema, String table, String oldColumn, String newColumn) {
-        return nameMapping.registerColumnOverride(schema, table, oldColumn, newColumn);
+        return internalNameMapping.registerColumnOverride(schema, table, oldColumn, newColumn);
     }
 
     /**
@@ -363,7 +392,7 @@ public final class Configuration {
      * @return previous override
      */
     public String registerColumnOverride(String table, String oldColumn, String newColumn) {
-        return nameMapping.registerColumnOverride(table, oldColumn, newColumn);
+        return internalNameMapping.registerColumnOverride(table, oldColumn, newColumn);
     }
 
     /**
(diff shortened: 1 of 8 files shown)


querydsl-sql/src/main/java/com/querydsl/sql/Configuration.java
```diff
@@ -301,7 +329,7 @@ public String registerSchemaOverride(String oldSchema, String newSchema) {
      * @return previous override value
      */
     public String registerTableOverride(String oldTable, String newTable) {
-        return nameMapping.registerTableOverride(oldTable, newTable);
+        return internalNameMapping.registerTableOverride(oldTable, newTable);
```
Shredder121: Are you sure this isn't supposed to be `nameMapping`?

querydsl-sql/src/main/java/com/querydsl/sql/Configuration.java
```diff
@@ -338,7 +366,7 @@ public SchemaAndTable registerTableOverride(String schema, String oldTable, Stri
      * @return previous override
      */
     public SchemaAndTable registerTableOverride(SchemaAndTable from, SchemaAndTable to) {
-        return nameMapping.registerTableOverride(from, to);
+        return internalNameMapping.registerTableOverride(from, to);
```
Shredder121: Are you sure this isn't supposed to be `nameMapping`?

querydsl-sql/src/main/java/com/querydsl/sql/Configuration.java
```diff
@@ -351,7 +379,7 @@ public SchemaAndTable registerTableOverride(SchemaAndTable from, SchemaAndTable
      * @return previous override
      */
     public String registerColumnOverride(String schema, String table, String oldColumn, String newColumn) {
-        return nameMapping.registerColumnOverride(schema, table, oldColumn, newColumn);
+        return internalNameMapping.registerColumnOverride(schema, table, oldColumn, newColumn);
```
Shredder121: Are you sure this isn't supposed to be `nameMapping`?

querydsl-sql/src/main/java/com/querydsl/sql/Configuration.java
```diff
@@ -363,7 +391,7 @@ public String registerColumnOverride(String schema, String table, String oldColu
      * @return previous override
      */
     public String registerColumnOverride(String table, String oldColumn, String newColumn) {
-        return nameMapping.registerColumnOverride(table, oldColumn, newColumn);
+        return internalNameMapping.registerColumnOverride(table, oldColumn, newColumn);
```
Shredder121: Are you sure this isn't supposed to be `nameMapping`?

querydsl-sql/src/test/java/com/querydsl/sql/ConfigurationTest.java
```diff
@@ -68,9 +74,16 @@ public void get_schema() {
         assertEquals("pub", configuration.getOverride(new SchemaAndTable("public", "")).getSchema());
         assertEquals("emp", configuration.getOverride(new SchemaAndTable("", "employee")).getTable());
         assertEquals("employees", configuration.getOverride(new SchemaAndTable("public", "employee")).getTable());
-//        assertEquals("pub", configuration.getSchema("public"));
-//        assertEquals("emp", configuration.getTable("", "employee"));
-//        assertEquals("employees", configuration.getTable("public", "employee"));
+
+        configuration.setDynamicNameMapping(new ChangeCaseNameMapping(LetterCase.UPPER));
+        String notDirectOverriden = "notDirectOverriden";
+        assertEquals(notDirectOverriden.toUpperCase(),
```
Shredder121: And a `toUpperCase()` call without a given `Locale` fails the Turkey test.

querydsl-sql/src/main/java/com/querydsl/sql/Configuration.java
balazs-zsoldos: The _internalNameMapping_ variable contains the original (current) name mapping functionality of Configuration class. The _nameMapping_ variable contains  the original functionality if no dynamicNameMapping is set otherwise a chain that uses the original functionality and if there is no result, asks for the passed NameMapping implementation. That is why the register functions must write into the internalNameMapping (original behavior) variable. Probably the name of the variables could be slightly different not to be confusing.

querydsl-sql/src/test/java/com/querydsl/sql/ConfigurationTest.java
balazs-zsoldos: Fix: The constructor of ChangeLetterCaseNameMapping accepts a Local parameter now.

Shredder121: I actually meant the test code in this particular comment.
But that is also a good point.

balazs-zsoldos: I see, fixed that. Also found a bug in the ChangeLetterCaseNameMapping class and fixed that, too :).

balazs-zsoldos:
> I'm not sure, but I see that nameMapping is queried, while internalNameMaping is being written to (with registerX.
> This in the tests works, but I think you mean to always query and register on the nameMapping.
> 
> Am I correct?

@Shredder121 _internalNameMapping_ is used for backward compatibility. All _register*_ functions use this variable. The _nameMapping_ variable allows contains

- the same as _internalNameMapping_ if the programmer did not set a dynamic name mapping
- a chained nameMapping that first checks _internalNameMapping_ and than falls back to the nameMapping provided by the programmer

Therefore query is done always on _nameMapping_ while register* functions are applied to _internalNameMapping_. If I were you, I would remove the _register*_ functions and the _internalNameMapping_ from Querydsl 5.0.0 to have a cleaner API.

Shredder121:
> If I were you, I would remove the _register*_ functions and the _internalNameMapping_ from Querydsl 5.0.0 to have a cleaner API.

Yes, that sounds like a good idea.

As for now.
> @Shredder121 _internalNameMapping_ is used for backward compatibility. All _register*_ functions use this variable.

If it works, then I suppose it's fine. I'll have to hear what the other team members think.

Could you add a `TODO` comment to remove the overrides from `Configuration`, and add some more test cases?

balazs-zsoldos:
Sorry for the delay. I was busy preparing to EclipseCon with a talk :). I made the register* functions deprecated if that is good for you. I also added 100% unit test coverage for the new classes in namemapping package.

balazs-zsoldos:
ping :)

Shredder121:
Thanks for pinging.
I don't have anything against it, but no strong feelings for it.
So, looks fine to me, but I'd also like a second opinion on the matter.

balazs-zsoldos:
If I need some discussions about the use-cases why I think this would be useful, please let me know.

However, the exact issue that forced me to create the patch:

 - PostgreSQL quotes column names if _connection.prepareSatement(sql, columnNames[]) are called no matter what. See: https://www.postgresql.org/message-id/CAHEBw=SPKrZz=R7038UmGw9UZAuA7ejm0WA_tmMJ0Uv82-e8mg@mail.gmail.com
 - On the other hand, the same function works with Oracle only if the table and column names are created with upper-case letters. This is true by default, if the database tables and column names are populated without quotes

We could have a solution to create all table names always with quoted upper-case, but we provide [OpenSource solutions](https://github.com/everit-org/jira-querydsl-support) to database of third party apps like Jira. These databases are populated without quotes on object names, that means the object names have lower-case in PostgreSQL and upper-case in Oracle.

We cannot make our code database independent, only if we always use quoting and always with the default letter-case of the database. That is not possible with the current Querydsl version, only if we specify every table and column name within the Configuration instance manually.

I think true database independent code can be written only if the programmer has the possibility to specify rules like: Generate SQL statements with quoted object names with converted to upper-case (or lower-case).

balazs-zsoldos:
ping. I see that there is some work again on the greatest project of java persistence.

I would be very happy if you could review this change and make a decision. Many of our OpenSource modules depend on Querydsl and we need to know if we can make these modules truly database independent with this patch or we need to find another way to solve capital-case issues.

querydsl-sql/src/test/java/com/querydsl/sql/ConfigurationTest.java
```diff
@@ -68,9 +78,51 @@ public void get_schema() {
...
         assertEquals("employees", configuration.getOverride(new SchemaAndTable("public", "employee")).getTable());
-//        assertEquals("pub", configuration.getSchema("public"));
-//        assertEquals("emp", configuration.getTable("", "employee"));
-//        assertEquals("employees", configuration.getTable("public", "employee"));
+
+        configuration.setDynamicNameMapping(new PreConfiguredNameMapping());
+        SchemaAndTable notOverriddenSchemaAndTable = new SchemaAndTable("notoverridden", "notoverridden");
+        assertEquals(notOverriddenSchemaAndTable, configuration.getOverride(notOverriddenSchemaAndTable));
+
+        configuration.setDynamicNameMapping(new ChangeLetterCaseNameMapping(LetterCase.UPPER, Locale.getDefault()));
```
Shredder121: I think using `Locale.getDefault()` is not right here.
then it's the same as just calling `.toLowerCase()`/`.toUpperCase()`.

balazs-zsoldos:
@Shredder121 Fixed

balazs-zsoldos:
ping

balazs-zsoldos:
ping

querydsl-sql/src/main/java/com/querydsl/sql/namemapping/ChainedNameMapping.java
```diff
@@ -0,0 +1,61 @@
+/*
+ * Copyright 2015, The Querydsl Team (http://www.querydsl.com/team)
```
Shredder121: If you can change these years to the current year, I'll merge this.

querydsl-sql/src/main/java/com/querydsl/sql/namemapping/ChangeLetterCaseNameMapping.java
```diff
@@ -0,0 +1,72 @@
+/*
+ * Copyright 2015, The Querydsl Team (http://www.querydsl.com/team)
```
Shredder121: If you can change these years to the current year, I'll merge this.

querydsl-sql/src/main/java/com/querydsl/sql/namemapping/NameMapping.java
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright 2015, The Querydsl Team (http://www.querydsl.com/team)
```
Shredder121: If you can change these years to the current year, I'll merge this.

querydsl-sql/src/main/java/com/querydsl/sql/namemapping/package-info.java
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright 2015, The Querydsl Team (http://www.querydsl.com/team)
```
Shredder121: If you can change these years to the current year, I'll merge this.

Shredder121:
👍 Thanks! Sorry for the incredibly long wait...

balazs-zsoldos:
Shall I delete the branch?

Shredder121:
Yes, the branch isn't needed anymore, the commits are now included.

PR #2019, opened 2016-09-17
jpa doc Customer entity code invalide "[" token

skyupward:
In Jpa document, Customer entity code has wrong token "[" in "setLastName" method. Replacing "[" to "{" make setLastName method become  valid.


Shredder121:
Ah, thanks for spotting that one!
Looks good to me.

Shredder121:
I'll just merge this, thanks again!

PR #2029, opened 2016-10-05
upgrade MongoDB Java Driver to 3.3.0 and Morphia to 1.2.1

cstrempfer:


Shredder121:
Just as an aside, there is some overlap with this pull request with respect to the version bump: #2008

cstrempfer:
@Shredder121 I can update this pull request, when https://github.com/querydsl/querydsl/pull/2008 is merged. 

How are major releases of dependencies are handled normally? Should there be two modules (like the Lucene modules) - one for MongoDB 2 and one for MongoDB 3?

timowest:
@cstrempfer A new module would be the correct way to introduce this support in a minor release

jwgmeligmeyling:
@mp911de  is this still relevant after #2108?

Do we still want to bump these versions? Perhaps to even newer versions?

mp911de:
Aiming with #2689 for 5.0, let's bump driver versions to 4.x. Not sure whether Morphia will still work though.

jwgmeligmeyling:
Closing as current master is already on MongoDB 3.4.0 (newer than 3.3.0) and Morphia 1.3.2 (newer than 1.2.1)

PR #2035, opened 2016-10-10
enable Morphia converters

cstrempfer:
Most of MongoDB queries is built without Morphia, so that converters are not automatically integrated. This pull request adds an explicit call to the mapper.


timowest:
Looks good

Shredder121:
Sorry for the delay, thanks for the PR!

PR #2037, opened 2016-10-11
make NOTE diagnostic printing in APT optional

### LINKED ISSUE #2032: error messages when running gradle plugin
The annotation processor of querydsl gives a number of error messages that should be info/debug messages:

Note: Running JPAAnnotationProcessor
Note: Serializing Embeddable types
Note: Generating ch.adnovum.jcan.jpa.audit.QAuditInfo for [ch.adnovum.jcan.jpa.audit.AuditInfo]

environment: gradle, windows, java 8

reegnz:
Since javac doesn't support hiding only NOTE diagnostic messages,
you have to hide them together with WARN. That's usually not what
you want however, since compiler WARN messages might point out
bugs in your code, so you would want to leave them on.

Solution is to not print NOTE by default and enable it with an
additional compiler arg with -A:
http://docs.oracle.com/javase/7/docs/technotes/tools/windows/javac.html#options

With this change that means you can pass -Aquerydsl.logInfo=true
will
enable NOTE messages, but by default they are not logged.

resolves #2032

### DIFF
--- a/querydsl-apt/src/main/java/com/querydsl/apt/AbstractQuerydslProcessor.java
+++ b/querydsl-apt/src/main/java/com/querydsl/apt/AbstractQuerydslProcessor.java
@@ -13,6 +13,8 @@
  */
 package com.querydsl.apt;
 
+import static com.querydsl.apt.APTOptions.*;
+
 import java.io.IOException;
 import java.io.Writer;
 import java.lang.annotation.Annotation;
@@ -26,7 +28,6 @@ import javax.lang.model.type.DeclaredType;
 import javax.lang.model.type.NoType;
 import javax.lang.model.type.TypeMirror;
 import javax.lang.model.util.ElementFilter;
-import javax.tools.Diagnostic;
 import javax.tools.Diagnostic.Kind;
 import javax.tools.JavaFileObject;
 
@@ -63,16 +64,19 @@ public abstract class AbstractQuerydslProcessor extends AbstractProcessor {
 
     private Context context;
 
+    private boolean shouldLogInfo;
+
     @Override
     public boolean process(Set<? extends TypeElement> annotations, RoundEnvironment roundEnv) {
-        processingEnv.getMessager().printMessage(Diagnostic.Kind.NOTE, "Running " + getClass().getSimpleName());
+        setLogInfo();
+        logInfo("Running " + getClass().getSimpleName());
 
         if (roundEnv.processingOver() || annotations.size() == 0) {
             return ALLOW_OTHER_PROCESSORS_TO_CLAIM_ANNOTATIONS;
         }
 
         if (roundEnv.getRootElements() == null || roundEnv.getRootElements().isEmpty()) {
-            processingEnv.getMessager().printMessage(Diagnostic.Kind.NOTE, "No sources to process");
+            logInfo("No sources to process");
             return ALLOW_OTHER_PROCESSORS_TO_CLAIM_ANNOTATIONS;
         }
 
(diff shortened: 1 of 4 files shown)


querydsl-apt/src/main/java/com/querydsl/apt/AbstractQuerydslProcessor.java
```diff
@@ -539,6 +540,24 @@ public SourceVersion getSupportedSourceVersion() {
...
     }
 
+    @Override
+    public Set<String> getSupportedOptions() {
+        Set<String> optionKeys = new HashSet<String>();
+        optionKeys.add(LOG_PROPERTY_KEY);
+        return optionKeys;
+    }
+
+    private boolean shouldLogInfo() {
```
timowest: put this into a field instead of resolving for each call

reegnz:
Hi, any updates on this?

querydsl-apt/src/main/java/com/querydsl/apt/AbstractQuerydslProcessor.java
```diff
@@ -51,6 +50,8 @@
 
     public static final Boolean ALLOW_OTHER_PROCESSORS_TO_CLAIM_ANNOTATIONS = Boolean.FALSE;
 
+    private static final String LOG_PROPERTY_KEY = "querydsl.apt.log.info";
```
Shredder121: Could you move this to `APTOptions`.

Shredder121:
Looks fine to me, just one comment.

querydsl-apt/src/main/java/com/querydsl/apt/AbstractQuerydslProcessor.java
reegnz: Sure, will do.

reegnz:
Moved property to APTOptions, squashed everything into a single commit.
What about tests? Should I have any? (To be honest, I am not too familiar with testing annotation processors...)

Shredder121:
>What about tests? Should I have any?

Yes, that's an annoying issue.
You could add a new subclass of `AbstractProcessorTest` that replaces the `System.setOut` to a `ByteArrayOutputStream`.
Make sure to override `getAPTOptions()` with the new property you want to test.

Then after `process()` is called, verify the stream is empty.

reegnz:
Added some sanity tests:
without args note logging is disabled
with -Aquerydsl.logInfo=false note logging is disabled
with -Aquerydsl.logInfo=true note logging is enabled

Shredder121:
Looks good to me!
@querydsl/developers What do you all think?

reegnz:
Any update @querydsl/developers?

querydsl-apt/src/test/java/com/querydsl/apt/NoteTest.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+
+import org.junit.Test;
+
+public class NoteTest extends AbstractProcessorTest {
+
+        private Collection<String> aptOptions;
```
timowest: indentation

querydsl-apt/src/test/java/com/querydsl/apt/AbstractProcessorTest.java
```diff
@@ -77,6 +77,14 @@ protected void compile(Class<? extends AbstractProcessor> processorClass, List<S
         }
     }
 
+    protected ByteArrayOutputStream getStdOut() {
+            return new ByteArrayOutputStream();
```
timowest: indentation

reegnz:
Fixed indentation, rebased on latest HEAD of master branch.

reegnz:
I would also look into the checkstyle rule on why it allowed that 8 spaces instead of 4 for indentation on the tests.

Shredder121:
That's.. at least interesting, I know we don't have the same checks for main and test source, but code style should probably be in both. (As far as code generation cooperates with those rules)

PR #2041, opened 2016-10-25
Enable AntMetaDataExporter "namingStrategyClass" and "beanSerializerClass" properties

bmleite:
AntMetaDataExporter "namingStrategyClass" and "beanSerializerClass" properties are being ignored and therefore one cannot change the defaults (DefaultNamingStrategy and BeanSerializer, respectively).


timowest:
Looks good

Shredder121:
Sorry for the delay, thanks for the PR!

PR #2054, opened 2016-11-18
2052: Fixed problem with custom types in ant metadata exporter

danielbalog:
Fixed problem with AntMetaDataExporter not being able to add custom types using build.xml file, or a gradle ant task.

This was already fixed for typeMapping, numericMapping and renameMapping. Now it also works for customTypes.

I also removed the setter/getter as these are not needed for nested elements

Check https://ant.apache.org/manual/develop.html for details on what setters are supported. You can also check it for more details on nested elements.

### DIFF
--- a/querydsl-sql-codegen/src/main/java/com/querydsl/sql/codegen/ant/AntMetaDataExporter.java
+++ b/querydsl-sql-codegen/src/main/java/com/querydsl/sql/codegen/ant/AntMetaDataExporter.java
@@ -176,7 +176,7 @@ public class AntMetaDataExporter extends Task {
     /**
      * custom types to use
      */
-    private String[] customTypes;
+    private List<String> customTypes = Lists.newArrayList();
 
     /**
      * scala generation mode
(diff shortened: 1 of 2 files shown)


danielbalog:
BTW: Example usage in ant task. This works for our project.

```
<queryDSLTask jdbcDriver="${db.driver}"
                  jdbcUrl="${db.url}"
                  jdbcUser="${db.username}"
                  jdbcPassword="${db.password}"
                  packageName="com.data.querydsl.classes"
                  targetFolder="release/gensrc"
                  spatial="true"
                  schemaPattern="internal_schema"
                  >
      <CustomType className="com.querydsl.sql.types.JSR310LocalDateTimeType"></CustomType>
    </queryDSLTask>
```

Shredder121:
:+1:
Thank you for the pull request!
To me this looks good.

Shredder121:
But the enforcer violation is an issue we have to suppress.
Since the original code didn't work.

danielbalog:
Ok I see. I admit that the code I checked in is an incompatible change. If you want however, I can add the getters/setters back with the same arguments, and write adapters for the new internal fields. Should only take a few minutes.

Let me know if this is what you want.

danielbalog:
Ok, so I put the getters/setters back in place for backwards compatibility. The enforcer violation is fixed, and my code still works. I guess this is useful for people who created the task programatically. This will still work for them.

timowest:
Why didn't the original version work?

danielbalog:
It's just not how ant parsing works. I believe the problem originates from how the AntMetaDataParser Task is (most likely) a copy of the Mojo used by maven (AbstractMetaDataExportMojo.java).

But there are actually subtle differences between Ant Tasks and Maven Mojo's in how they parse XML into runtime objects.

I simply updated the AntMetaDataParser to be compatible with Ant xml.

Issue #2074, opened 2017-01-01
Derby DELETE doesn't work with a schema.


gordski:
When using a schema that isn't the default schema, Derby rejects DELETE queries with WHERE clauses that don't reference the schema:

1. `DELETE FROM schema.table WHERE table.column = 'value'   -- Doesn't work`
2. `DELETE FROM schema.table WHERE schema.table.column = 'value' -- Does work`

When configuring QueryDSL with the 'printSchema' option set the Derby templates generate (1).

PR #2075, opened 2017-01-01
Added requiresSchemaInWhere option to SQLTemplates.

gordski:
This option allows templates to generate where clauses with the pattern
`<SCHEMA>.<TABLE>.<COLUMN>` in the column specs.

Derby doesn't appear to support `<TABLE>.<COLUMN>` when using
`<SCHEMA>.<TABLE>` in the from unless the default schema is set and matches
the `<SCHEMA>` in the from clause.

#2074

### DIFF
--- a/querydsl-sql/src/main/java/com/querydsl/sql/SQLSerializer.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/SQLSerializer.java
@@ -471,7 +471,21 @@ public class SQLSerializer extends SerializerBase<SQLSerializer> {
         dmlWithSchema = false;
 
         if (metadata.getWhere() != null) {
-            append(templates.getWhere()).handle(metadata.getWhere());
+            serializeForWhere(metadata);
+        }
+    }
+
+    void serializeForWhere(QueryMetadata metadata) {
+        boolean requireSchemaInWhere = templates.isRequiresSchemaInWhere();
+        boolean originalDmlWithSchema = dmlWithSchema;
+
+        if (requireSchemaInWhere) {
+            dmlWithSchema = true;
+        }
+        append(templates.getWhere()).handle(metadata.getWhere());
+
+        if (requireSchemaInWhere) {
+            dmlWithSchema = originalDmlWithSchema;
         }
     }
 
@@ -634,7 +648,7 @@ public class SQLSerializer extends SerializerBase<SQLSerializer> {
         skipParent = false;
 
         if (metadata.getWhere() != null) {
-            append(templates.getWhere()).handle(metadata.getWhere());
+            serializeForWhere(metadata);
         }
     }
 
--- a/querydsl-sql/src/main/java/com/querydsl/sql/SQLTemplates.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/SQLTemplates.java
@@ -126,6 +126,8 @@ public class SQLTemplates extends Templates {
 
     private final boolean useQuotes;
 
+    private final boolean requiresSchemaInWhere;
+
     private boolean printSchema;
 
     private String createTable = "create table ";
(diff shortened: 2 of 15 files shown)


querydsl-sql/src/main/java/com/querydsl/sql/SQLTemplates.java
```diff
@@ -250,14 +252,15 @@ public SQLTemplates build() {
 
     @Deprecated
     protected SQLTemplates(String quoteStr, char escape, boolean useQuotes) {
-        this(Keywords.DEFAULT, quoteStr, escape, useQuotes);
+        this(Keywords.DEFAULT, quoteStr, escape, useQuotes, false);
     }
 
-    protected SQLTemplates(Set<String> reservedKeywords, String quoteStr, char escape, boolean useQuotes) {
+    protected SQLTemplates(Set<String> reservedKeywords, String quoteStr, char escape, boolean useQuotes, boolean requiresSchemaInWhere) {
```
Shredder121: Add in an additional constructor matching the old one.
For backwards compatibility.

gordski:
I've put in the constructor matching the previous signature. I didn't mark it deprecated, happy to change that if you think it should be.

I've left the non-Derby templates using the longer constructor for clarity.

querydsl-sql/src/main/java/com/querydsl/sql/SQLSerializer.java
```diff
@@ -471,7 +471,19 @@ void serializeForDelete(QueryMetadata metadata, RelationalPath<?> entity) {
...
+
+    void serializeForWhere(QueryMetadata metadata) {
+        boolean requireSchemaInWhere = templates.isRequiresSchemaInWhere();
+        if (requireSchemaInWhere) {
+            dmlWithSchema = true;
+        }
+        append(templates.getWhere()).handle(metadata.getWhere());
+
+        if (requireSchemaInWhere) {
+            dmlWithSchema = false;
```
Shredder121: Would it be safer to just store the old value of `dmlWithSchema`, and restore it here?

gordski: Makes sense. Should I change the other methods to do the same?

`serializeForDelete`, `serializeForMerge`,`serializeForInsert` and `serializeForUpdate`

Shredder121: Ohh I didn't know it was already like that.
I suppose that can be changed in a different PR, which we also can do.

gordski: It looks like that member is only changed in that class. There is a `setDmlWithSchema` but I couldn't find any references.

I'm happy to do the new PR as well if you like.

Shredder121:
I think the only thing we are missing is a test.
Could you add a test, which asserts the `query.toString()` to have the schema in the query?

Somewhere in `SQLSerializerTest` looks fine. Maybe name it `schemaInWhere()`?

querydsl-sql/src/main/java/com/querydsl/sql/SQLSerializer.java
Shredder121: If you could, that would be awesome!

Shredder121:
Just a question, have you already tried `DerbyTemplates.builder().printSchema()`?
As for now it doesn't yet append the schema.

Because during serialization, in a few visitors it has to know whether to print the schema, but it would need to check first in what stage it is.

gordski:
Yes the issue I have seen is using the `printSchema`. You get `DELETE schema.table WHERE table.column = "foo"` which Derby rejects unless `schema` is the default schema.

My change still only appends the schema to the entities in the `WHERE` if configured with `printSchema`

gordski:
@Shredder121 It would be good to get this merged. Is there anything else I need to do?

Shredder121:
Looks good!
@querydsl/developers it can be merged

gordski:
Great thanks!

PR #1523, opened 2015-09-02
Default variable case customization

Shredder121:
This PR is the same as and closes #1458.

PR #2084, opened 2017-01-15
Generating field names as string constants

### LINKED ISSUE #2083: Generating field names as string constants
It would be great if instead

`public final StringPath myField = createString("myField");`

the following would be generated:

`public static final String MY_FIELD = "myField"`
`public final StringPath myField = createString(MY_FIELD);`

It can be used as constant for any client code.

### LINKED ISSUE #2596: Support generating field name constants
It can often be helpful to reference the actual string value of a particular field. This can be done currently by getting the metadata:

```qUser.firstName.getMetadata().getName()```

However this is verbose and makes the code difficult to read if repeated several times. It would be nice if we could optionally specify that the codegen should generate constants for the field names as well. That way we could instead write:

```qUser.FIRST_NAME```

or similar which would be much more concise.

### PR DESCRIPTION
**Pull request by oleg-smith**
Fixes #2083 
Fixes #2314
Fixes #2596

### DIFF
--- a/querydsl-apt/src/main/java/com/querydsl/apt/DefaultConfiguration.java
+++ b/querydsl-apt/src/main/java/com/querydsl/apt/DefaultConfiguration.java
@@ -13,27 +13,62 @@
  */
 package com.querydsl.apt;
 
-import static com.querydsl.apt.APTOptions.*;
-
-import java.lang.annotation.Annotation;
-import java.util.*;
-
-import javax.annotation.Nonnull;
-import javax.annotation.Nullable;
-import javax.annotation.processing.RoundEnvironment;
-import javax.lang.model.element.*;
-import javax.lang.model.type.TypeMirror;
-
 import com.google.common.base.Function;
 import com.google.common.base.Splitter;
 import com.google.common.base.Strings;
 import com.mysema.codegen.model.ClassType;
-import com.querydsl.codegen.*;
+import com.querydsl.codegen.CodegenModule;
+import com.querydsl.codegen.DefaultVariableNameFunction;
+import com.querydsl.codegen.EmbeddableSerializer;
+import com.querydsl.codegen.EntitySerializer;
+import com.querydsl.codegen.EntityType;
+import com.querydsl.codegen.ProjectionSerializer;
+import com.querydsl.codegen.QueryTypeFactory;
+import com.querydsl.codegen.Serializer;
+import com.querydsl.codegen.SerializerConfig;
+import com.querydsl.codegen.SimpleSerializerConfig;
+import com.querydsl.codegen.SupertypeSerializer;
+import com.querydsl.codegen.TypeMappings;
 import com.querydsl.core.annotations.Config;
 import com.querydsl.core.annotations.QueryProjection;
 import com.querydsl.core.annotations.QueryType;
 import com.querydsl.core.types.Expression;
 import com.querydsl.core.util.Annotations;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import javax.annotation.processing.RoundEnvironment;
+import javax.lang.model.element.Element;
+import javax.lang.model.element.ExecutableElement;
+import javax.lang.model.element.Modifier;
+import javax.lang.model.element.PackageElement;
+import javax.lang.model.element.TypeElement;
+import javax.lang.model.element.VariableElement;
+import javax.lang.model.type.TypeMirror;
+import java.lang.annotation.Annotation;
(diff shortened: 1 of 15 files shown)

### CODE REVIEW

**Comment by Shredder121** [comment]
Keep in mind not all column names are lower camelcase'd.

Also, there are times where column name and Java identifier rules differ.
Java identifiers cannot start with a number, or contain spaces.

**Comment by oleg-smith** [comment]
Fixed considering Java identifiers naming.
I'm assuming that if field name is not in lower Camel case, then it will be translated as upper case, which is also suitable for this case.

File: querydsl-codegen/src/main/java/com/querydsl/codegen/EntitySerializer.java [hunk 1]
```diff
@@ -620,6 +663,10 @@ public void serialize(EntityType model, SerializerConfig config,
 
     protected void serialize(EntityType model, Property field, Type type, CodeWriter writer,
             String factoryMethod, String... args) throws IOException {
+
+        String constantFieldName = constantFieldName(field.getEscapedName());
+        writer.publicStaticFinal(new ClassType(String.class), constantFieldName, "\"" + field.getName() + "\"");
```
timowest: use Types.String instead of new ClassType(String.class)
also could these additions be toggable with an option?

oleg-smith: @timowest sure, how would you prefer me passing the option? A code example can help.

oleg-smith: @timowest ^^

ykrasnouski: @oleg-smith,
Probably option to toggle this feature could be added as @Named constructor param like keywords param in EntitySerializer constructor.
`@Named("keywords") Collection<String> keywords`

Then default value for it could be binded in CodegenModule.java:69
`bind(KEYWORDS, Collections.<String>emptySet());`

 And in DefaultConfiguration.java you could rebind it to a value gotten from options map.
```
if (options.containsKey(QUERYDSL_PACKAGE_SUFFIX)) {
            module.bind(CodegenModule.PACKAGE_SUFFIX, Strings.nullToEmpty(options.get(QUERYDSL_PACKAGE_SUFFIX)));
        }
```

Shredder121: Sorry for the radio silence.
Yes, that sounds good.
You can kind of see what places you need to update in [this PR](https://github.com/querydsl/querydsl/pull/1523/files).

oleg-smith:
@timowest @Shredder121 @ykrasnouski could you review?

oleg-smith:
^^

querydsl-apt/src/main/java/com/querydsl/apt/DefaultConfiguration.java
```diff
@@ -205,6 +240,12 @@ public DefaultConfiguration(
         }
         module.bind(CodegenModule.VARIABLE_NAME_FUNCTION_CLASS, variableNameFunction);
 
+        boolean namesAsConstants = options.containsKey(QUERYDSL_NAMES_AS_CONSTANTS)
```
timowest: preferably use the same pattern as elsewhere
```
if (options.containsKey(QUERYDSL_NAMES_AS_CONSTANTS)) {
    module.bind(CodegenModule.NAMES_AS_CONSTANTS, Boolean.valueOf(options.get(QUERYDSL_NAMES_AS_CONSTANTS)))
}
```

querydsl-apt/src/main/java/com/querydsl/apt/DefaultConfiguration.java
```diff
@@ -13,28 +13,63 @@
...
-import javax.annotation.processing.RoundEnvironment;
-import javax.lang.model.element.*;
-import javax.lang.model.type.TypeMirror;
-
 import com.google.common.base.Function;
 import com.google.common.base.Splitter;
 import com.google.common.base.Strings;
 import com.mysema.codegen.model.ClassType;
-import com.querydsl.codegen.*;
+import com.querydsl.codegen.CodegenModule;
```
timowest: why all these package refactorings?
still to the necessary changes

oleg-smith: Can we merge it as is? As it's been a while without update.

jwgmeligmeyling (changes requested):
PR needs to be updated with current master.

PR #2093, opened 2017-02-08
Adds methods to SQLListenerContext to retrieve SQLBindings.

### LINKED ISSUE #1839: Extend SQLListenerContext
I'm trying to integrate SqlDetailedListeners into our existing SQL logging framework.  We log the type of query, and include the SQL parameters and ResultSet info if there is any.  None of this is saved in the SQLListenerContext so it's hard to have a unified logging framework,

I will submit a Pull Review with suggested changes.

karahatherly:
This was added as an alternative to just getting an SQL string. Internally,
SQLListenerContextImpl now only takes SQLBindings and the getSQL and
getSQLStatements methods delegate to SQLBindings#getSQL. This required
shuffling around AbstractSQLQuery and SQL*Clause to create bindings sooner to
pass through to addSQL even though we derive the same data later on inside
AbstractSQLClause#setParameters - I figured this was preferable to larger-scale
refactoring.

Fixes #1839

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -281,6 +281,7 @@
                   <excludes>
                     <exclude>com/querydsl/core/types/dsl/PathBuilderValidator</exclude>
                     <exclude>com/querydsl/sql/Union</exclude>
+                    <exclude>com/querydsl/sql/SQLListenerContextImpl</exclude>
                   </excludes>
                   <compatibilityType>BACKWARD_COMPATIBLE_USER</compatibilityType>
                   <dumpDetails>true</dumpDetails>
--- a/querydsl-sql/src/main/java/com/querydsl/sql/AbstractSQLQuery.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/AbstractSQLQuery.java
@@ -277,7 +277,7 @@ public abstract class AbstractSQLQuery<T, Q extends AbstractSQLQuery<T, Q>> exte
             SQLSerializer serializer = serialize(false);
             queryString = serializer.toString();
             logQuery(queryString, serializer.getConstants());
-            context.addSQL(queryString);
+            context.addSQL(getSQL(serializer));
             listeners.rendered(context);
 
             listeners.notifyQuery(queryMixin.getMetadata());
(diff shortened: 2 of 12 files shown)


karahatherly:
The failing tests are interesting - as far as I can tell I've surfaced an existing bug rather than created a new one. `null` seems to be an acceptable constant to pass through to (for example) an `in` clause. However, after doing so calling `SQLQuery#getSQL` will NPE on trying to put `null` into an `ImmutableList.Builder`. Adding the following test (modelled off of SelectBase#in_null) surfaces this bug on master:
    
    @Test
    public void in_null_npe_fun() {
        assertEquals(2, query().from(employee).where(employee.id.in(1, null)).getSQL().getBindings().size());
    }

I've tested out changing `SQLBindings` to not use an `ImmutableList` (so that `null` becomes an acceptable value) and this works, but doing so would break binary compatibility. I'm not really sure how to proceed - any input would be appreciated.

karahatherly:
Also: a bunch of tests fail if the timezone is Sydney. I've got a change locally to explicitly set UTC in maven config to save others getting burnt by this. Would you prefer a separate PR for that? Or is it cool if I piggy back it onto this one?

timowest:
@karahatherly Thanks for the contribution, could you make a separate PR for the timezone fix?

timowest:
Looks good

karahatherly:
@timowest thanks, I'll put up another PR setting an explicit timezone later today.

Any thoughts on [the remaining issue with `SQLBindings`](https://github.com/querydsl/querydsl/pull/2093#issuecomment-278545269)? I was thinking the least intrusive approach might just be to add a new method which just returns a List, deprecate the one returning an ImmutableList and we can remove it in the next major release. Thoughts?

karahatherly:
I've gone ahead with the changes I proposed to `SQLBindings` - lmk if they look alright.

karahatherly:
@timowest mind taking a look at the last commit and merging if you're happy with how I fixed the issue with `SQLBindings`?

timowest:
Looks good

PR #2099, opened 2017-02-16
Explicitly sets a timezone for querydsl-sql tests.

### LINKED ISSUE #2005: Subquery alias does not work in querydsl 4
After upgrading from querydsl 3.x to 4.1.2 path aliases do not wok as they used to.

Working example:

```
new JPAQuery<Tuple>(em)
    .select(QPerson.person.id.as("idAlias"))
    .from(QPerson.person)
    .orderBy(new OrderSpecifier<>(DESC, stringPath("idAlias")))
    .fetch();
```

Broke (worked in 3.x):

```
new JPAQuery<Tuple>(em)
    .select(QPerson.person.id.as("idAlias"),
        JPAExpressions.select(count.as(numberPath(Long.class, "childrenCount")))
        .from(QChildren.children)
        .where(QChildren.children.father.eq(QPerson.person))
)
    .from(QPerson.person)
    .orderBy(new OrderSpecifier<>(DESC, stringPath("childrenCount")))
    .fetch();
```

Stacktrace reads:

```
javax.persistence.PersistenceException: org.hibernate.exception.SQLGrammarException: could not prepare statement
    at org.hibernate.jpa.spi.AbstractEntityManagerImpl.convert(AbstractEntityManagerImpl.java:1692)
    at org.hibernate.jpa.spi.AbstractEntityManagerImpl.convert(AbstractEntityManagerImpl.java:1602)
    at org.hibernate.jpa.internal.QueryImpl.getResultList(QueryImpl.java:492)
    at com.querydsl.jpa.impl.AbstractJPAQuery.getResultList(AbstractJPAQuery.java:160)
    at com.querydsl.jpa.impl.AbstractJPAQuery.fetch(AbstractJPAQuery.java:202)
```

While still working example constructs SQL which ends `order by col_0_0_ desc` (generated column name) in broken example query ends with `order by childrenCount desc` but the entire query does not have such alias.

karahatherly:
Fixes #2005

### DIFF
--- a/querydsl-sql/pom.xml
+++ b/querydsl-sql/pom.xml
@@ -259,6 +259,7 @@
                 <value>target/derby.log</value>
             </property>
           </systemProperties>
+          <argLine>-Duser.timezone=UTC</argLine>
         </configuration>
       </plugin>         
       


querydsl-sql/pom.xml
```diff
@@ -259,6 +259,7 @@
                 <value>target/derby.log</value>
             </property>
           </systemProperties>
+          <argLine>-Duser.timezone=UTC</argLine>
```
karahatherly: Apparently in java 8 setting `user.timezone` in `systemProperties` results in it getting read too late and being cached as the system default. Using `argLine` instead works around that.

Shredder121: Huh. okay. TIL.

Shredder121:
Question, would you also be able to submit a PR to QUERYDSL_3? 😄

Shredder121:
Also, this should "fix #2098" I think.

querydsl-sql/pom.xml
Shredder121: Just thought of something.

The JaCoCo plugin needs to update the `argLine` before it's passed in.
Could you add the `argLine` property in the `<properties />` section?
It can look like this:
```xml
<properties>
  ...
  <argLine>${argLine} -Duser.timezone=UTC</argLine>
  ...
</properties>
```

karahatherly:
Recreated as https://github.com/querydsl/querydsl/pull/2104, thanks to @Shredder121 for pointing out my issue number typo

PR #2101, opened 2017-02-16
Add javax.inject:javax.inject:1:jar as runtime dependency

### LINKED ISSUE #2026: Runtime error due to POM - javax.inject [provided]
Hello,

I am configuring a SpringBoot application with Maven and QueryDSL for Hibernate.
Everything compiles and deploys correctly, the server starts, I starts a request, it opens the transaction...
But as soon as I do this :

`qd = new HibernateQueryFactory(sessionFactory.getCurrentSession());`

I get a runtime error :

```
java.lang.NoClassDefFoundError: javax/inject/Provider
    at com.asinter.agileqms.dao.MyDaoImpl.getHibernateQueryFactory(MyDaoImpl.java:22) ~[classes/:3.1.0-SNAPSHOT]
    at com.asinter.agileqms.dao.MyDaoImpl.getServiceType(MyDaoImpl.java:28) ~[classes/:3.1.0-SNAPSHOT]
    at com.asinter.agileqms.dao.MyDaoImpl$$FastClassBySpringCGLIB$$9f8a6888.invoke(<generated>) ~[classes/:3.1.0-SNAPSHOT]
    at org.springframework.cglib.proxy.MethodProxy.invoke(MethodProxy.java:204) ~[spring-core-4.3.3.RELEASE.jar:4.3.3.RELEASE]
    at org.springframework.aop.framework.CglibAopProxy$CglibMethodInvocation.invokeJoinpoint(CglibAopProxy.java:720) ~[spring-aop-4.3.3.RELEASE.jar:4.3.3.RELEASE]

Caused by: java.lang.ClassNotFoundException: javax.inject.Provider
    at org.apache.catalina.loader.WebappClassLoaderBase.loadClass(WebappClassLoaderBase.java:1285) ~[catalina.jar:8.5.5]
```

In the POM.XML, javax.inject is in the scope [provided] of querydsl-apt, as told in the documentation.
If I remove the scope, or if I add GUICE as a dependancy, it fixes the issue.

I'm using Tomcat 8.5

grimly:
Since the javax.inject.Provider interface is imported and used in every constructor of the *QueryFactory classes ( example at https://github.com/querydsl/querydsl/blob/master/querydsl-jpa/src/main/java/com/querydsl/jpa/impl/JPAQueryFactory.java#L17 ), it should be a runtime dependency and not only provided.

Fixes #2026

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -136,6 +136,11 @@
         <version>${slf4j.version}</version>
         <scope>provided</scope>
       </dependency>
+      <dependency>
+        <groupId>javax.inject</groupId>
+        <artifactId>javax.inject</artifactId>
+        <version>1</version>
+      </dependency>
     </dependencies>
   </dependencyManagement>
 
--- a/querydsl-codegen/pom.xml
+++ b/querydsl-codegen/pom.xml
@@ -37,7 +37,6 @@
     <dependency>  
       <groupId>javax.inject</groupId>  
       <artifactId>javax.inject</artifactId>  
-      <version>1</version>  
     </dependency>
 
     <dependency>
--- a/querydsl-jpa/pom.xml
+++ b/querydsl-jpa/pom.xml
@@ -95,7 +95,6 @@
     <dependency>
       <groupId>javax.inject</groupId>
       <artifactId>javax.inject</artifactId>
-      <version>1</version>
     </dependency>
     
     <dependency>
--- a/querydsl-jpa/pom.xml
+++ b/querydsl-jpa/pom.xml
@@ -92,6 +92,11 @@
       <version>${project.version}</version>
       <scope>provided</scope>
     </dependency>
+    <dependency>
+      <groupId>javax.inject</groupId>
+      <artifactId>javax.inject</artifactId>
+      <version>1</version>
+    </dependency>
     
     <dependency>
       <groupId>com.querydsl</groupId>


Shredder121:
Where do we declare it as `provided`?
Of course, marking it as dependency might be clearer, but I suppose that we could add it to the root `pom.xml` then and inherit the declared version.

grimly:
The "provided" dependency comes from the [querydsl-apt artifact](https://github.com/querydsl/querydsl/blob/master/querydsl-jpa/pom.xml#L89) which depends on the [querydsl-codegen artifact](https://github.com/querydsl/querydsl/blob/master/querydsl-apt/pom.xml#L26) which (at last) depends on [javax.inject](https://github.com/querydsl/querydsl/blob/master/querydsl-codegen/pom.xml#L37)

I don't know if it must be included into the root pom.
Some projects such as the lucene artifacts doesn't require this dependency.
This is a political choice at this point and I'm not a such serious contributor to have my word on it.

Shredder121:
I meant to have them in a `dependencyManagement` section, so we don't need to update versions left and right.
But that's probably better to do it in one go.
Looks good though. 👍

grimly:
Indeed this is a better way to do it.
I missed the dependency management part of the root pom since [it's really small](https://github.com/querydsl/querydsl/blob/master/pom.xml#L126) and I thought there was none.

I didn't went through the whole project so I might have missed other occurrences of javax.inject declarations to removed their version.

PR #2103, opened 2017-02-20
Store original state of dmlWithSchema

gordski:
This is a follow on PR from the discussion in #2075.

It protects the original state of the dmlWithSchema flag rather than explicitly setting it to `false`.


Shredder121:
Looks good, thanks!

timowest:
Looks good

PR #2104, opened 2017-02-21
Explicitly sets a timezone for querydsl-sql tests.

### LINKED ISSUE #2098: Tests fail when running in Australia/Sydney (& presumably others). We should set an explicit timezone in maven config.

karahatherly:
Fixes #2098 

Recreation of https://github.com/querydsl/querydsl/pull/2099 because I messed up on issue numbers.

### DIFF
--- a/querydsl-sql/pom.xml
+++ b/querydsl-sql/pom.xml
@@ -21,7 +21,7 @@
       org.joda.time.*;version="[1.6,3)",
       ${osgi.import.package.root}
     </osgi.import.package>
-
+    <argLine>-Xms256m -Xmx512m -Duser.timezone=UTC</argLine>
   </properties>
 
   <dependencies>


Shredder121:
Ah, I see you opened a new PR.
You can also edit the PR's description when you're the author.

But that's for the next time then.

querydsl-sql/pom.xml
```diff
@@ -259,6 +259,7 @@
                 <value>target/derby.log</value>
             </property>
           </systemProperties>
+          <argLine>${surefire.argLine} -Duser.timezone=UTC</argLine>
```
Shredder121: You can remove this line entirely.

querydsl-sql/pom.xml
```diff
@@ -21,7 +21,7 @@
       org.joda.time.*;version="[1.6,3)",
       ${osgi.import.package.root}
     </osgi.import.package>
-
+    <surefire.argLine/>
```
Shredder121: This line should be:
```xml
<argLine>${argLine} -Duser.timezone=UTC</argLine>
```

Shredder121: Hmm, I see that Maven complains about a recursive definition, I'll have to think about a solution.

karahatherly:
@Shredder121 unfortunately I couldn't find a way to change the source branch (which contained the wrong issue number) for the PR in the github UI - figured recreating was harmless enough.

Shredder121:
Yes, creating another PR is fine, but you can also push to the branch and the changes will reflect on here.
It's your fork and branch, so force pushing is also an option.

Shredder121:
Looks good, sorry I didn't notice it was changed.

timowest:
Looks good

Shredder121:
👍 Now we just need to wait for some docker magic from @johnktims.
👀 
And then we have working tests again.

PR #2149, opened 2017-05-15
Add SQL symbolic type to java type mapping for maven plugin

wwalery:
Example, map Oracle type NVARCHAR2, pom.xml:
```java
<sqlTypeMappings>
  <sqlTypeMapping>
    <sqlType>nvarchar2</sqlType>
    <javaType>java.lang.String</javaType>
  </sqlTypeMapping>
<sqlTypeMapping>
```

### DIFF
--- a/querydsl-sql-codegen/src/main/java/com/querydsl/sql/codegen/support/SQLTypeMapping.java
+++ b/querydsl-sql-codegen/src/main/java/com/querydsl/sql/codegen/support/SQLTypeMapping.java
@@ -1,0 +1,52 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.querydsl.sql.codegen.support;
+
+import com.querydsl.sql.Configuration;
+
+/**
+ * {@code SQLTypeMapping} customizes mappings SQL symbolic type to data types.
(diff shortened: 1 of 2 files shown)


Shredder121:
Could you elaborate on why this would be necessary, when there's already a more fine-grained [`TypeMapping`](https://github.com/querydsl/querydsl/blob/master/querydsl-sql-codegen/src/main/java/com/querydsl/sql/codegen/support/TypeMapping.java)?
It's just to get a feel of why the change is necessary, and if necessary how we can make this change fit a little better.

wwalery:
For TypeMapping we need to use map  with three arguments Table/Column/Java type. 
For SQLTypeMapping we need only SQL type name. 
Thus for mapping SQL type to java we need only one XML block in the pom.xml. 
For example, in our current project (Oracle DB) we have more than 100 tables with one or more fields of type NVARCHAR2. But with TypeMapping it's necessary to create XML block for each field of this type
Also it's possible to use CustomType and for a first time I tried to use it, but...  it requires an additional project (subproject/module etc) for implementing successor of Type.
Another alternative is to make Table/Column as optional for TypeMapping, but IMHO it needs more changes in the code

querydsl-sql-codegen/src/main/java/com/querydsl/sql/codegen/support/SQLTypeMapping.java
```diff
@@ -0,0 +1,52 @@
...
+
+import com.querydsl.sql.Configuration;
+
+/**
+ * {@code SQLTypeMapping} customizes mappings SQL symbolic type to data types.
+ *
+ * @author Walery Wysotsky
+ *
+ */
+public class SQLTypeMapping implements Mapping {
```
timowest: should this be TypeNameMapping instead? the SQL prefix feels redundant

wwalery: "SQLTypeMapping" is used because it's mapping from SQL field types to Java. On my mind "TypeNameMapping" is more global and not describe what exactly we map - java to SQL or SQL to java... But if you need to change class name I will not mind

PR #2157, opened 2017-06-16
Fix Hibernate 5.2+ incompatibility

### LINKED ISSUE #1917: Hibernate 5.2 method signature incompatibilty
I tried out the latest Query DSL 4.1.2 and Hibernate 5.2, and hibernate appears to have changed method signatures.

```
java.lang.NoSuchMethodError: org.hibernate.Session.createQuery(Ljava/lang/String;)Lorg/hibernate/Query;
    at com.querydsl.jpa.hibernate.DefaultSessionHolder.createQuery(DefaultSessionHolder.java:36)
    at com.querydsl.jpa.hibernate.AbstractHibernateQuery.createQuery(AbstractHibernateQuery.java:104)
    at com.querydsl.jpa.hibernate.AbstractHibernateQuery.createQuery(AbstractHibernateQuery.java:97)
    at com.querydsl.jpa.hibernate.AbstractHibernateQuery.fetch(AbstractHibernateQuery.java:174)
```

The `Query` class moved from `org.hibernate` to `org.hibernate.query` so the APIs don't match anymore.

I'm sure there are more broken areas as well.

Shredder121:
Closes #1917

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -53,8 +53,8 @@
     <firebird.version>2.2.5</firebird.version>
 
     <!-- JPA deps -->
-    <hibernate.version>4.3.11.Final</hibernate.version>
-    <hibernate.validator.version>4.3.1.Final</hibernate.validator.version>
+    <hibernate.version>5.2.12.Final</hibernate.version>
+    <hibernate.validator.version>5.3.1.Final</hibernate.validator.version>
     <eclipselink.version>2.6.2</eclipselink.version>
 
     <guava.version>18.0</guava.version>
--- a/querydsl-jpa/src/test/java/com/querydsl/jpa/support/TeradataDialect.java
+++ b/querydsl-jpa/src/test/java/com/querydsl/jpa/support/TeradataDialect.java
@@ -41,6 +41,7 @@ import org.hibernate.type.StandardBasicTypes;
  *
  * @author Jay Nance
  */
+@Deprecated // what was this used for? @timowest?
 public class TeradataDialect extends Dialect {
 
     /**
@@ -139,11 +140,6 @@ public class TeradataDialect extends Dialect {
     @Override
     public String getForUpdateString() {
         return "";
-    }
-
-    @Override
-    public boolean supportsIdentityColumns() {
-        return false;
     }
 
     @Override


Shredder121:
Target is master branch, so 4.x, but backport seems feasible.

querydsl-jpa/src/main/java/com/querydsl/jpa/hibernate/DefaultSessionHolder.java
```diff
@@ -33,12 +35,24 @@ public DefaultSessionHolder(Session session) {
 
     @Override
     public Query createQuery(String queryString) {
-        return session.createQuery(queryString);
+        try {
+            return MethodHack.createQuery.invoke(session, queryString);
```
timowest: forgive my ignorance, but how does this solve the issue? Isn't there still the org.hibernate.Query vs org.hibernate.query.Query issue in the source code?

timdvail: As I understand it -- by using reflection to obtain and invoke the method, the exact method signature never gets compiled into the querydsl bytecode.  This allows it to call either hibernate 5.1 (and earlier) or 5.2 version of the method where the signature changed slightly from returning org.hibernate.Query to org.hibernate.query.Query.  Yes, the Query class moved, but they still have org.hibernate.Query in place (and deprecated) with the latter subclassing the former so some degree of polymorphism is still maintained and old source would still compile -- but once compiled it expects the compiled-in signature.

querydsl-jpa/src/main/java/com/querydsl/jpa/hibernate/SessionHolder.java
```diff
@@ -24,6 +27,30 @@
  */
 public interface SessionHolder {
 
+    /**
+     * Workaround for Hibernate 5.2's return type changes.
+     *
+     * Because we can't have the return type in source code form, we have to use reflection.
+     * This means that the method lookup happens just by name.
```
Shredder121: @timowest Because the method `org.hibernate.Query.createQuery(Ljava/lang/String)Lorg/hibernate/SQLQuery` (and others) doesn't exist anymore, we have to invoke it by name.
Although on closer inspection I see that I missed a bunch more.
```
java.lang.NoSuchMethodError: org.hibernate.SQLQuery.addScalar(Ljava/lang/String;)Lorg/hibernate/SQLQuery;
java.lang.NoSuchMethodError: org.hibernate.Session.close()Ljava/sql/Connection;
And probably others
```
I now notice what kind of mess they made, I think the only way is to either ditch pre 5.2 or to figure out some additional maven artifact for pre 5.2.

kartoffelsup: +1 for a new release with 5.2 support (maybe even QDSL5?) :)

Shredder121:
@timowest There are more methods that would need supporting, which is too hard to test.
Do you want to drop support for <5.2? Do you rather want an artifact with a dependency on the old version we were supporting?
This is a minor release that's coming up, so it's sort-of acceptable. Maybe a little fast but I'd also rather have wanted Hibernate didn't do it.

timowest:
I am ok with dropping support for older HIbernate versions.

Shredder121:
I think doing it like @olivergierke suggested is better, however there haven't been any plans yet for Querydsl 5.
But I suppose we'll have to now. 👍

johnktims:
Querydsl is now compatible with hibernate 5.

PR #2172, opened 2017-07-25
Fixed NEXTVAL Operator for MS SQLServer (starting with 2012)

tpuica:
Hi,
The nextval operator is wrong for SQL Server. It should be "next value for ". 
See https://docs.microsoft.com/en-us/sql/t-sql/functions/next-value-for-transact-sql

Thanks.
theodor

### DIFF
--- a/querydsl-sql/src/main/java/com/querydsl/sql/SQLServer2012Templates.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/SQLServer2012Templates.java
@@ -64,6 +64,7 @@ public class SQLServer2012Templates extends SQLServerTemplates {
 
     protected SQLServer2012Templates(Set<String> keywords, char escape, boolean quote) {
         super(keywords, escape, quote);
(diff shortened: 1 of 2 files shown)


Shredder121:
So would the correct syntax be `next value for myseq`?
If that's correct, could you update that?

https://github.com/querydsl/querydsl/blob/master/querydsl-sql/src/test/java/com/querydsl/sql/SQLServer2012TemplatesTest.java#L106-L110

Shredder121:
Could you also link the relevant documentation for completeness sake?

Shredder121:
Also, I think something went wrong on our end, if you could also change it to be the more concise `assertSerialized`?
```java
assertSerialized(nextval, "next value for myseq");
```

tpuica:
Test updated. 
It is though unclear for me where or how to link the relevant documentation (the link to the Microsoft documentation was in the pull request comment from the beginning)

Shredder121:
Ah, then disregard that, sorry.

timowest:
Looks good

PR #2177, opened 2017-08-04
Show property names list when QueryInit path fails

nealeu:
Having scratched my head a fair bit around QueryInit over time, I thought I'd check out the code to see what I can and cannot specify as a QueryInit path.

This patch just makes it much clearer as to what QueryDSL can see (e.g. that it can see super class properties, but without need for _super - which was my guesswork).


nealeu:
@timowest Being on 3.x I realise you'd poss have wanted this on master and then cherry-pick to maint branch.  Happy to redo pull request if needed that way, but I would guess it works either way at your end.

Shredder121:
Either way works, no problem.

Shredder121:
:+1: I think it looks fine, thanks for the contribution.

Now to see if there is someone else to merge it.

Could you also make a PR targeting master?

PR #2178, opened 2017-08-06
Show property names list when QueryInit path fails

nealeu:
Version of PR #2177 for master branch


timowest:
Looks good

nealeu:
That should fix the checkstyle issue.

PR #2183, opened 2017-08-13
Adding KEY as a reserved word for QueryDSL 3.x

jerometerry:
MS SQL Server rejects queries for columns named key without being escaped. [MS SQL Reserved Keywords](https://docs.microsoft.com/en-us/sql/t-sql/language-elements/reserved-keywords-transact-sql)

Without KEY being part of the default reserved words, I need to override the MS SQL Templates and override `requiresQuotes`. 

An alternative design would be to allow addition of keywords via configuration, but that is needless complexity IMHO. KEY is in the default list of keywords in QueryDSL 4.

### DIFF
--- a/querydsl-sql/src/main/java/com/mysema/query/sql/SQLTemplates.java
+++ b/querydsl-sql/src/main/java/com/mysema/query/sql/SQLTemplates.java
@@ -71,7 +71,7 @@ public class SQLTemplates extends Templates {
                     "GLOBAL", "GRANT", "GROUP", "GROUPING", "HAVING", "HOLD",
                     "HOUR", "IDENTITY", "IN", "INDICATOR", "INNER", "INOUT",
                     "INSENSITIVE", "INSERT", "INT", "INTEGER", "INTERSECT",
-                    "INTERSECTION", "INTERVAL", "INTO", "IS", "JOIN", "LAG",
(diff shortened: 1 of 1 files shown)


Shredder121:
You can also put the list of keywords, specific to SQL Server in the SQLServerTemplates, and use that to call the super constructor.

jerometerry:
I couldn't access a constructor to pass a list of keywords. Only solution I could come up with was this

```
public class SQLServerAdditionalKeywords
{
    private static final Set<String> additionalReservedKeywords = Sets.newHashSet("KEY");

    public static boolean isReservedWord(String identifier)
    {
        return additionalReservedKeywords.contains(identifier.toUpperCase());
    }
}

public class SS2012 extends SQLServer2012Templates
{
    @Override
    protected boolean requiresQuotes(final String identifier, final boolean precededByDot)
    {
        if (SQLServerAdditionalKeywords.isReservedWord(identifier))
        {
            return true;
        }
        else
        {
            return super.requiresQuotes(identifier, precededByDot);
        }
    }
}
```

jerometerry:
@Shredder121 Are you recommending that I update this PR to add additional keywords required by SQL Server to SQLServerTemplates, instead of updating SQLTemplates?

PR #2184, opened 2017-08-13
Adding KEY as a reserved word for QueryDSL 3.x

jerometerry:
MS SQL Server rejects queries for columns named key without being escaped. [MS SQL Reserved Keywords](https://docs.microsoft.com/en-us/sql/t-sql/language-elements/reserved-keywords-transact-sql)

This PR is an alternative to [PR 2183](https://github.com/querydsl/querydsl/pull/2183). Instead of adding "KEY" to SQLTemplates.SQL_RESERVED_WORDS, construct an ImmutableSet in SQLServerTemplates consisting of the union of SQLTemplates.SQL_RESERVED_WORDS and MS SQL specific keywords.

### DIFF
--- a/querydsl-sql/src/main/java/com/mysema/query/sql/SQLServerTemplates.java
+++ b/querydsl-sql/src/main/java/com/mysema/query/sql/SQLServerTemplates.java
@@ -14,7 +14,9 @@
 package com.mysema.query.sql;
 
 import java.sql.Types;
+import java.util.Set;
 
+import com.google.common.collect.ImmutableSet;
 import com.mysema.query.QueryFlag;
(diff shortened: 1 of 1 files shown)


jerometerry:
@Shredder121 What are your thoughts on getting this into the 3.x release?

Shredder121:
Shouldn't be too hard.
I suppose this will be the time I'll look in the correctness of my release profile.

johnktims:
We haven't supported querydsl 3 for a while now. I'm going to close this PR since it's probably irrelevant at this point, but if I'm wrong, comment and we'll try to release this fix.

PR #2193, opened 2017-08-29
Upgrade mongodb-driver and morphia

ikonglong:
Upgrade java mongodb driver from v2.10.0 to v3.5.0;
Upgrade morphia from v0.105 to v1.3.2.


timowest:
Looks good, but this dependency bump should be probably applied on a major or minor version of Querydsl.

Shredder121:
Luckily the next version is going to be a minor version bump.

PR #2201, opened 2017-09-08
Fix missing scope for jda dependency in querydsl-apt

### LINKED ISSUE #2138: jdo-api dependency lost its scope in 4.1.4
querydsl-apt 4.1.4 has a compile-time dependency on javax.jdo:jdo-api. This dependency should be (used to be) provided scope but the scope was lost in commit https://github.com/querydsl/querydsl/commit/9d00a069aa3a0ea8c24a171aa9c1582ded23ddad

Shredder121:
Fixes #2138


timowest:
Looks good

PR #2218, opened 2017-11-05
allow eviction of DEFAULT template cache - this leaks memory, JVM get OOM fast if the expressions are big.

s50600822:
TemplateFactory.cache leaks memory, JVM get OOM fast if the expressions are big.
Provide methods to clearCache optionally.

### DIFF
--- a/querydsl-core/src/main/java/com/querydsl/core/types/TemplateFactory.java
+++ b/querydsl-core/src/main/java/com/querydsl/core/types/TemplateFactory.java
@@ -16,11 +16,15 @@ package com.querydsl.core.types;
 import java.math.BigDecimal;
 import java.util.Locale;
 import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import com.google.common.base.Function;
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.querydsl.core.types.Template.Element;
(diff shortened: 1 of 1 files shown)


Shredder121:
:+1:
I'm not opposed to the idea.
However could you explain a little on how the cache could keep growing?
Maybe there's something else we also need to address first?

karahatherly:
> However could you explain a little on how the cache could keep growing?

`TemplateFactory.cache` is a `ConcurrentHashMap` so its size is bound only by the number of cache misses. On a project which uses querydsl like Jira does this can result in OOMEs as @s50600822 states. Exposing methods to clear it allows us to

> Maybe there's something else we also need to address first?

I think additionally we should change the `ConcurrentHashMap` to a guava cache and configure it at run-time based on a `CacheBuilderSpec` representation in a property, to allow querydsl consumers to tune that cache if necessary. We'd then still want to expose the `invalidate*` and `cleanUp` methods from `Cache`, similar to what's done in this PR for the `Map`.

timowest:
I'd also like to see a use case description first, how this cache could grow too big. There should be finite amount of operation serialization templates in a project.

s50600822:
It was just a matter of scale. Apparently there is a platform that heavily relies on user inputs and context to build queries to database. 
Adopting serverless architecture of that platform forces the JVMs to serve request from all different customers with different contexts since there's no sticky session.
(that wasn't my idea btw :< ).

Shredder121:
Could you elaborate on the scale? Is it arbitrary amounts of applications in one JVM?

dodolzg:
I have the same problem, java.lang.OutOfMemoryError.
JVM shows that there are lots of char[], string which refrence sql template, however it can be reached by GC ROOT.

s50600822:
> Could you elaborate on the scale? Is it arbitrary amounts of applications in one JVM?

put it this way if you can't think of how big the mem leak is, how about a limit it will never use up says 20MB or something ? The code that generate that many query String into the cache probably need to be fixed, but the library could also help limiting the blast radious.

s50600822:
@Shredder121 do you want to merge it, I don't have write access.

jwgmeligmeyling:
Wouldn't a `Collections.synchronizedMap(new WeakHashMap<String, Template>());` be more appropiate here?

It seems this cache is mostly here to reuse frequently used `Template` instances among operators. Because these are mostly defined in singleton instances (Templates), these are constantly in memory anyway.

Some voices are raising to slowly migrate off from Guava: #2324

jwgmeligmeyling (changes requested):
Please don't merge this yet, I am working on this patch for the 5.x release.

PR #2234, opened 2017-12-29
Dockerize oracle, mysql, postgres, cubrid, firebird, db2, and mongo

johnktims:
This PR fixes our failing travis builds.


johnktims:
I dockerized oracle and db2, but the drivers aren't publicly available so we can't enable those tests in travis.

.travis.yml
```diff
@@ -2,23 +2,17 @@ language: java
...
-  postgresql: "9.3"
+  - docker
 env:
   matrix:
     - MODULE=!querydsl-scala PROFILES=all,travis,examples
     - MODULE=!querydsl-jpa-codegen PROFILES=jpa,hibernate5,travis,examples
     - MODULE=querydsl-scala PROFILES=all,travis
-install:
-  - mvn -B -q --projects ${MODULE} -P${PROFILES} install -DskipTests=true
+install: /bin/true
```
johnktims: This saves us several minutes on the build.

Shredder121: How so? Or is that done in-docker?
As long as the build output (due to artifact download) isn't too crazy it's 👍

Shredder121:
Do you think this is also backportable? 🤔

johnktims:
It can certainly be backported. I can probably get that done this coming weekend.

.travis.yml
johnktims: It saves time because we don't have to run a full `mvn install` before starting the tests. We don't need those artifacts so using `/bin/true` effectively skips that step. If you omit `install:`, travis will add a `mvn install` for you.

timowest:
Looks good

PR #2236, opened 2018-01-08
Update oracle keywords list

### LINKED ISSUE #2225: Incorrect Reserved Word Set for Oracle
Library: querydsl-sql-4.1.4
Oracle: 12.2 and 11.2
Java: 8

I'm running into an issue with paged queries against Oracle whenever the projection contains the column named "role":

    Caused by: java.sql.SQLSyntaxErrorException: ORA-00904: "MY_TABLE"."role": invalid identifier

	at oracle.jdbc.driver.T4CTTIoer11.processError(T4CTTIoer11.java:494) ~[ojdbc8-12.2.0.1.jar:12.2.0.1.0]
	... 
	at oracle.jdbc.driver.OraclePreparedStatementWrapper.executeQuery(OraclePreparedStatementWrapper.java:1165) ~[ojdbc8-12.2.0.1.jar:12.2.0.1.0]
	at com.querydsl.sql.AbstractSQLQuery.fetch(AbstractSQLQuery.java:446) ~[querydsl-sql-4.1.4.jar:?]

Here's the query generated from SQLQuery.fetch()...

    select a.*, rownum rn from (
        select my_table.app, audit_log."role", my_table.dtime, my_table.ip, ... 
        from my_table
        where my_table.dtime >= ? and my_table.dtime <= ?
        order by my_table.dtime asc
     ) a) where rn > 0 and rownum <= 10'

I noticed that "role" is the only column in the projection which had quotes around it. When I repeat the above query manually with the SQL*Plus client, I get the same ORA-00904 error, but if I remove the quotes, the query works as expected. 

I traced the cause to the Keywords.ORACLE field, which loads the "oracle" text file as a resource. The link at the top of that resource calls out an [Oracle 10g doc](https://docs.oracle.com/cd/B19306_01/em.102/b40103/app_oracle_reserved_words.htm), but I don't think that doc is an appropriate reference, as it pertains to the Oracle Enterprise Manager. I think the appropriate references for the keywords resource would be the following, as they pertain specifically to SQL in Oracle:

- [10g Database SQL Reference : Oracle Database Reserved Words](https://docs.oracle.com/cd/B19306_01/server.102/b14200/ap_keywd.htm#i690190)
- [11g Database SQL Language Reference : Oracle SQL Reserved Words](https://docs.oracle.com/cd/E11882_01/server.112/e41084/ap_keywd001.htm#SQLRF55621)
- [12c Database SQL Language Reference : Oracle SQL Reserved Words](https://docs.oracle.com/database/122/SQLRF/Oracle-SQL-Reserved-Words.htm#SQLRF55621)

Shredder121:
Fixes #2225.

So we used the wrong list, can @chaserb confirm this is the right one?


chaserb:
@Shredder121 , yes, this is the correct list. Thanks for the fix!

timowest:
Looks good

Issue #2215, opened 2017-10-27
Document API-based MongodbSerializer


mp911de:
We'd like to MongoDB's Document-API with `MongodbSerializer` to use it with the asynchronous/reactive MongoDB driver. Any chance of migration or providing a new type (side-by-side) supporting `Document`?

olivergierke:
I'd second that. The more mature the 3.x drivers get, the more we risk the `DBObject`-based API being dropped in 4.x. That would effectively make Querydsl useless on MongoDB, which would be a shame. Spring Data currently maintains quite a bit of code to make sure we fall back to that legacy API only for Querydsl and it would be cool if we could get rid of it.

Shredder121:
Now that's a good idea, and inevitable when they decide to axe the old API.

PR #2239, opened 2018-01-15
Initial draft of MongoDB Document API support

mp911de:
This pull request introduces Document API support for MongoDB (see #2215) as additional implementation, without changing `DBObject` support.

Morphia is entirely based on `DBObject`. Tests are a bit of a mixture of objects and `Document`s right now. Happy to dive into a discussion.

### DIFF
--- a/querydsl-mongodb/src/main/java/com/querydsl/mongodb/document/AbstractFetchableMongodbQuery.java
+++ b/querydsl-mongodb/src/main/java/com/querydsl/mongodb/document/AbstractFetchableMongodbQuery.java
@@ -14,20 +14,20 @@
 package com.querydsl.mongodb.document;
 
 import com.google.common.base.Function;
+import com.google.common.collect.HashMultimap;
+import com.google.common.collect.Multimap;
 import com.mongodb.ReadPreference;
 import com.mongodb.client.FindIterable;
 import com.mongodb.client.MongoCollection;
 import com.mongodb.client.MongoCursor;
 import com.mysema.commons.lang.CloseableIterator;
 import com.querydsl.core.*;
-import com.querydsl.core.types.Expression;
-import com.querydsl.core.types.OrderSpecifier;
-import com.querydsl.core.types.Path;
-import com.querydsl.core.types.Predicate;
+import com.querydsl.core.types.*;
 import org.bson.Document;
 
 import javax.annotation.Nullable;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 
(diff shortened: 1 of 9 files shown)


querydsl-mongodb/src/main/java/com/querydsl/mongodb/document/AbstractFetchableMongodbQuery.java
```diff
@@ -0,0 +1,286 @@
...
+import java.util.List;
+
+/**
+ * {@link Fetchable} Mongodb query with a pluggable Document to Bean transformation.
+ *
+ * @param <K> result type
+ * @param <Q> concrete subtype
+ * @author Mark Paluch
+ */
+public abstract class AbstractFetchableMongodbQuery<K, Q extends AbstractFetchableMongodbQuery<K, Q>>
```
mp911de: There's a split between some base functionality and the actual execution to enable different execution models (async, reactive) to be built on top of `AbstractMongodbQuery`.

Shredder121:
Just wanted to mention, looks nice so far.
👀

So what would you recommend with the split of execution modes? How would you see the reactive mongodbquery work?

mp911de:
The main issue to tackle is query execution required for joins/relations. Since RxJava 1 is EOL, the only reactive driver is the Reactive Streams one. Maybe we could introduce something like continuations that are agnostic in regard to the underlying execution model.

This increases complexity in general but could be a nice approach to support both execution models from a single base class.

The idea is to introduce something like callback interfaces along the desired operation/criteria to query on a particular collection.

olivergierke:
Hey guys, any chance we can get some feedback on your plans going forward. The MongoDB Java driver team has indicated that it plans to remove the deprecated `DBObject` based APIs in some future version and we don't want to end up with the Spring Data Querydsl support ending up broken in a driver release.

I wouldn't mind getting the support for the new APIs introduced in a parallel type hierarchy so that we can move Spring Data over to the new APIs.

Shredder121:
I myself am not well-versed in the world of reactive programming, so maybe you, @mp911de can maintain the mongodb subproject once it's in the mainline? I don't know if it's something you are able/willing to do?

olivergierke:
We're not talking anything reactive here but the fact that the MongoDB Java driver switched from using `DBObject` to `Document` for its synchronous operations a while ago.

mp911de:
This pull request does not contain any reactive bits. However, there's a split between base functionality and the actual class which executes the query to not make assumptions about the execution model within basic query creation functionality.

Shredder121:
> We're not talking anything reactive here

Oh, sorry. In that case I wouldn't mind migrating to Document exclusively, for Querydsl 5.
Unless you want both `DBObject` and `Document` support for now? Or both in Querydsl 4 and 5 respectively?

> This pull request does not contain any reactive bits. However, there's a split between base functionality and the actual class which executes the query to not make assumptions about the execution model within basic query creation functionality.

*EDIT*: saw Mark's comment just now, yes I know about the split, but I'd rather you take a bit of responsibility, so you can get the mongodb submodule (and with assist also core) up to spec for later having reactive queries.

Then I think the process can be a bit more sped up, since I've been really busy lately, as you probably noticed.

jwgmeligmeyling:
@odrotbohm @mp911de Is this PR actually sufficient to port Spring Data over to the new API already? Or is this really still the initial draft?

Our current set of maintainers has little experience or ongoing work with MongoDB, so it is hard for me to review this properly. I am very willing to integrate it though if it helps you forward and helps the overall compatibility with `DBObject` being deprecated.

That said, what do you think about bumping the dependency versions of the MongoDB driver and/or Morphia (see #2029)? Which version is Spring Data currently targetting?

mp911de:
Let me have a look and potentially rebase the code. MongoDB 4.1 is btw. the most recent version. Spring Data MongoDB 3.1 (to be released in late October) is targeting the MongoDB 4.1 driver.

I will leave a comment with further details once I had a look.

jwgmeligmeyling:
Great, thanks! Lets finally get this out of the way after 2 years 😉

mp911de:
Meh! I've deleted my repository in the meantime and so this PR base got lost. I suggest creating a new PR from https://github.com/mp911de/querydsl/tree/issue/mongodb-document-api and closing this one. We fixed a few issues in Spring Data (we've inlined the Querydsl code) and the new PR would include also the fixes.

mp911de:
Closing in favor of #2689.

PR #2241, opened 2018-01-16
Explicitly declare NonUniqueResultException for fetchOne

### LINKED ISSUE #2232: Missing NonUniqueResultException from Fetchable.fetchOne() method documentation
```
Description copied from interface: com.querydsl.core.Fetchable 
Get the projection as a unique result or null if no result is found
Overrides:
fetchOne in class ProjectableSQLQuery
Returns:
first result or null
```
If more than one result is returned, NonUniqueResultException is thrown.

reegnz:
Fetchable.fetchOne() declares in it's javadoc that it throws
NonUniqueResultException. It should therefore be explicitly declared in
the method signature as well.
All implementations must also declare in their signature that they throw.

This also has the desirable side-effect of the javadoc of all of the
implementors inheriting the @throws section from the javadoc of the
interface as well.

Fixes #2232

### DIFF
--- a/querydsl-collections/src/main/java/com/querydsl/collections/AbstractCollQuery.java
+++ b/querydsl-collections/src/main/java/com/querydsl/collections/AbstractCollQuery.java
@@ -196,7 +196,7 @@ public abstract class AbstractCollQuery<T, Q extends AbstractCollQuery<T, Q>> ex
     }
 
     @Override
-    public T fetchOne() {
+    public T fetchOne() throws NonUniqueResultException {
         queryMixin.setUnique(true);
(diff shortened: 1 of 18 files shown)


timowest:
Looks good

Shredder121:
The exception is actually a RuntimeException, since it's an opt in to handle it, it's not required.
But on the other hand, it's better if it is copied into the Javadoc.
Because it's not a compile error on client code this can go in any release, so I'll merge it.
Thanks!

Shredder121:
For now attaching it to 5.0, so it will at least end up in some changelog.

PR #2252, opened 2018-02-08
Simplify array creation

yorlov:
There are two styles to convert a collection to an array: either using a pre-sized array (like c.toArray(new String[c.size()])) or using an empty array (like c.toArray(new String[0]).

In older Java versions using pre-sized array was recommended, as the reflection call which is necessary to create an array of proper size was quite slow. However since late updates of OpenJDK 6 this call was intrinsified, making the performance of the empty array version the same and sometimes even better, compared to the pre-sized version.


Shredder121:
👍 Looks fine to me, thanks!
Although to be consistent, can you still do the presized variant?

yorlov:
yes, no problem

timowest:
Looks good

Issue #2190, opened 2017-08-23
Delete with join


pelizza:
Hi,

It would be great to have support for delete operations with join. For example:
```
dsl.delete(SOMETHING)
                .join(ANOTHER)
                .where(SOMETHING.id.in(ids))
                .where(ANOTHER.type.eq("My type"))
                .execute();
```

Thanks!

Issue #776, opened 2014-05-27
PostgreSQL: UPDATE ... SET ... FROM ... WHERE ...


beloglazov:
PostgreSQL supports the `from` clause in update queries, which is useful when you want to update a subset of rows based on a where condition doing an implicit join. In the current version of querydsl I can't find how this can be done. Is this feature supported in some way or is planned to be supported? Thanks!

PR #2253, opened 2018-02-09
Expose fields needed for dialect-specific SQL syntax

rdnetto4:
Hi,
This PR makes some minor changes to allow consumers to implement dialect-specific syntax by subtyping the AbstractSQLUpdateClause, etc. I've used this to implement UPDATE FROM (#776) & DELETE USING (#2190) support for Postgres in our codebase. I'm happy to upstream them if there's interest, but wasn't sure whether we're interested in supporting them.

I've also tweaked the Travis build to use mvnvm, so we can rely on the same version of Maven being used locally and in CI.


Shredder121:
First thoughts:

Looks good, but are these all? I recall *Query types; postgres, Oracle.

We have a custom import order,
`static`s, `java`, `javax`, then `org`, `com`.
Which I'll encode in the checkstyle rules since I'm working on those anyway :sweat_smile:

rdnetto4:
> Looks good, but are these all? I recall *Query types; postgres, Oracle.

The subtypes of AbstractSQLQuery (PostgresSQLQuery, OracleQuery, etc.) have very little non-boilerplate code in them. e.g OracleQuery just has 5 one-line methods. That said, I suppose it would make more sense for consumers who are only targetting one dialect to subtype those instead of copying and pasting. I'll apply the same pattern to those too.

> We have a custom import order
Oops, will fix. :sweat_smile:

rdnetto4:
Imports should be in the right order now. Let me know if there's any other changed you'd like me to make.

timowest:
Are the SQLSerializer changes necessary? It is not really meant to be a template class. Open it up for extension is IMHO misleading, since templates are the way to support new dialects.

rdnetto4:
> Are the SQLSerializer changes necessary? It is not really meant to be a template class. Opening it up for extension is IMHO misleading, since templates are the way to support new dialects.

They unfortunately are. The customization provided by templates is effectively limited to replacing keywords and adding flags - it's not able to handle something like "I want to append the qualified names of the tables and their aliases here", which is what you really need to add dialect-specific features.

I think it makes sense for us to distinguish between minor tweaks needed to handle ambiguities in the spec or variations in its implementation (differences in keywords, to join a dummy table, etc.) and structural changes needed to support dialect-specific features - we should templates for the former, and subtype serialiser for the latter. For example, #776 summarises the syntaxes used for UPDATE FROM like so:
```
CUBRID: UPDATE ud, sales SET ...
MySQL: UPDATE ud, sales SET ...
Oracle: UPDATE (select ... from ud inner join ...) SET ...
PostgreSQL: UPDATE ud SET ... FROM sales
SQL Server: UPDATE ud SET ... FROM ud INNER JOIN ...
```
The complexity of handling all of them in a single class is non-trivial, and likely why that feature hasn't been implemented until now.

In comparison, subclassing the serialiser makes this pretty straight-forward. To give you an idea of what this looks like, I've attached our Postgres-specific implementation:
[ExtendedSQLUpdateClause.java](https://github.com/querydsl/querydsl/files/1735588/ExtendedSQLUpdateClause.java.txt)

rdnetto4:
@timowest @Shredder121 would appreciate it if you could let me know if you think this is the way forward or not - I'm keen to get this merged & released as soon as possible.

Shredder121:
Looks good to me, yes.
But could you tell something about mvnvm?

Shredder121:
Anyway, could you tell something about mvnvm @rdnetto4?

rdnetto4:
> But could you tell something about mvnvm?

Mvnvm is a script that pins the Maven version via `mvnvm.properties`. Without it, the version of maven used in CI depends purely on whatever happens to be on the image used in CI, which may not be the same as the version installed locally.

> This is a good candidate for Querydsl 5, because the methods will have moved.

Because the methods have only been moved into the base class, the ABI is unchanged, so a major version bump should be unnecessary. That said, if that's going to be the next release I have no issues with it going out as part of it.

Shredder121:
Well you do have a point.
I thought it over, this release sounds fine to me.

rdnetto4:
Awesome, so it's fine to merge now?

PR #2267, opened 2018-03-09
Fix inverted equals expression in ne()

### LINKED ISSUE #2197: FetchableSubQueryBase#ne(T) incorrectly make an equals() instead of not-equals()
When looking at the code of `com.querydsl.core.support.FetchableSubQueryBase#ne(T)` we noticed that an equals is done:

`
    @Override
    public BooleanExpression ne(T constant) {
        return eq(Expressions.constant(constant));
    }
`

it should be corrected to

`
return ne(Expressions.constant(constant));
`

Affected version of QueryDsl: 4.1.4.

Shredder121:
Fixes #2197 
@timowest is this still candidate for 4.2.0? or for the next version?


timowest:
Looks good, can be included in 4.2.0

Shredder121:
Done!

PR #2273, opened 2018-03-18
Fixes for release

timowest:


Shredder121:
:+1:
Could you also release Querydsl 3?
We don't have a schedule yet on when to drop support, but it would be good to get another release in.
If you have the time.

PR #2280, opened 2018-03-26
[SQL] Apply column name override in executeWithKey call

balazs-zsoldos:
Column names should be overridden from the configuration in case
the auto-generated keys are retrieved from the database after
executing an Insert statement.

### DIFF
--- a/querydsl-sql/src/main/java/com/querydsl/sql/dml/AbstractSQLInsertClause.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/dml/AbstractSQLInsertClause.java
@@ -325,6 +325,7 @@ public abstract class AbstractSQLInsertClause<C extends AbstractSQLInsertClause<
                 for (int i = 0; i < target.length; i++) {
                     Path<?> path = entity.getPrimaryKey().getLocalColumns().get(i);
                     String column = ColumnMetadata.getName(path);
+                    column = configuration.getColumnOverride(entity.getSchemaAndTable(), column);
                     target[i] = column;
                 }
                 stmt = connection().prepareStatement(queryString, target);
--- a/querydsl-sql/src/test/java/com/querydsl/sql/InsertBase.java
+++ b/querydsl-sql/src/test/java/com/querydsl/sql/InsertBase.java
@@ -31,6 +31,7 @@ import org.junit.Before;
 import org.junit.Ignore;
 import org.junit.Test;
 
+import com.querydsl.core.QueryException;
 import com.querydsl.core.QueryFlag.Position;
 import com.querydsl.core.Tuple;
 import com.querydsl.core.testutil.ExcludeIn;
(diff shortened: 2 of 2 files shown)


Shredder121:
Quick glance at it, looks fine. Would some sort of test be viable that verifies it's applied?

balazs-zsoldos:
@Shredder121 See my second commit. It seems that some of the JDBC drivers do not care about the passed column names in case executeWithKey is used. I did set the test function to run on Derby and Hsqldb as they use the passed parameter for sure.

Shredder121:
👍

balazs-zsoldos:
ping :)

balazs-zsoldos:
ping

balazs-zsoldos:
ping

balazs-zsoldos:
ping

Shredder121:
At the moment I have little time to merge PRs, as well as also have some changes in my working directory that must come first.
So when my personal issues are out of the way I can do it.

Sorry for the long long wait, appreciated though!

balazs-zsoldos:
@Shredder121 Thanks for the info!

balazs-zsoldos:
ping

I would have a couple of other ideas to contribute, but only if it is handled anyhow. E.g.:

 - Possibility of attaching customizers to codegenerator: E.g.: making Q or bean classes import specific interface(s) if some fields are present. By having this, we can write more reusable code for classes with the same fields inside
 - Supporting java.time package
 - etc.

Shredder121:
Supporting java.time sounds good, but how would you handle configuration for "importing specific interfaces", and what will be the end result?

balazs-zsoldos:
> but how would you handle configuration for "importing specific interfaces", and what will be the end result?

I would choose not to decide the configuration and real logic :). I would add the possibility somehow to modify the class generation via "plugins". It would be the responsibility of the plugin creators to find out the configuration logic and the goal of that plugin. The interface example:

A project that used the same fields in many tables for auditlog can have also a plugin that makes every class implement the myproject.Auditable interface that has the relevant fields (lastChangeBy, lastChangeAt, ...).

Another feature that came up in the last days that QBean could be enhanced to support enums. If the Pojo has an enum type, it can convert from the string representation to Enum by calling Enum.valueOf(). I would be happy to discuss this with the project owner and send a PR if I see that it might be pulled in one day.

My problem now is that I am not sure when my PR would be accepted (or rejected). I know that maintaining an OSS project is a lot of work, but it would be good to know what we can expect. This PR is a very simple bugfix (with one line of change) and I am wondering if it is ever pulled into a new release.

balazs-zsoldos:
Ping.

I am not sure who checks merge requests now. But this is waiting for so long and the bugfix is so simple and tested... Could you, please, merge it into the main branch if everything is ok?

idosal:
@balazs-zsoldos Merged. Sorry for the wait. As you may have noticed, quite a few PRs have been merged in the last months. Unfortunately, this one kind of fell through the cracks. If you have any other ideas/PRs we'd be happy to hear. Thanks!

balazs-zsoldos:
@idosal Thanks a lot!

PR #2286, opened 2018-04-12
Allow blank "schemaPattern" in the Maven plugin

### LINKED ISSUE #2282: Handling empty schemaPattern in querydsl-maven-plugin
In the querydsl-maven-plugin. when the `schemaPattern` parameter is specified as:
```xml
<schemaPattern></schemaPattern>
```
it is converted to `null` instead of an empty string. This is a well-known Maven limitation.

However, these two values have different meanings when the [`DatabaseMetadata`](https://docs.oracle.com/javase/8/docs/api/java/sql/DatabaseMetaData.html#getTables-java.lang.String-java.lang.String-java.lang.String-java.lang.String:A-) class processes them:
```
"" retrieves those without a schema
<code>null</code> means that the schema name should not be used to narrow the search
```
There was a similar issue raised earlier, for the `namePrefix` parameter. This was handled by using the special string `BLANK` to represent an empty string value.

> **PR #1811 - Fix empty value handling**

Can we implement similar logic for the `schemaPattern` parameter (including multi-value handing)?

metacubed:
Fixes #2282

### DIFF
--- a/querydsl-maven-plugin/src/main/java/com/querydsl/maven/AbstractMetaDataExportMojo.java
+++ b/querydsl-maven-plugin/src/main/java/com/querydsl/maven/AbstractMetaDataExportMojo.java
@@ -18,6 +18,7 @@ import java.sql.Connection;
 import java.sql.DriverManager;
 import java.sql.SQLException;
 import java.util.Comparator;
+import java.util.regex.Pattern;
 
 import org.apache.maven.artifact.manager.WagonManager;
 import org.apache.maven.plugin.AbstractMojo;
@@ -431,7 +432,7 @@ public class AbstractMetaDataExportMojo extends AbstractMojo {
             exporter.setTargetFolder(new File(targetFolder));
             exporter.setNamingStrategy(namingStrategy);
             exporter.setCatalogPattern(catalogPattern);
-            exporter.setSchemaPattern(schemaPattern);
+            exporter.setSchemaPattern(processBlankValues(schemaPattern));
             exporter.setTableNamePattern(tableNamePattern);
             exporter.setColumnAnnotations(columnAnnotations);
             exporter.setValidationAnnotations(validationAnnotations);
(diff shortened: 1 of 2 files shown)


metacubed:
@Shredder121, can you point me to the 4.2.2 branch? I'll re-target this PR to that branch instead of master.

Shredder121:
That would be QUERYDSL_4 👍
thanks

metacubed:
Done, @Shredder121

querydsl-maven-plugin/src/main/java/com/querydsl/maven/AbstractMetaDataExportMojo.java
```diff
@@ -689,4 +690,14 @@ private static String emptyIfSetToBlank(String value) {
...
+
+    private static String processBlankValues(String value) {
+        if (value == null) {
+            return null;
+        }
+        return BLANK_VALUE_PATTERN.matcher(value).replaceAll(BLANK_VALUE_REPLACEMENT);
+    }
+
+    private static final Pattern BLANK_VALUE_PATTERN = Pattern.compile("(^|,)BLANK(,|$)", Pattern.CASE_INSENSITIVE);
+    private static final String BLANK_VALUE_REPLACEMENT = "$1$2";
```
Shredder121: Why have you decided to do it differently than `emptyIfSetToBlank`?
If this is a better way, why not move it all over?

metacubed: @Shredder121, the two scenarios require different behaviors, and I didn't want to break backward compatibility for the existing API:

`emptyIfSetToBlank` (existing):
- Converts `null` to an empty string
- Converts `"BLANK"` to an empty string
- Works with a single value

`processBlankValues` (new):
- Keeps `null` values as `null`
- Converts `"BLANK"` to an empty string
- Works with multiple comma-separated values (not a compatibility concern)

This change in behavior is required since the two values are treated differently by the `DatabaseMetadata` class.

If you are ok with changing the null-handling behavior of `namePrefix`, then the two functions can be merged.

metacubed:
@Shredder121, are you ok with this PR?

metacubed:
@johnktims, could you take a look at this PR? Also let me know if I need to change the target branch.

johnktims:
@metacubed Yes, please change the target branch to `master`.

metacubed:
@johnktims, done. Please take a look.

jwgmeligmeyling:
Made up to date with master in https://github.com/querydsl/querydsl/tree/I2282-EmptySchemaPattern

metacubed:
@johnktims, rebased and updated to the latest master (as of July 30).

johnktims:
Thanks for the contribution @metacubed !

PR #2312, opened 2018-05-25
Translate SQLException thrown during connection acquisition

chrisribble:
I noticed (while debugging a connection failure in an application) that SQLQueryFactory.DataSourceProvider's get method always throws RuntimeException when an SQLException is thrown during connect instead of delegating construction of the exception to the defined SQLExceptionTranslator.

This forces API consumers to implement custom logic to catch RuntimeException and then do the mapping on the RuntimeException cause (if not null).

This is obviously suboptimal, so I propose delegating to the SQLExceptionTranslation defined in the Configuration (which defaults to DefaultSQLExceptionTranslator if none is defined by the user).

I have not written tests yet as I wanted to get feedback on the change first, but I can write unit tests once there is some agreement on the approach.

### DIFF
--- a/querydsl-sql/src/main/java/com/querydsl/sql/SQLQueryFactory.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/SQLQueryFactory.java
@@ -34,9 +34,11 @@ public class SQLQueryFactory extends AbstractSQLQueryFactory<SQLQuery<?>> {
     static class DataSourceProvider implements Provider<Connection> {
 
         private final DataSource ds;
+        private final Configuration configuration;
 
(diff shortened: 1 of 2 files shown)


chrisribble:
@Shredder121, do you have any objections to this as is? Any reason not to merge now that it has tests?

Shredder121:
Looks good to me, thanks!

chrisribble:
@Shredder121, done; this now targets the QUERYDSL_4 branch. Hopefully I did it right. :)

chrisribble:
@Shredder121 any reason not to go ahead and merge this?

jwgmeligmeyling:
Thanks! 👍

PR #2314, opened 2018-06-07
Static properties referencing to the default variable properties

encircled:
Hi guys,
This PR allows to generate additional static properties, which are just a reference to properties of the default variable, for instance: 

```java
public class QPerson extends EntityPathBase<Person> {

    public static final QPerson person = new QPerson("person");
 
    public final StringPath firstName = createString("firstName");

    public static final StringPath firstName_ = person.firstName;

}
```

so it is possible to write:

`QPerson.firstName_` instead of `QPerson.person.firstName`. I believe that it will noticeably reduce amount of redundant code in some projects.

I've added `createDefaultVariableStaticProperties` Config param for this and it should be backward-compatible

### DIFF
--- a/querydsl-apt/src/main/java/com/querydsl/apt/AbstractQuerydslProcessor.java
+++ b/querydsl-apt/src/main/java/com/querydsl/apt/AbstractQuerydslProcessor.java
@@ -330,8 +330,8 @@ public abstract class AbstractQuerydslProcessor extends AbstractProcessor {
         String typeName = type.toString();
 
         if (typeName.startsWith(Collection.class.getName())
-         || typeName.startsWith(List.class.getName())
-         || typeName.startsWith(Set.class.getName())) {
+                || typeName.startsWith(List.class.getName())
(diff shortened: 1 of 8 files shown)


Shredder121:
> `QPerson.firstName_` instead of `QPerson.person.firstName`. I believe that it will noticeably reduce amount of redundant code in some projects.

What limits you from 1) statically importing the default variable `person.firstName` or 2) defining fields at the top of your java class for the query types you need?



> This PR allows to generate additional static properties

I'm not really a fan of this, we already have a bunch of things we generate and we can already have clashes when a property has the same name as the class' simple name.

encircled:
Thats why it is configurable, I can add name checks as well.

Such static imports might clash with other variables (and pretty often are forbidden by style convention).
Top-level definition is possible, but it is boilerplate code from my POV

jwgmeligmeyling:
This is redundant. Users should just statically import person and use `person.firstName`. Users may not even restrict themselves to the default alias, so what would `Person.person_` then mean?

PR #2322, opened 2018-06-18
Make SpatialSupport available in non-apt use case

### LINKED ISSUE #2320: HibernateDomainExporter does not register spatial types
## tl;dr
Is it possible to either

* configure the Maven plugin generating code from a set of `hbm.xml` files
* configure the `JPADomainExporter` to register spatial types 

I use version `4.1.3`. 

## Context

Section `2.1.4. Generating the model from hbm.xml files` describes a way [to generate code from XML mapping files](http://www.querydsl.com/static/querydsl/4.1.3/reference/html_single/#d0e196). Having a simple Spring Boot Application:

```java
@SpringBootApplication
public class Application implements CommandLineRunner {

    public static void main(String[] args) {
        SpringApplication.run(Application.class, args);
    }

    @Autowired
    @Qualifier("entityManagerFactory")
    private EntityManagerFactory entityManagerFactory;

    @Bean
    @Primary
    public SessionFactory sessionFactory() {
        if (entityManagerFactory.unwrap(SessionFactory.class) == null) {
            throw new NullPointerException("factory is not a hibernate factory");
        }
        return entityManagerFactory.unwrap(SessionFactory.class);
    }

    @Bean
    public EntityManagerFactory entityManagerFactory(DataSource datasource, JpaProperties properties)
            throws IOException {
        LocalContainerEntityManagerFactoryBean emf = new LocalContainerEntityManagerFactoryBean();
        emf.setJpaVendorAdapter(new HibernateJpaVendorAdapter());
        emf.setJpaPropertyMap(properties.getProperties());
        emf.setDataSource(datasource);
        emf.afterPropertiesSet();
        return emf.getNativeEntityManagerFactory();
    }

    @Override
    public void run(String... args) throws Exception {
        createExporter("Q").execute();
    }

    private JPADomainExporter createExporter(String prefix) {
        final Metamodel metamodel = entityManagerFactory.getMetamodel();
        return new JPADomainExporter(prefix, new File("src/main/java"), metamodel);
    }

}
```

I first tried to use `HibernateDomainExporter` but the above config also worked out for `JPADomainExporter`. The mapping files are declared within a `META-INF/persistence.xml` on the classpath, like so:

```xml
<persistence xmlns="http://xmlns.jcp.org/xml/ns/persistence"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    xsi:schemaLocation="http://xmlns.jcp.org/xml/ns/persistence
             http://xmlns.jcp.org/xml/ns/persistence/persistence_2_1.xsd"
    version="2.1">
    <persistence-unit name="default-unit">
        <description>Default Mappings</description>
        <mapping-file>/hbm/mapping-resource_1.hbm.xml</mapping-file>
        <mapping-file>/hbm/mapping-resource_2.hbm.xml</mapping-file>

        <!-- et cetera --> 

    </persistence-unit>
</persistence>
```

This works rather well. However, I do have spatial types which are not mapped by default. The documentation hints me to [add `<spatial>true</spatial>` to the Maven plugin config](http://www.querydsl.com/static/querydsl/4.1.3/reference/html_single/#spatial), but I did not find a way to configure the maven codegen plugin to work with `hbm.xml` mappings.

ridoo:
Restricting the registration of spatial paths to apt-only cases makes
it impossible to use spatial paths when using `JPADomainExporter` for
example. Therefore, `SpatialSupport` has been moved to `querydsl-codegen`
module to be available from `AbstractDomainExporter`.

Fixes #2320

### DIFF
--- a/querydsl-apt/src/main/java/com/querydsl/apt/DefaultConfiguration.java
+++ b/querydsl-apt/src/main/java/com/querydsl/apt/DefaultConfiguration.java
@@ -205,8 +205,12 @@ public class DefaultConfiguration implements Configuration {
         }
         module.bind(CodegenModule.VARIABLE_NAME_FUNCTION_CLASS, variableNameFunction);
 
-        if (SpatialSupport.isSpatialOnClassPath()) {
+        try {
+            // register additional mappings if querydsl-spatial is on the classpath
+            Class.forName("com.querydsl.spatial.GeometryExpression");
             SpatialSupport.addSupport(module);
+        } catch (Exception e) {
+            // do nothing
         }
 
         defaultSerializerConfig = new SimpleSerializerConfig(entityAccessors, listAccessors,
--- a/querydsl-apt/src/main/java/com/querydsl/apt/DefaultConfiguration.java
+++ b/querydsl-apt/src/main/java/com/querydsl/apt/DefaultConfiguration.java
@@ -205,12 +205,8 @@ public class DefaultConfiguration implements Configuration {
         }
         module.bind(CodegenModule.VARIABLE_NAME_FUNCTION_CLASS, variableNameFunction);
 
-        try {
-            // register additional mappings if querydsl-spatial is on the classpath
-            Class.forName("com.querydsl.spatial.GeometryExpression");
+        if (SpatialSupport.isSpatialOnClassPath()) {
             SpatialSupport.addSupport(module);
-        } catch (Exception e) {
-            // do nothing
         }
 
         defaultSerializerConfig = new SimpleSerializerConfig(entityAccessors, listAccessors,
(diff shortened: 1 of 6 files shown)


ridoo:
I fear that this is a breaking change as moving `SpatialSupport` has needed to make it availble from `AbstractDomainExporter`. I could create a backwards compatible `SpatialSupport` in the apt-module which delegates to the new one (should have to be renamed then).

jwgmeligmeyling:
I'm currently reworking spatial support as JTS has moved to another group ID in Maven and we need to support both. I have also formerly implemented codegen extension points. So I actually plan to move the spatial related codegen to the spatial module which then will also allow it to be used in other modules as well.

Its on my agenda for inclusion in the 5.x release (where we can make breaking changes).

PR #2330, opened 2018-06-28
Expressions: fix equals method

petrkotek:
I've discovered that when I compare two _same_ predicates which include expressions generated by `Expressions.asString("foo")`, `.equals()` method returns `false`.

Example `Expressions.asString("foo").equals(Expressions.asString("foo"))`
- Expected result: `true`
- Actual result: `false`

This PR adds the `equals` method with appropriate implementation for the objects returned by `Expressions.asString()`, `Expressions.asNumber()`, ... etc

Also add appropriate test coverage.

Looking forward to any feedback.

### DIFF
--- a/querydsl-core/src/main/java/com/querydsl/core/types/dsl/Expressions.java
+++ b/querydsl-core/src/main/java/com/querydsl/core/types/dsl/Expressions.java
@@ -1895,6 +1895,18 @@ public final class Expressions {
                     return this.mixin.accept(v, context);
                 }
 
+                @Override
+                public boolean equals(Object o) {
+                    if (o == this) {
+                        return true;
+                    } else if (o instanceof BooleanExpression) {
+                        BooleanExpression other = (BooleanExpression) o;
+                        return (other.mixin.equals(this.mixin));
+                    } else {
+                        return false;
+                    }
+                }
+
             };
         }
     }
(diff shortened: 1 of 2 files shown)


shrds:
I think this will also fix the query projection orderBy on string issue. see more details of the issue I have experienced. if the name in my entity object is longer than 1 letter then the orderBy name fails and gives me unexpected results.

Shredder121:
👍 Looks nice, thanks.
Could you either target `QUERYDSL_4`, or open an extra one and target `QUERYDSL_4`?
I believe this should be fixed in there as well.

shrds:
@petrkotek I tested your changes and when I use  orderBy = Expressions.asString("name").asc();  in my query it works with the overridden equals() method.
However when I use the StringPath I get unexpected behavior in the resultset.
Expressions.asString(definition.name.stringValue()).asc();

I tried Overriding the equals method in [StringPath.java](https://github.com/petrkotek/querydsl/blob/7258df0c10fcfeee7cea10c5113676445c02c9b4/querydsl-core/src/main/java/com/querydsl/core/types/dsl/StringPath.java) but that still fails with not grouping the children under one parent id together and it duplicates the parent id with rest of the child records. 

Do you know if I need to make any other changes?

jwgmeligmeyling:
Thanks for your contribution!

PR #2335, opened 2018-07-05
Fix bugs in EclipseLinkHandler#iterate()

rdicroce:
EclipseLinkHandler's implementation of iterate() is heavily bugged:
- Some environments (e.g. WildFly) wrap the Query, causing the
instanceof check to fail, which prevents a cursor from being used. Fix
this by unwrapping the Query first.
- Telling EclipseLink to use a scrollable cursor, then telling it to use
a result set type of FORWARD_ONLY, makes no sense and causes the SQL
Server JDBC driver to throw an exception. Fix this by using a regular
CursoredStream instead.
- The original code never tells EclipseLink to release objects, causing
any benefit from using a cursor to be lost since all objects are
retained until the cursor is closed. Fix this by releasing objects after
completing each page.

With these changes, iterate() now works correctly on WildFly 13 with EclipseLink 2.7.2 and mssql-jdbc 6.5.3.

### DIFF
--- a/querydsl-jpa/src/main/java/com/querydsl/jpa/EclipseLinkHandler.java
+++ b/querydsl-jpa/src/main/java/com/querydsl/jpa/EclipseLinkHandler.java
@@ -17,10 +17,11 @@ import java.io.Closeable;
 import java.io.IOException;
 import java.util.Iterator;
 
+import javax.persistence.PersistenceException;
 import javax.persistence.Query;
 
+import org.eclipse.persistence.config.HintValues;
 import org.eclipse.persistence.config.QueryHints;
-import org.eclipse.persistence.config.ResultSetType;
 import org.eclipse.persistence.jpa.JpaQuery;
 import org.eclipse.persistence.queries.Cursor;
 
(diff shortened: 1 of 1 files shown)


querydsl-jpa/src/main/java/com/querydsl/jpa/EclipseLinkHandler.java
```diff
@@ -54,20 +55,40 @@ public boolean createNativeQueryTyped() {
     @SuppressWarnings("unchecked")
     @Override
     public <T> CloseableIterator<T> iterate(Query query, FactoryExpression<?> projection) {
+        boolean canUseCursor = false;
+        try {
+            canUseCursor = query.unwrap(Query.class) instanceof JpaQuery;
```
jwgmeligmeyling, line 60: You can just unwrap `org.eclipse.persistence.jpa.JpaQuery` here.

querydsl-jpa/src/main/java/com/querydsl/jpa/EclipseLinkHandler.java
```diff
@@ -54,20 +55,40 @@ public boolean createNativeQueryTyped() {
...
+        } catch (PersistenceException e) { } // can't unwrap, just ignore the exception
```
jwgmeligmeyling, line 61: Then this is just your else branch, so you can eliminate the if / else block.

jwgmeligmeyling:
Could you rebase on the current master so that the tests run?

querydsl-jpa/src/main/java/com/querydsl/jpa/EclipseLinkHandler.java
rdicroce: It's been 2 years but IIRC I did it this way as a defense against possible buggy wrappers. The Javadoc for javax.persistence.Query is clear that if the requested class can't be unwrapped, then unwrap() should throw a PersistenceException. But there's no way to guarantee a wrapper will implement that correctly - it could just return null instead. Still want me to change it?

rdicroce:
Sorry about the delay - I was furloughed for the last 2 weeks. I've rebased the changes.

querydsl-jpa/src/main/java/com/querydsl/jpa/EclipseLinkHandler.java
jwgmeligmeyling: Actually I recently stumbled upon buggy wrappers as well, so lets keep it like this.

PR #2354, opened 2018-08-11
Hibernate 5 3 fix proper solution

PavlikPolivka:
My attempt at fixing querydsl#2326

### DIFF
--- a/querydsl-core/src/main/java/com/querydsl/core/support/SerializerBase.java
+++ b/querydsl-core/src/main/java/com/querydsl/core/support/SerializerBase.java
@@ -13,13 +13,19 @@
  */
 package com.querydsl.core.support;
 
-import java.util.*;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.querydsl.core.JoinFlag;
 import com.querydsl.core.QueryFlag;
 import com.querydsl.core.types.*;
+
+import static java.util.Collections.unmodifiableMap;
 
 /**
  * {@code SerializerBase} is a stub for Serializer implementations which serialize query metadata to Strings
@@ -101,7 +107,7 @@ public abstract class SerializerBase<S extends SerializerBase<S>> implements Vis
             allLabels.put(entry.getKey(), entry.getValue().toString());
         }
 
-        return Collections.unmodifiableMap(allLabels);
+        return unmodifiableMap(allLabels);
     }
 
     protected int getLength() {
--- a/querydsl-core/src/main/java/com/querydsl/core/support/SerializerBase.java
+++ b/querydsl-core/src/main/java/com/querydsl/core/support/SerializerBase.java
@@ -25,13 +25,12 @@ import com.querydsl.core.types.*;
  * {@code SerializerBase} is a stub for Serializer implementations which serialize query metadata to Strings
  *
  * @param <S> concrete subtype
- *
  * @author tiwe
  */
-public abstract class SerializerBase<S extends SerializerBase<S>> implements Visitor<Void,Void> {
+public abstract class SerializerBase<S extends SerializerBase<S>> implements Visitor<Void, Void> {
 
     private static final Set<Operator> SAME_PRECEDENCE = ImmutableSet.<Operator>of(Ops.CASE,
-        Ops.CASE_WHEN, Ops.CASE_ELSE, Ops.CASE_EQ, Ops.CASE_EQ_WHEN, Ops.CASE_EQ_ELSE);
+            Ops.CASE_WHEN, Ops.CASE_ELSE, Ops.CASE_EQ, Ops.CASE_EQ_WHEN, Ops.CASE_EQ_ELSE);
 
     private final StringBuilder builder = new StringBuilder(128);
 
@@ -41,7 +40,8 @@ public abstract class SerializerBase<S extends SerializerBase<S>> implements Vis
 
     private String anonParamPrefix = "_";
 
-    private Map<Object,String> constantToLabel;
+    private Map<Object, String> constantToNamedLabel;
+    private Map<Object, Integer> constantToNumberedLabel;
 
     @SuppressWarnings("unchecked")
     private final S self = (S) this;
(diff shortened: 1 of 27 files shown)


kdebski85:
Some Firebird test failed with GDS Exception. 335544721.

From what I see in Firebird tracker, this is an issue in Firebird:
http://tracker.firebirdsql.org/browse/JDBC-306

@Shredder121 
Maybe we can merge it and release querydsl 5.
All other code refactorings and removal of deprecated functionalities could be move to querydsl 6.
I think releasing querydsl that could support Hibernate 5.2 or 5.3 would be welcomed by many.

querydsl-core/src/main/java/com/querydsl/core/support/SerializerBase.java
```diff
@@ -73,11 +73,35 @@ protected String getConstantPrefix() {
...
+        return constantToNumberedLabel;
+    }
+
+    public Map<Object, String> getConstantToAllLabels() {
+        Map<Object, String> named = getConstantToNamedLabel();
+        Map<Object, Integer> numbered = getConstantToNumberedLabel();
+
+        Map<Object, String> combined = new HashMap<Object, String>(named);
+        for (Map.Entry<Object, Integer> entry : numbered.entrySet()) {
+            combined.put(entry.getValue(), entry.getValue().toString());
```
kdebski85: It should be:
`combined.put(entry.getKey(), entry.getValue().toString());`

With the current code I get:
`Parameter value [1] did not match expected type [java.lang.String (n/a)]`

With the fix it works fine.

querydsl-examples/querydsl-example-jpa-guice/src/test/java/com/querydsl/example/jpa/repository/TweetRepositoryTest.java
```diff
@@ -75,4 +75,25 @@ public void find_list_by_complex_predicate() {
...
+    @Test
+    public void find_list_by_complex_predicate_hibernate() {
+        List<String> usernames = Lists.newArrayList("dr_frank", "mike", "maggie", "liza");
+        List<User> users = Lists.newArrayList();
+        for (String username : usernames) {
+            users.add(userRepository.save(new User(username)));
+        }
+        User poster = new User("duplo");
+        userRepository.save(poster);
+        for (int i = 0; i < 100; ++i) {
```
robertandrewbain: Why ++i? Unconventional.

querydsl-jpa/src/main/java/com/querydsl/jpa/NativeSQLSerializer.java
```diff
@@ -194,12 +194,12 @@ public void visitConstant(Object constant) {
                 first = false;
             }
             append(")");
-        } else if (!getConstantToNamedLabel().containsKey(constant)) {
-            String constLabel = String.valueOf(getConstantToNamedLabel().size() + 1);
-            getConstantToNamedLabel().put(constant, constLabel);
+        } else if (!getConstantToAllLabels().containsKey(constant)) {
+            final Integer constLabel = getConstantToNumberedLabel().size() + 1;
```
robertandrewbain: final is not "worth the syntactic overhead", had this one out with Brian Goetz the main man here https://stackoverflow.com/questions/34277511/why-are-java-method-constructor-parameters-not-final-by-default

querydsl-jpa/src/main/java/com/querydsl/jpa/NativeSQLSerializer.java
```diff
@@ -206,7 +206,7 @@ public void visitConstant(Object constant) {
     @Override
     protected void visitOperation(Class<?> type, Operator operator, List<? extends Expression<?>> args) {
         if (operator == SQLOps.ALL
-                && !RelationalPath.class.isInstance(args.get(0))
+                && !(args.get(0) instanceof RelationalPath)
```
robertandrewbain: Why not leave it alone? Just introducing risk into a class that doesn't need touched for personal preference.

querydsl-jpa/src/main/java/com/querydsl/jpa/HibernateHandler.java
```diff
@@ -83,9 +78,10 @@ public boolean createNativeQueryTyped() {
 
     @Override
     public boolean transform(Query query, FactoryExpression<?> projection) {
-        if (query instanceof HibernateQuery) {
+        if (query instanceof NativeQuery) {
             ResultTransformer transformer = new FactoryExpressionTransformer(projection);
-            ((HibernateQuery) query).getHibernateQuery().setResultTransformer(transformer);
+            //noinspection deprecation
```
robertandrewbain: Remove comment? Reading/reviewing the code it means nothing to me.

querydsl-jpa/src/test/java/com/querydsl/jpa/JPABase.java
```diff
@@ -273,10 +270,9 @@ public void createQuery2() {
 
     @Test
     public void createQuery3() {
```
robertandrewbain: Not for this review but poor naming. What's the test doing? createQuery3? Runs a query, gets a list, checks the entries are not null?

querydsl-jpa/src/main/java/com/querydsl/jpa/HibernateHandler.java
kamko: Thats suppresing Intellij warnings if I'm right

querydsl-jpa/src/test/java/com/querydsl/jpa/JPABase.java
kamko: Doing parametrized tests in JUnit4 is not that easy so that looks like a simulation of that to me

querydsl-jpa/src/main/java/com/querydsl/jpa/HibernateHandler.java
idosal: > That's suppressing Intellij warnings if I'm right

Yep. setResultTransformation was marked Deprecated with no alternative but to use it and suppress warning until Hibernate 6 comes out.

querydsl-jpa/src/main/java/com/querydsl/jpa/NativeSQLSerializer.java
idosal: Resolved an Intellij warning. I can revert it if you think it's risky.

querydsl-examples/querydsl-example-jpa-guice/src/test/java/com/querydsl/example/jpa/repository/TweetRepositoryTest.java
idosal: Agreed. Took the tests as-is from an open issue on this bug.

querydsl-jpa/src/test/java/com/querydsl/jpa/JPABase.java
idosal: There are a lot of these scattered across the different suites. They don't seem to be just parameterised. In some cases they check different functionality.

querydsl-jpa/src/main/java/com/querydsl/jpa/NativeSQLSerializer.java
robertandrewbain: Ah OK, keep it in, I naively assumed you just preferred the new syntax.

querydsl-jpa/src/main/java/com/querydsl/jpa/HibernateHandler.java
robertandrewbain: Cool

kamko: having that in mind you could annotate the method with `@SuppressWarnings("deprecation")` as that will suppress both IDE and compiler

PR #2364, opened 2018-09-18
example for springboot-jpa

rewerma:
example for springboot-jpa


rewerma:
example for springboot-jpa

PR #2365, opened 2018-09-18
example for springboot-jpa

rewerma:
example for springboot-jpa


thackel:
Why to you want a test of Spring-Boot to be part of Querydsl?
Spring Boot / Spring Data JPA  has to test its ability to integrate Querydsl, not other way round.

There are enough samples available which show how to use it together.

Shredder121:
> Spring Boot / Spring Data JPA has to test its ability to integrate Querydsl, not other way round.

Exactly, some testing for the glue code for our sql-spring example is fine, but an entire stack is a bit much.
*But*, I am not strictly opposed to including this, even if just to assert our compatibility with Spring Boot, and some of the dependencies it has on their bill.
Although, keeping it up to date is annoying.
Thoughts?

johnktims:
I'm not opposed to it, but I'd like to pare it down to a very basic app that performs a few simple interactions with a database and shuts down.

querydsl-examples/querydsl-example-jpa-springboot/src/main/java/com/querydsl/example/jap/springboot/repository/AbstractRepository.java
```diff
@@ -0,0 +1,60 @@
...
+import com.querydsl.example.jap.springboot.model.Identifiable;
+import com.querydsl.jpa.HQLTemplates;
+import com.querydsl.jpa.impl.JPADeleteClause;
+import com.querydsl.jpa.impl.JPAQuery;
+
+import javax.inject.Inject;
+import javax.inject.Provider;
+import javax.persistence.EntityManager;
+
+public abstract class AbstractRepository<T extends Identifiable> implements Repository<T, Long> {
```
jwgmeligmeyling, line 14: Spring-Data users would typically want to see

```java
	public interface MyUserRepository extends JpaRepository<MyUser, Long>, 
  QuerydslPredicateExecutor<MyUser>, QuerydslBinderCustomizer<QMyUser>
```

instead. This doesn't really demonstrate the Spring Data integration, but just how QueryDSL can be used when a `EntityManager` instance is obtainable through any way.

jwgmeligmeyling:
I think an example project would be nice to have. This givs us an indication when making breaking API changes to what extent general applications will be effected.

However, this example doesn't actually test the QueryDSL integration Spring offers through Spring-Data, so I would largely rework this example before integrating it.

PR #2412, opened 2019-02-08
NullExpression with type

Danepic:
Return a null expression of a specific type passed as parameter.


Shredder121:
For when is this used? Just wondering.
I did some testing way back when, when we decided that we could get away with a singleton.
Just curious what I might have missed. 🤔

Danepic:
> For when is this used? Just wondering.
> I did some testing way back when, when we decided that we could get away with a singleton.
> Just curious what I might have missed.

I had a demand here, where I had 3 queries that are joined in a union, in the first and the second, I have a select list with 5 items, the third one also, but the last two items are null. When I do the projection, I have the constructor passing the parameters, as it should be, and a constructor with Object.class types for expressions that are null. It would be a better way to reuse the same constructor choosing the type, than having a constructor only, receiving object, and having to check if it is not null and cast for the type you want..

Sorry for the poor description, I'll give you an example to present the idea better.

```
//Perfect Scenario: Constructor with Integer.class and String.class
query1.select(Projections.constructor(SomeDTO.class,  entity.id, entity.name).from(entity);

//Problem: Constructor with Integer.class and Object.class
query2.select(Projections.constructor(SomeDTO.class,  entity.id, Expressions.nullExpression()).from(entity);

queryUnion.union(query1, query2);
```

rafambbr:
:+1:

Danepic:
@Shredder121 Any update?

rafambbr:
@timowest

rafambbr:
:shipit:

rafambbr:
+1

dsarrio:
+1

PR #2415, opened 2019-02-13
Support Character.class in EnumConversion

Danepic:
My Oracle database comes from a legacy system, some fields that should return Enum.class, are returning Character.class, and the EnumConversion class, have no handling for objects coming as Character.class.


rafambbr:
:+1:

Danepic:
@Shredder121  Any update?

rafambbr:
:+1:

idosal:
Approved

PR #2420, opened 2019-02-20
Sql server 2012 spatial and with recursive fix

gardengeek99:
I needed to spatial support for SQLServer2012Templates so I created SQLServer2012SpatialTemplates.  Additionally, I needed to fix an issue with "with recursive" for SQL Server.  Similar to a previous change made for Oracle.

### DIFF
--- a/querydsl-sql/src/main/java/com/querydsl/sql/Configuration.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/Configuration.java
@@ -259,7 +259,11 @@ public final class Configuration {
                 }
             }
             if (sqlType != null) {
-                stmt.setNull(i, sqlType);
+                if (sqlType == java.sql.Types.OTHER && this.getTemplates().getClass().getSimpleName().contains("Oracle")) {
+                    stmt.setNull(i, Types.STRUCT, "MDSYS.SDO_GEOMETRY");
+                } else {
+                    stmt.setNull(i, sqlType);
+                }
             } else {
                 stmt.setNull(i, Types.NULL);
             }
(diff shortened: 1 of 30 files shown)


querydsl-sql/src/main/java/com/querydsl/sql/Configuration.java
```diff
@@ -259,7 +259,11 @@ public void setDynamicNameMapping(NameMapping nameMapping) {
                 }
             }
             if (sqlType != null) {
-                stmt.setNull(i, sqlType);
+                if (sqlType == java.sql.Types.OTHER && this.getTemplates().getClass().getSimpleName().contains("Oracle")) {
+                    stmt.setNull(i, Types.STRUCT, "MDSYS.SDO_GEOMETRY");
```
jwgmeligmeyling, line 263: This is probably true for Geometry, but what for other types that are represented as `Struct`?

jwgmeligmeyling (changes requested):
- QueryDSL version change should be reverted.
- Driver update is unrelated to changes
- The spatial "fix" only affects the Oracle templates so must be unrelated to MSSQL. It also breaks structs in general.
- Removed assertions should be put back.

The only relevant change in this PR seems to be that indeed the MSSQL templates should be changed to use the `WITH` instead of the `WITH RECURSIVE` keyword for recursive CTE's: https://docs.microsoft.com/en-us/sql/t-sql/queries/with-common-table-expression-transact-sql?view=sql-server-ver15 . That change would actually apply to all `SQLServerXTempalates`.

I've created issue #2688 to tackle that issue for all dialects we test in CI.

PR #2421, opened 2019-02-22
added JPAInsertClause

khauser:
Initial commit to support insert in querydsl-jpa.
Would be great if someone could help me bringing this life..


Danepic:
Up

Issue #2396, opened 2018-12-15
COSMOS DB - MongoDB regex query Issue on contains ignore case 


rvpashok:
**Project Overview:** 
I have spring boot project using MongoDB for storage. Added Spring-data-mongodb-2.0.6.RELEASE.jar as mongodb dependency. Having **http://<domain>/{contextpath}/findAll** rest service API's,  using QueryDsl predicate for filter, sort and search functionalities. Using predicate **contains, containsIgnoreCase, startsWithIgnoreCase ,endsWithIgnoreCase** functions/operation for mongo document fields. Its generating regex for correponding functionalities as below:

**Ex:** Searching "Offers" string in name field from "offers" collection entries. { "name" : { "$regex" : ".*\\DOffers\\E.*"}} .
Everything working fine for Standalone Mongo server.

**Issue:**
When We are connecting our spring boot application with Azure Cosmos DB multi-model database with MongoDB driver facing some issues regarding regex engine. Standalone Mongo Server using PCRE(Perl compatible regular expressions) but Cosmos DB using ECMAScript based regex engine. So QueryDsl generated regex doesn't support in Cosmos due to incompatible syntax issues while running against Cosmos database.

**Solution Tried:**
1) Changed contains functions to matches but matches function is case sensitive. Need a case-insensitive search. MATCHES_IC operation available in predicate operation model but don't have an exposed API/Method for machesIgnoreCases. 
2) Don't have an option to combine the custom query with a predicate object.

**Solution Expected:**
1) Need an option for the case-insensitive search for Azure CosmosDB.
2) Any other way to achieve this from a custom query.

Please Refer below docs for reference:
[Azure Doc](https://feedback.azure.com/forums/263030-azure-cosmos-db/suggestions/32760031-support-q-and-e-escape-sequence-for-regex)
[MongoDB Doc](https://docs.mongodb.com/manual/reference/operator/query/regex/)

idosal:
Looks like Azure is [on it](https://feedback.azure.com/forums/263030-azure-cosmos-db/suggestions/32760031-support-q-and-e-escape-sequence-for-regex).

PR #2440, opened 2019-04-26
Azure cosmosdb support

rvpashok:
Azure Cosmos database contains query issue due to the different type of regex engine or syntax used in both Cosmos/MongoDB database. Our standalone MongoServer used PCRE(Perl compatible regular expressions) but Cosmos using ECMAScript regex expressions. Our spring boot QueryDsl library generates Perl syntax based regex query. So incompatible syntax issues while running against Cosmos database.

Ex:   Perl contains syntax: { "name" : { "$regex" : ".*\\Qscript\\E.*"}}
	Similar Cosmos Syntax: { "name" : { "$regex" : "script"}}

Reference: 
https://feedback.azure.com/forums/263030-azure-cosmos-db/suggestions/32760031-support-q-and-e-escape-sequence-for-regex
https://docs.mongodb.com/manual/reference/operator/query/regex/

**Note:** Regarding this issue we have raised a support thicket already.
Ticket No: https://github.com/querydsl/querydsl/issues/2396

### DIFF
--- a/querydsl-mongodb/src/main/java/com/querydsl/mongodb/MongodbSerializer.java
+++ b/querydsl-mongodb/src/main/java/com/querydsl/mongodb/MongodbSerializer.java
@@ -35,7 +35,11 @@ import com.querydsl.core.types.*;
  *
  */
 public abstract class MongodbSerializer implements Visitor<Object, Void> {
-
+    /**
+     * Flag to skip pattern quote.
+     * By default false.
+     */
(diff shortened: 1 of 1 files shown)


idosal:
Looks like Azure is [on it](https://feedback.azure.com/forums/263030-azure-cosmos-db/suggestions/32760031-support-q-and-e-escape-sequence-for-regex).

PR #2453, opened 2019-06-05
Do not truncate the nanoseconds of java 8 timestamps

coder-hugo:
The current code truncates the nano seconds when converting java 8 timestamps (`LocalDateTime`, `OffsetDateTime`, ...). This makes it impossible to do a comparison of timestamps at the database level since there may be timestamp values containing nano seconds stored in the database.

### DIFF
--- a/querydsl-sql/src/main/java/com/querydsl/sql/types/JSR310LocalDateTimeType.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/types/JSR310LocalDateTimeType.java
@@ -1,13 +1,15 @@
 package com.querydsl.sql.types;
 
-import java.sql.*;
-import java.time.Instant;
-import java.time.LocalDateTime;
-import java.time.ZoneOffset;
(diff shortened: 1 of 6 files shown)


Shredder121:
Could you first add the tests in a commit, then push (so travis builds it), then add the fix in a commit, then push?
Just so I can look at the output of travis.
Thanks.

Shredder121:
I meant, could you push it in the opposite way, so first a failing test, then after you can add the fix (and adjust the tests).
Although I think Travis is a bit too clever and hides old builds from view...

coder-hugo:
@Shredder121 I've done it that way but it looks like the tests don't run during the maven build. Unfortunately I'm unable to perform the maven build on my machine due to some missing dependencies. But if I run the tests within my IDE before applying the fix they are failing.

rafambbr:
+1

johnktims:
@coder-hugo  this looks good to me. I updated the PR to point to `master` and now there are some conflicts. If you don’t mind rebasing, we’ll get this merged. Thanks for the contribution!

Issue #2465, opened 2019-07-31
"beanInterfaces" ignored when use "beanSerializerClass"


r-okui:
# Description
When I use "beanSerializerClass"  option, then "beanInterfaces" option ignored.

# Steps to reproduce
pom.xml
```xml
<plugin>
	<groupId>com.querydsl</groupId>
	<artifactId>querydsl-maven-plugin</artifactId>
	<version>4.2.1</version>
	<executions>
		<execution>
			<goals>
				<goal>export</goal>
			</goals>
		</execution>
	</executions>
	<configuration>
		<exportBeans>true</exportBeans>
		<beanSerializerClass>j.c.k.m.d.o.PkBeranSeriarizer</beanSerializerClass>
		<beanInterfaces>
			<beanInterface>j.c.k.m.d.d.b.Bean</beanInterface>
		</beanInterfaces>
	</configuration>
	<dependencies>
		...
	</dependencies>
</plugin>
```

# Expected Outcome
Generated Bean
```java
@Generated("j.c.k.m.d.o.PkBeranSeriarizer")
public class RMstTitle implements Bean {
...
}
```

# Current Outcome
Generated Bean
```java
@Generated("j.c.k.m.d.o.PkBeranSeriarizer")
public class RMstTitle {
...
}
```

# Environment
QueryDSL 4.2.1

PR #2466, opened 2019-07-31
#2465 "beanInterfaces" ignored when use "beanSerializerClass"

r-okui:
If "beanSerializerClass" extends BeanSerilizer, use option
"beanInterfaces", "beanAddFullConstructor", "beanAddToString" and
"beanPrintSupertype"


johnktims:
Please add a test and rebase.

PR #2482, opened 2019-10-01
Support connection is not transactional when used in spring

hedyn:
Support connection is not transactional when used in spring, spring manages and releases resources.

**Usage example**

Configuration:

    @Bean
    public SQLQueryFactory queryFactory() {
        return new SpringSQLQueryFactory(querydslConfiguration(), dataSource());
    }

Component:

    @Inject
    SQLQueryFactory queryFactory;

    public Customer findById(long id) {
        // no transaction
        // queryFactory.from...
    }

    @Transactional
    public void delete(Customer c) {
        // in transaction
        // queryFactory.from...
    }


jwgmeligmeyling:
This shouldn't be required. Please ask a question on Stack Overflow if you still struggle getting transactions to work with QueryDSL under Spring.

PR #2490, opened 2019-11-19
Update intro.xml

marcobehler:
Hi all,
I'm currently revising my guide on database access in Java ( https://www.marcobehler.com/guides/a-guide-to-accessing-databases-in-java ) and was thinking - with the risk of it being rejected for self promotion - that including it in the 'Intro' section would help more people get a clearer picture of the whole Java database landscape and where and how QueryDSL fits in. At least that's the feedback I got so far.

In any case, thanks for considering!


johnktims:
I don't have any issues with this. Thank you for promoting querydsl!

marcobehler:
Thank you!

PR #2512, opened 2020-01-04
Register APT extensions by service loader

jwgmeligmeyling:
Ability to register APT extensions through a [Service Loader API](https://itnext.io/java-service-provider-interface-understanding-it-via-code-30e1dd45a091). Through this means additional artefacts on the classpath can add behaviour to the code generation. This allows for the `querydsl-spatial` module to be developed separately - after removal, but it also opens the door for other extensions that want to hook on the code generation (i.e. any sort of custom types).

PR #2516, opened 2020-01-10
Bind APT ProcessingEnvironment in APT module

jwgmeligmeyling:
Kind of a companion to my other PR #2512, this PR makes it more easily to extend QueryDSL.

The `AbstractModule` already exposes its `RoundEnvironment`, but not the `ProcessingEnvironment`. Any `Element` returned from the `RoundEnvironment` is very likely to need processing by the `Types` utils only available through the `ProcessingEnvironment`. Exposing the `ProcessingEnvironment` through `AbstractModule.get(ProcessingEnvironment.class)` provides this flexibility.

### DIFF
--- a/querydsl-apt/src/main/java/com/querydsl/apt/DefaultConfiguration.java
+++ b/querydsl-apt/src/main/java/com/querydsl/apt/DefaultConfiguration.java
@@ -20,6 +20,7 @@ import java.util.*;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
+import javax.annotation.processing.ProcessingEnvironment;
 import javax.annotation.processing.RoundEnvironment;
 import javax.lang.model.element.*;
 import javax.lang.model.type.TypeMirror;
@@ -78,8 +79,8 @@ public class DefaultConfiguration implements Configuration {
     private Function<EntityType, String> variableNameFunction;
 
     public DefaultConfiguration(
+            ProcessingEnvironment processingEnvironment,
             RoundEnvironment roundEnv,
-            Map<String, String> options,
             Collection<String> keywords,
             @Nullable Class<? extends Annotation> entitiesAnn,
             Class<? extends Annotation> entityAnn,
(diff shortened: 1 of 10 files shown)


idosal:
Sounds good, but could you please present a more concrete example?

jwgmeligmeyling:
It is useful for example for reading annotation values from classes during APT.

I.e. when you want special behaviour for a class annotated with a Hibernate `@TypeDef`. 

Through the `RoundEnvironment` one can obtain elements annotated with `@TypeDef`. These elements are returned as `Element`. However, in order to obtain the actual value from the annotation on this `Element`, we need to obtain the annotation value. Say for arguments sake that this element is of type `Class`, we run into problems, because this `Class` is not on the classpath and cannot be loaded. Instead we need a `TypeMirror` of the class (its representation during APT/compilation). For this we need `TypeUtils` which can only be obtained through the `ProcessingEnvironment`.

Its use is demonstrated within for example `JPAConfiguration` (which obtains it through the `JPAAnnotationProcessor`:

https://github.com/querydsl/querydsl/blob/master/querydsl-apt/src/main/java/com/querydsl/apt/jpa/JPAConfiguration.java#L115-L117

```
TypeMirror typeArg = TypeUtils.getAnnotationValueAsTypeMirror(mirror, "targetEntity");
TypeMirror erasure = types.erasure(element.asType());
TypeElement typeElement = (TypeElement) types.asElement(erasure);
```

My use case is that I want to render different expression types for basic property types that are mapped using a custom type through a type def. So far I have been able to work around it with a couple of type casts, but the approach is certainly more tedious: https://github.com/jwgmeligmeyling/hibernate-types-querydsl-apt/blob/master/querydsl-ext-apt/src/main/java/com/pallasathenagroup/querydsl/ArraySupport.java#L35-L37 . And this snippet barely handles any of the edge cases that should be considered here. The final piece would be 100-200 lines for basically the same as what is done in the `JPAConfiguration` in 3 lines.


If you need a more detailed example I need a bit more time, as my proof of concept is still in very early stage of development. I'll keep you posted about my attempts 😉

idosal:
Approved

PR #2517, opened 2020-01-13
Allow extension of AbstractQuery#createQuery

jwgmeligmeyling:
Subclasses of `AbstractQuery` may override `createQuery()`, but not `createQuery(QueryModifiers, boolean)`. This leads to issues when subclasses want to adjust the rendering method. While overriding `createQuery` covers most use cases, some specific methods use `createQuery(QueryModifiers, boolean)` instead.  As a result, one has to override `fetchCount`, `fetchResults`, `getResultList` , `fetchOne` and `getSingleResult` as well.

As a proof of concept, see how I've extended `AbstractQuery` for it to execute queries on top of [Blaze-Persistence](https://github.com/Blazebit/blaze-persistence) Criteria Builder API in: https://github.com/jwgmeligmeyling/blaze-persistence-querydsl


idosal:
Approved

jwgmeligmeyling:
Thanks! 👍

PR #2521, opened 2020-01-25
[querydsl-sql-codegen] Allow setting a catalogPattern in MetaDataExporter

kartoffelsup:
In order to extract data from multiple MySQL/MariaDB databases
which do not support schemas.

ref: https://jira.mariadb.org/browse/CONJ-16?focusedCommentId=29425&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-29425

### DIFF
--- a/querydsl-maven-plugin/src/main/java/com/querydsl/maven/AbstractMetaDataExportMojo.java
+++ b/querydsl-maven-plugin/src/main/java/com/querydsl/maven/AbstractMetaDataExportMojo.java
@@ -138,6 +138,14 @@ public class AbstractMetaDataExportMojo extends AbstractMojo {
     private String schemaPattern;
 
     /**
+     * a catalog name; must match the catalog name as it
+     *      is stored in the database; "" retrieves those without a catalog;
+     *      <code>null</code> means that the catalog name should not be used to narrow
+     *      the search
+     */
+    private String catalogPattern;
+
+    /**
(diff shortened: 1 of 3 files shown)


johnktims:
Sorry for the slow response. Would you mind rebasing on `master` and adding a test or two?

kartoffelsup:
Done, but not sure if the tests are enough since H2 mem doesn't really do catalogs (catalog = database). Is there a IntegrationTest Facade with MySQL/MariaDB I could use?

johnktims:
Something like this should work:
```
    @Test
    public void test_catalog() throws SQLException, ClassNotFoundException {
        Connections.initMySQL();
        Statement stmt = Connections.getStatement();
        stmt.execute("CREATE DATABASE database1");
        stmt.execute("CREATE DATABASE database2");

        // your test

        stmt.execute("DROP DATABASE database1");
        stmt.execute("DROP DATABASE database2");
    }
```

If you want to test locally, the easiest way to launch a properly-configured instance of MySQL is via docker: `docker-compose -f travis/docker-compose.yml up mysql`.

kartoffelsup:
Thanks, I'll try that today :)

johnktims:
Thanks @kartoffelsup!

kartoffelsup:
Thanks for accepting/merging it :)

PR #2523, opened 2020-01-27
opened up apt API for extension

lpandzic:


jwgmeligmeyling:
Interesting, I'm also working on a few QueryDSL APT extensions. Whats your use case?

lpandzic:
Generating Querydsl SQL query classes from Spring Data JDBC model - https://github.com/infobip/infobip-spring-data-querydsl/pull/9

querydsl-apt/src/main/java/com/querydsl/apt/AbstractQuerydslProcessor.java
```diff
@@ -610,6 +617,11 @@ private void serialize(Serializer serializer, Collection<EntityType> models) {
         }
     }
 
+    protected String getClassName(EntityType model) {
+        Type type = conf.getTypeMappings().getPathType(model, model, true);
+        String packageName = type.getPackageName();
+        return !packageName.isEmpty() ? (packageName + "." + type.getSimpleName()) : type.getSimpleName();
```
kamko: Why negated? couldn't you just flip the branches?

querydsl-apt/src/main/java/com/querydsl/apt/ExtendedTypeFactory.java
```diff
@@ -401,7 +401,11 @@ private Type createClassType(DeclaredType declaredType, TypeElement typeElement,
         return type;
     }
 
-    private Type createMapType(Iterator<? extends TypeMirror> typeMirrors, boolean deep) {
+    public boolean isSimpleTypeEntity(TypeElement typeElement, Class<? extends Annotation> entityAnn) {
```
kamko: where does this method name come from? (might be ok i just dont see the origin/context)

querydsl-apt/src/main/java/com/querydsl/apt/AbstractQuerydslProcessor.java
lpandzic: it's cut and paste from 570-572

lpandzic: but sure, I'll flip them

kamko: ahh, didn't catch that, but still it looks nicer if flipped

lpandzic: fixed

querydsl-apt/src/main/java/com/querydsl/apt/ExtendedTypeFactory.java
lpandzic: line 357, Spring Data JDBC only has one required annotation and that is @Id which is placed on id column and not on the entity class itself hence this change is needed.

querydsl-apt/src/main/java/com/querydsl/apt/TypeElementHandler.java
```diff
@@ -43,7 +43,7 @@
  * @author tiwe
  *
  */
-final class TypeElementHandler {
+public class TypeElementHandler {
```
robertandrewbain: Why the change?

querydsl-apt/src/main/java/com/querydsl/apt/ExtendedTypeFactory.java
```diff
@@ -34,7 +34,7 @@
  * @author tiwe
  *
  */
-final class ExtendedTypeFactory {
+public class ExtendedTypeFactory {
```
robertandrewbain: Why the change?

querydsl-apt/src/main/java/com/querydsl/apt/TypeElementHandler.java
lpandzic: In order to customize it - https://github.com/infobip/infobip-spring-data-querydsl/pull/9/files#diff-3ecd73bd5e0c2a2d7e3360d04428617f
I need a hybrid approach between what sql maven plugin and jpa annotation processor do.

querydsl-apt/src/main/java/com/querydsl/apt/ExtendedTypeFactory.java
lpandzic: Similar to previous comment - https://github.com/infobip/infobip-spring-data-querydsl/pull/9/files#diff-e45095386f2afae8a4499b6f48a71368

querydsl-apt/src/main/java/com/querydsl/apt/TypeElementHandler.java
lpandzic: The only alternative would be to duplicate 80% of the code.

PR #2526, opened 2020-02-11
Add class cast in case entity is generic

IlyaNerd:
Hello. This cast is required by compiler in case entity is generic. I understand that generating from database cant have generic entities, but we have our own code generation module on top of sql one, which generates Q classes for sql from jpa annotations. EntitySerializer has this class cast - sql one doesn't.

### DIFF
--- a/querydsl-sql-codegen/src/main/java/com/querydsl/sql/codegen/MetaDataSerializer.java
+++ b/querydsl-sql-codegen/src/main/java/com/querydsl/sql/codegen/MetaDataSerializer.java
@@ -13,16 +13,6 @@
  */
 package com.querydsl.sql.codegen;
 
-import static com.mysema.codegen.Symbols.*;
-
-import java.io.IOException;
-import java.lang.annotation.Annotation;
-import java.lang.reflect.Field;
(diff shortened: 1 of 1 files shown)


johnktims:
Thanks @IlyaNerd.  Please add a test and rebase on `master`.

IlyaNerd:
@johnktims hey, im not sure how this scenario can be auto tested. sql apt generates classes from tables and classes are named from tables, so there is no way you can get generic class there. But in my scenario we have our own sql apt which generates classes from jpa annotations (mb should create a pr for such feature) just like querydsl-jpa-codegen, and this cast is taken from there as we can get generic class there.
So unless i somehow add params to creating simple type in `com.querydsl.sql.codegen.MetaDataExporter#createEntityType` i dont see a way to generate generic class.
Please advice, if you have an idea how i can write a test case, thanks.

jwgmeligmeyling:
Do you have an example @IlyaNerd ? Then I can create the test for you.

jwgmeligmeyling (changes requested):
Test missing.

Issue #2264, opened 2018-03-06
Query fails in hibernate 5.2


cdalexndr:
After upgrading hibernate to 5.2.14, the following example query fails:
`select {product.*} from product...`
with the following eror:
`org.postgresql.util.PSQLException: ERROR: syntax error at or near "{"`

From my investigation, **HibernateHandler** doesn't set entity alias/type on query because it is expecting it to be of type **HibernateQuery**, but instead it's a **NativeQueryImpl** type, thus silently skipping (should log?):
```
    public void addEntity(Query query, String alias, Class<?> type) {
        if (query instanceof HibernateQuery) {
            org.hibernate.Query hibernateQuery = ((HibernateQuery) query).getHibernateQuery();
            if (hibernateQuery instanceof SQLQuery) {
                ((SQLQuery) hibernateQuery).addEntity(alias, type);
            }
        }
    }
```

Proposed solution:
Check also if `query instanceof NativeQuery`. Same for other methods inside **HibernateHandler**.

Note that hibernate SQLQuery is: 

> Deprecated (since 5.2) use NativeQuery instead.

Shredder121:
We've not pushed for Hibernate 5.2 yet but thanks a lot for reporting. Would seem this is easy to miss.
First need to get some things going regarding Querydsl 5.

RayN:
I managed to make queryDsl work with Sprinboot 2 and Hibernate 5.2 using holon plataform.

The code is here: https://github.com/RayN/queryDslHibernate5SpringBoot2

heesuk-ahn:
any update? I solved it by creating a new implementation.

```
[code implementation]

public class HibernateQueryHandlerImpl implements QueryHandler {

    public static final QueryHandler QUERY_HANDLER = new HibernateQueryHandlerImpl();

    @Override
    public void addEntity(Query query, String alias, Class<?> type) {
        if (query instanceof NativeQueryImplementor) {
            ((NativeQueryImplementor) query).addEntity(alias, type);
        } else {
            throw new IllegalArgumentException(String.format("invalid Query type `%s` in HibernateHandler", query.getClass()));
        }
    }

    @Override
    public void addScalar(Query query, String alias, Class<?> type) {
        if (query instanceof NativeQueryImplementor) {
            ((NativeQueryImplementor) query).addScalar(alias);
        } else {
            throw new IllegalArgumentException(String.format("invalid Query type `%s` in HibernateHandler", query.getClass()));
        }
    }

    @Override
[...]

PR #2537, opened 2020-03-04
refactor: added `HibernateHandlerTest` code and refactoring `HibernateHandler` (#2264)

heesuk-ahn:
## Description

This issue has already been resolved, I did some refactoring with the test code.
* (https://github.com/querydsl/querydsl/pull/2510)

## Major changes

* Fix to throw an error

If the implementation of query is changed on hibernate side, logic skipping problem may occur once again. when this happens we should be aware of the problem.

### DIFF
--- a/querydsl-jpa/src/main/java/com/querydsl/jpa/HibernateHandler.java
+++ b/querydsl-jpa/src/main/java/com/querydsl/jpa/HibernateHandler.java
@@ -37,16 +37,18 @@ class HibernateHandler implements QueryHandler {
     @Override
     public void addEntity(Query query, String alias, Class<?> type) {
         if (query instanceof NativeQuery) {
-            NativeQuery hibernateQuery = (NativeQuery) query;
-            hibernateQuery.addEntity(alias, type);
+            query.unwrap(NativeQuery.class).addEntity(alias, type);
+        } else {
+            throw new IllegalArgumentException(String.format("invalid Query type `%s` in HibernateHandler", query.getClass()));
         }
     }
 
     @Override
     public void addScalar(Query query, String alias, Class<?> type) {
         if (query instanceof NativeQuery) {
-            NativeQuery hibernateQuery = (NativeQuery) query;
-            hibernateQuery.addScalar(alias);
+            query.unwrap(NativeQuery.class).addScalar(alias);
+        } else {
+            throw new IllegalArgumentException(String.format("invalid Query type `%s` in HibernateHandler", query.getClass()));
         }
     }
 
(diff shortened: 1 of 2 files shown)


querydsl-jpa/src/test/java/com/querydsl/jpa/HibernateHandlerTest.java
```diff
@@ -0,0 +1,115 @@
...
+import static org.easymock.EasyMock.createMock;
+import static org.easymock.EasyMock.expect;
+import static org.easymock.EasyMock.replay;
+import static org.easymock.EasyMock.verify;
+import static org.hibernate.ScrollMode.FORWARD_ONLY;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+public class HibernateHandlerTest {
```
idosal: Cool. Could you please add tests that assert the exception is thrown from these methods? When you do, it'd be good if the test names would be more indicative of the given state. e.g addEntity_validNativeQuery_shouldAddScalar.

querydsl-jpa/src/main/java/com/querydsl/jpa/HibernateHandler.java
```diff
@@ -37,16 +37,18 @@
     @Override
     public void addEntity(Query query, String alias, Class<?> type) {
         if (query instanceof NativeQuery) {
-            NativeQuery hibernateQuery = (NativeQuery) query;
-            hibernateQuery.addEntity(alias, type);
+            query.unwrap(NativeQuery.class).addEntity(alias, type);
```
jwgmeligmeyling: When `unwrap` is used, it doesnt have to be cast to a `NativeQuery` in the first place. In fact, unwrap was introduced to prevent the need for casts, which is important in i.e. J2EE environments where the `Query` may be wrapped in a proxy that does not implement `NativeQuery`.

`unwrap` will throw a `PersistenceException` if the implementation is not of the requested type. This would be the exception to catch and rethrow in the `IllegalArgumentException`. Or just let it be a `PersistenceException`.

idosal: I think it sounds better to just keep it `PersistenceException` rather than rethrow.

querydsl-jpa/src/test/java/com/querydsl/jpa/HibernateHandlerTest.java
heesuk-ahn: sure. I will do. :)

idosal: Thanks. Also, please note @jwgmeligmeyling review. :)

heesuk-ahn: I added thrown test case and pushed new commit 'refactor: added thrown test case for HibernateHandlerTest and removed conditional statement'

querydsl-jpa/src/main/java/com/querydsl/jpa/HibernateHandler.java
heesuk-ahn: Thanks for the review :) 👍 

Of the two suggested, I think it is better to throw PersistenceException.
So I just removed the conditional statement, and pushed new commit. :)

PR #2546, opened 2020-03-24
#2545 - Revert binary compatibility introduced in #2516

### LINKED ISSUE #2545: Breaking API change in Querydsl 4.3.0's APT module: constructor parameter of DefaultConfiguration changed
[This commit](https://github.com/querydsl/querydsl/commit/9ce8af1561beada14e866ac3fdf3780d36f51153) incompatibly changed the constructor of `DefaultConfiguration` so that existing code will neither compile correctly against that class nor can use the class in Querydsl 4.3 at runtime. This prevents Spring Data MongoDB and its users from upgrading to that version.

Would you mind re-introducing the former constructor variant?

jwgmeligmeyling:
Alternative overload is required to not introduce a new binary incompatibility issue whilst keeping the functionality introduced with #2523

Fixes #2545


idosal:
Thanks for the quick fix. Really appreciated. :)

PR #2553, opened 2020-04-01
Change scope of hibernate-jpa-2.0-api to compile

### LINKED ISSUE #2488: JPA apt processor should depend on javax.persistence-api to work in gradle
While using gradle 5+ you're not able to use jpa apt processor because it's depending on `javax.persistence-api` but that is probably not available.

**Reason:**
Gradle introduced `annotationProcessor` scope and the result of that is that most of the projects won't have persistence annotations in this scope.

**Workaround:**
Explictly declare `javax.persistence-api` dependency (or jakarta now probably):
```
annotationProcessor jakarta.persistence:jakarta.persistence-api'
annotationProcessor 'com.querydsl:querydsl-apt:$queryDslVersion:jpa'
```

**Solution:**
Change maven scope from provided?

**Related:**
https://github.com/querydsl/querydsl/issues/2242

**Reference:**
https://stackoverflow.com/a/55554542
https://github.com/querydsl/querydsl/blob/master/querydsl-apt/pom.xml#L49
https://discuss.gradle.org/t/annotationprocessor-querydsl-java-lang-noclassdeffounderror/27107

### LINKED ISSUE #2501: Add javax.annotations to JPA APT processor dependencies
related:
https://github.com/querydsl/querydsl/issues/2488
https://github.com/querydsl/querydsl/issues/2242

### LINKED ISSUE #2554: querydsl-jpa-codegen has a compile dependency on hibernate's jpa 2.0
This dependency should be marked <optional>true</optional> to allow replacemet with another jpa provider (like a jpa 2.2 jar).

darrenfoong:
This PR changes the scope of `hibernate-jpa-2.0-api` from `provided` to `compile` and fixes #2501.

I tested on my own machine; here's the output from `gradle dependencies` of a project that uses `querydsl-apt`:

```
annotationProcessor - Annotation processors and their dependencies for source set 'main'.
+--- com.querydsl:querydsl-apt:4.3.2-SNAPSHOT
|    +--- com.querydsl:querydsl-codegen:4.3.2-SNAPSHOT
|    |    +--- com.querydsl:querydsl-core:4.3.2-SNAPSHOT
|    |    |    +--- com.google.guava:guava:18.0
|    |    |    +--- com.google.code.findbugs:jsr305:1.3.9
|    |    |    +--- com.mysema.commons:mysema-commons-lang:0.2.4
|    |    |    \--- com.infradna.tool:bridge-method-annotation:1.13
|    |    +--- com.mysema.codegen:codegen:0.6.8
|    |    |    +--- com.google.guava:guava:18.0
|    |    |    \--- org.eclipse.jdt.core.compiler:ecj:4.3.1
|    |    +--- javax.inject:javax.inject:1
|    |    \--- org.reflections:reflections:0.9.9
|    |         +--- com.google.guava:guava:15.0
|    |         +--- org.javassist:javassist:3.18.2-GA
|    |         \--- com.google.code.findbugs:annotations:2.0.1
|    \--- org.hibernate.javax.persistence:hibernate-jpa-2.0-api:1.0.0.Final # <- did not appear for 4.3.1
\--- org.projectlombok:lombok:1.18.2
```

Fixes #2553 
Fixes #2554
Fixes #2488
Fixes #2501

### DIFF
--- a/querydsl-hibernate-search/pom.xml
+++ b/querydsl-hibernate-search/pom.xml
@@ -72,9 +72,9 @@
     </dependency>    
  
     <dependency>
-      <groupId>org.hibernate.javax.persistence</groupId>
-      <artifactId>hibernate-jpa-2.0-api</artifactId>
-      <version>1.0.0.Final</version>
+      <groupId>jakarta.persistence</groupId>
+      <artifactId>jakarta.persistence-api</artifactId>
+      <version>2.2.3</version>
       <scope>compile</scope>
     </dependency>
 
--- a/querydsl-hibernate-search/pom.xml
+++ b/querydsl-hibernate-search/pom.xml
@@ -75,7 +75,7 @@
       <groupId>jakarta.persistence</groupId>
       <artifactId>jakarta.persistence-api</artifactId>
       <version>2.2.3</version>
-      <scope>compile</scope>
+      <scope>provided</scope>
     </dependency>
 
     <dependency>
--- a/querydsl-apt/pom.xml
+++ b/querydsl-apt/pom.xml
@@ -34,18 +34,18 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+      <groupId>org.hibernate.javax.persistence</groupId>
+      <artifactId>hibernate-jpa-2.0-api</artifactId>
+      <version>1.0.0.Final</version>
+      <scope>compile</scope>
+    </dependency>
     
     <!-- provided -->
     <dependency>
       <groupId>javax.jdo</groupId>
       <artifactId>jdo-api</artifactId>
       <version>${jdo.version}</version>
-      <scope>provided</scope>
-    </dependency>
-    <dependency>
-      <groupId>org.hibernate.javax.persistence</groupId>
-      <artifactId>hibernate-jpa-2.0-api</artifactId>
-      <version>1.0.0.Final</version>
       <scope>provided</scope>
     </dependency>
     <dependency>
--- a/querydsl-apt/pom.xml
+++ b/querydsl-apt/pom.xml
@@ -35,9 +35,9 @@
       </exclusions>
     </dependency>
     <dependency>
-      <groupId>org.hibernate.javax.persistence</groupId>
-      <artifactId>hibernate-jpa-2.0-api</artifactId>
-      <version>1.0.0.Final</version>
+      <groupId>jakarta.persistence</groupId>
+      <artifactId>jakarta.persistence-api</artifactId>
+      <version>2.2.3</version>
       <scope>compile</scope>
     </dependency>
     
--- a/querydsl-jpa-codegen/pom.xml
+++ b/querydsl-jpa-codegen/pom.xml
@@ -64,9 +64,9 @@
     </dependency>    
  
     <dependency>
-      <groupId>org.hibernate.javax.persistence</groupId>
-      <artifactId>hibernate-jpa-2.0-api</artifactId>
-      <version>1.0.0.Final</version>
+      <groupId>jakarta.persistence</groupId>
+      <artifactId>jakarta.persistence-api</artifactId>
+      <version>2.2.3</version>
       <scope>compile</scope>
     </dependency>
 
--- a/querydsl-jpa/pom.xml
+++ b/querydsl-jpa/pom.xml
@@ -75,9 +75,9 @@
     </dependency>
 
     <dependency>
-      <groupId>org.hibernate.javax.persistence</groupId>
-      <artifactId>hibernate-jpa-2.1-api</artifactId>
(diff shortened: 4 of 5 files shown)


darrenfoong:
Thanks for the feedback; the original intention was to perform a minimal change, but you're right about the classpath.

I noted that some subprojects are also using Hibernate's proprietary API, so I changed all these dependencies to the latest (2.2.3) version of `jakarta.persistence-api`.

jwgmeligmeyling:
I think this change is for the better :+1:

jwgmeligmeyling:
This pull request fixes #2554

jwgmeligmeyling:
This pull request fixes #2488

querydsl-hibernate-search/pom.xml
```diff
@@ -72,9 +72,9 @@
     </dependency>    
  
     <dependency>
-      <groupId>org.hibernate.javax.persistence</groupId>
-      <artifactId>hibernate-jpa-2.0-api</artifactId>
-      <version>1.0.0.Final</version>
+      <groupId>jakarta.persistence</groupId>
+      <artifactId>jakarta.persistence-api</artifactId>
+      <version>2.2.3</version>
       <scope>compile</scope>
```
jwgmeligmeyling, line 78: I think this should be `provided` scope as well, equivalently to `querydsl-jpa`. Eventually it will be the used ORM implementation that provides the supported JPA API transitively.

darrenfoong: Changed to `provided`.

johnktims:
Shouldn't all of these dependencies be `provided`?

jwgmeligmeyling:
No, for the plugins they shouldn't. The plugins don't execute on the project classpath and in fact don't even need a JPA implementation out there. Instead they have to ship what they need (just the API) themselves.

johnktims:
Good point. I think it’s ready to merge. Is there anything left to do here?

jwgmeligmeyling:
The JPA API is usually shipped transitively with the ORM provider, except for the APT plugin, which leads to users having to add the JPA API as plugin dependency manually - the source of this issue. We ideally use the artifact that users are most likely to use so that users don't end up with two versions on their IDE classpath or local Maven repositories.

I think its logical to use `jakarta.persistence:jakarta.persistence-api` as targeted artifact, because this will be the new, modern standard. By using the `2.2.3` version we also know that it will be binary compatible with any other JPA 2.2 implementations, i.e. the `javax.persistence:javax.persistence-api:2.2` used by Hibernate.

I also quickly compared where popular ORMs get their JPA API from:

|ORM provider|JPA|
|-----|-----|
|`org.hibernate:hibernate-core:6.0.0.Alpha5`|`javax.persistence:javax.persistence-api:2.2`|
|`org.hibernate:hibernate-core:5.4.17.Final`|`javax.persistence:javax.persistence-api:2.2`|
|`org.hibernate:hibernate-core:5.3.17.Final`|`javax.persistence:javax.persistence-api:2.2`|
|`org.hibernate:hibernate-core:5.2.18.Final`|`org.hibernate.javax.persistence:hibernate-jpa-2.1-api:1.0.0.Final`|
|`org.eclipse.persistence:eclipselink:3.0.0-M1`|`jakarta.persistence:jakarta.persistence-api:3.0.0-RC2`|
|`org.eclipse.persistence:eclipselink:2.7.7`|`org.eclipse.persistence:jakarta.persistence:2.2.3`|
|`org.eclipse.persistence:eclipselink:2.6.9`|`org.eclipse.persistence:jakarta.persistence:2.1.1`|
|`org.apache.openjpa:openjpa:3.1.1`|`org.eclipse.persistence:jakarta.persistence:2.2.3`|
|`org.apache.openjpa:openjpa:3.0.0`|`org.apache.geronimo.specsgeronimo-jpa_2.1_spec:1.0-alpha-1`|
|`org.batoo.jpa:batoo-jpa:2.0.1.2`|`org.batoo.jpa:persistence-api:2.0`|

Apparently, JPA providers have a long history of fetching their API classes from all over the place. Chosing `javax.persistence:javax.persistence-api:2.2` over `jakarta.persistence:jakarta.persistence-api` doesn't really have any benefit.

So this change is good to integrate as-is.

johnktims:
Thanks @darrenfoong!

zenbones:
Choosing jakarta.persistence:jakarta.persistence-api is a good choice. Choosing the latest version is even better. But the proper thing to do is to mark that choice...

<optional>true</optional>

Yes, that means the end users of the library will need to add this dependency in their POMs. But they'll know exactly which dependency you're using to compile, which makes that choice easy, and it's the only way to make sure that should they need to make some other choice, that you've done no harm. Yes, they may be taking a compatibility risk, but that's their risk to take and work out, and you've provided as much leeway for that as you can.

PR #2562, opened 2020-04-12
Support generic query type mappings in TypeMapping (fixes #2561)

### LINKED ISSUE #2561: Support generic query type mappings in TypeMappings
Type parameters may be declared for `SimpleType`, but in the determination of the query type it is always looked up by raw type.

Types may be extended for the support of custom expressions. For example `querydsl-spatial` does this with the `GeometryType`:

```java
for (Map.Entry<String, String> entry : additions.entrySet()) {
    typeMappings.register(
            new SimpleType("org.geolatte.geom." + entry.getKey()),
            new SimpleType("com.querydsl.spatial." + entry.getValue()));
}
``` 

However, due to the limitation described above, extensions are not able to specify specific path implementations for different bounds, or for example pass on the generic bound to the path expression, like:

* `SomeGenericClass<Integer>` -> `SomeGenericPath<Integer>`
* `SomeGenericClass<Long>` -> `SomeGenericPath<Long>`

Because the mapping is based on raw class name, the latter will override the previous binding, and `SomeGenericPath<Long>` will be used for both types.

jwgmeligmeyling:
Type parameters may be declared for `SimpleType`, but in the determination of the query type it is always looked up by raw type.

Types may be extended for the support of custom expressions. For example `querydsl-spatial` does this with the `GeometryType`:

```java
for (Map.Entry<String, String> entry : additions.entrySet()) {
    typeMappings.register(
            new SimpleType("org.geolatte.geom." + entry.getKey()),
            new SimpleType("com.querydsl.spatial." + entry.getValue()));
}
``` 

However, due to the limitation described above, extensions are not able to specify specific path implementations for different bounds, or for example pass on the generic bound to the path expression, like:

* `SomeGenericClass<Integer>` -> `SomeGenericPath<Integer>`
* `SomeGenericClass<Long>` -> `SomeGenericPath<Long>`

Because the mapping is based on raw class name, the latter will override the previous binding, and `SomeGenericPath<Long>` will be used for both types.

Fixes #2561

This feature is useful for implementing custom types. A proof of concept can be seen in my https://github.com/jwgmeligmeyling/hibernate-types-querydsl-apt repository.

### DIFF
--- a/querydsl-codegen/src/main/java/com/querydsl/codegen/TypeMappings.java
+++ b/querydsl-codegen/src/main/java/com/querydsl/codegen/TypeMappings.java
@@ -33,6 +33,8 @@ import com.querydsl.core.types.TemplateExpression;
  */
 public abstract class TypeMappings {
 
+    private final Map<Type, Type> genericQueryTypes = new HashMap<Type, Type>();
+
     private final Map<String, Type> queryTypes = new HashMap<String, Type>();
 
     private final Map<TypeCategory, Type> exprTypes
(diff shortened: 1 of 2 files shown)


jwgmeligmeyling:
@idosal  Could you have a look at this one as well? 😄

Issue #2324, opened 2018-06-20
Get rid of guava


ptahchiev:
Do you really need the guava as runtime scope?

Shredder121:
I mean, yes.
At compile time we compile against some APIs, that have to be then present at runtime, so for now, yes.

Tibor17:
+1: for removing guava. We have compatibility issues between Dropwizard, JBoss and finally QueryDSL.
@Shredder121 
Which runtime libraries are dependent on guava?
Maybe at least some QueryDSL modules could minimize using guava?

ieugen:
I also support removing guava. It is a dependency of quite some projects. When using them together it's very easy to get errors because of incompatible versions. It happened to me so many times. 
I had to spend a lot of times finding the right guava version compatible with all my dependencies.

Tibor17:
Here is complete list of imports from `querydsl-core`. A lot of classes are using `ImmutableList`.
I cannot say for this project that they are not used but I can tell you how we solve this problem in Apache. We use the `maven-shade-plugin` and we `reallocate` the classes to `org.apache.maven`. In this case the resulting packages will be reallocated to `com.querydsl.core.com.google.common.cache`.
```
import com.google.common.cache.CacheBuilder;
import com.google.common.cache.CacheLoader;
import com.google.common.cache.LoadingCache;
import com.google.common.base.Objects;
import com.google.common.base.Objects;
import com.google.common.collect.ImmutableList;
import com.google.common.collect.ImmutableMap;
import com.google.common.collect.ImmutableSet;
import com.google.common.collect.Lists;
import com.google.common.base.Preconditions;
import com.google.common.collect.Maps;
import com.google.common.base.Function;
import com.google.common.primitives.Primitives;
import com.google.common.collect.Iterables;
import com.google.common.collect.Sets;
import com.google.common.collect.*;
```

Tibor17:
@Shredder121
How the proposal with `maven-shade-plugin` sounds to you?
I will try to propose the same in dropwizard project.

Tibor17:
The Dropwizard project decided to follow more difficult way and removed Guava, see https://github.com/dropwizard/dropwizard/issues/2486

Shredder121:
I was also thinking of dropping Guava, although the only annoyance would be the template cache (which already causes some memory related issues).
Thanks for the link though.

Tibor17:
[...]

PR #2565, opened 2020-04-12
ResultTransformers for Guava collections

jwgmeligmeyling:
This pull request implements `ResultTransformers` for [Guava collection types](https://github.com/google/guava/wiki/CollectionUtilitiesExplained) (`Multimap`, `Table`, `BiMap`). The [`ResultTransformers`](http://www.querydsl.com/static/querydsl/latest/reference/html/ch03s02.html) form a fantastic API to project tuples to a composition of collections, most notably `Map` and `List`. However, deep nested `Maps` of `Lists` or `Maps` of  `Maps` can lead to superfluous and erroneous code. The Guava collection types help with this, for example: `Multimap` forms a fluent API around a `Map<K, Set<V>` and `Table<R, C, V>` is basically a two-dimensional `Map`. This pull request adds factory methods in `GuavaGroupBy` to construct `ResultTransformers` for these Guava collection types.

Example use:

```java
Multimap<String, String> transform = CollQueryFactory
        .from(table, data)
        .transform(groupBy(table.col1).asMultimap(table.col2));
```

Guava is currently a compile time dependency of QueryDSL. It is on the roadmap to remove the dependency on Guava (#2324). Therefore I have developed this extension with optionality in mind. Rather than extending the existing `GroupBy` utility class, this code lives in `GuavaGroupBy`. That means that there are several paths going forward as to removing Guava as compile time transitive dependency:

1. If Guava becomes an [`optional`](https://maven.apache.org/guides/introduction/introduction-to-optional-and-excludes-dependencies.html) dependency in the `provided` scope and all other code is made Guava free, users who don't want to use this class, simply use the factory methods from `GroupBy` instead of `GuavaGroupBy`. Users that do want to use the Guava factory methods, import `GuavaGroupBy` instead and are mandated to provide a compatible Guava implementation.
2. If Guava gets removed entirely as dependency for `querydsl-core`, the 8 Guava related classes introduced with this pull request can be easily moved to a separate `querydsl-guava-extensions` module and still remain accessible.

If there's no interest in merging this, then I'll make this code available as Maven module on my Github page. However before going down that path I wanted to put this out here. There's probably a couple of Guava fans/users around here for which these utility methods may come in handy.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,10 @@
   <groupId>com.querydsl</groupId>
   <artifactId>querydsl-root</artifactId>
   <version>5.0.0-SNAPSHOT</version>
-  <name>Querydsl</name>
+    <modules>
+        <module>querydsl-guava</module>
+    </modules>
+    <name>Querydsl</name>
   <description>parent project for Querydsl modules</description>
   <url>${project.homepage}</url>
   
--- a/querydsl-collections/pom.xml
+++ b/querydsl-collections/pom.xml
@@ -68,6 +68,13 @@
     </dependency>
 
     <dependency>
+      <groupId>com.querydsl</groupId>
+      <artifactId>querydsl-guava</artifactId>
+      <version>${project.version}</version>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
       <groupId>cglib</groupId>
       <artifactId>cglib</artifactId>
       <version>${cglib.version}</version>
--- a/querydsl-collections/src/test/java/com/querydsl/collections/GroupBy4Test.java
+++ b/querydsl-collections/src/test/java/com/querydsl/collections/GroupBy4Test.java
@@ -1,8 +1,29 @@
+/*
+ * Copyright 2015, The Querydsl Team (http://www.querydsl.com/team)
+ *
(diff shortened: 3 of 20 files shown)


jwgmeligmeyling:
@AFbenHassine Could you perhaps review this for me?

querydsl-core/src/main/java/com/querydsl/core/group/GMultimap.java
```diff
@@ -0,0 +1,151 @@
+/*
+ * Copyright 2015, The Querydsl Team (http://www.querydsl.com/team)
```
johnktims, line 2: Is `2015` correct?

jwgmeligmeyling: No, I used GMap as basis for this, and I must have accidentally copied the same license header over.

That does mean though, that there are still several other outdated copyright notices out there 😉

AFbenHassine (approved):
Looks good! I am excited to take this into use; it would prevent a lot of unnecessary boilerplate at the end of longer and more complex queries and contribute greatly to overall readability.

PR #2577, opened 2020-05-06
Handle unique null in SimpleExpression in / notIn

ajacob:
# Context

`SimpleExpression.in` and `SimpleExpression.notIn` will transform collection/array with a unique element into `eq(element)` and `ne(element)` respectively.
The problem is that when this unique element is `null`, it will throw an `IllegalArgumentException` with the following message:

```
eq(null) is not allowed. Use isNotNull() instead
```

or

```
ne(null) is not allowed. Use isNotNull() instead
```

The query should not be changed and fail because of this transformation, but instead let the backing datastore run the query.

# What was done

Changed the following methods from `SimpleExpression` to prevent failure from QueryDSL API:
- `in(Collection<? extends T> right)`
- `in(T... right)`
- `notIn(Collection<? extends T> right)`
- `notIn(T... right)`

### DIFF
--- a/querydsl-core/src/main/java/com/querydsl/core/types/dsl/SimpleExpression.java
+++ b/querydsl-core/src/main/java/com/querydsl/core/types/dsl/SimpleExpression.java
@@ -14,6 +14,7 @@
 package com.querydsl.core.types.dsl;
 
 import java.util.Collection;
+import java.util.Collections;
 
 import javax.annotation.Nullable;
 
@@ -189,10 +190,14 @@ public abstract class SimpleExpression<T> extends DslExpression<T> {
      */
     public BooleanExpression in(Collection<? extends T> right) {
         if (right.size() == 1) {
-            return eq(right.iterator().next());
-        } else {
-            return Expressions.booleanOperation(Ops.IN, mixin, ConstantImpl.create(right));
+            T elem = right.iterator().next();
+
+            if (elem != null) {
+                return eq(elem);
+            }
         }
+
+        return Expressions.booleanOperation(Ops.IN, mixin, ConstantImpl.create(right));
     }
 
     /**
@@ -203,7 +208,11 @@ public abstract class SimpleExpression<T> extends DslExpression<T> {
      */
     public BooleanExpression in(T... right) {
         if (right.length == 1) {
-            return eq(right[0]);
+            if (right[0] != null) {
+                return eq(right[0]);
+            } else {
+                return Expressions.booleanOperation(Ops.IN, mixin, ConstantImpl.create(Collections.singleton(null)));
+            }
         } else {
             return Expressions.booleanOperation(Ops.IN, mixin, ConstantImpl.create(ImmutableList.copyOf(right)));
         }
@@ -291,10 +300,14 @@ public abstract class SimpleExpression<T> extends DslExpression<T> {
      */
     public BooleanExpression notIn(Collection<? extends T> right) {
         if (right.size() == 1) {
-            return ne(right.iterator().next());
-        } else {
-            return Expressions.booleanOperation(Ops.NOT_IN, mixin, ConstantImpl.create(right));
+            T elem = right.iterator().next();
+
+            if (elem != null) {
+                return ne(elem);
+            }
         }
+
+        return Expressions.booleanOperation(Ops.NOT_IN, mixin, ConstantImpl.create(right));
     }
 
     /**
@@ -305,7 +318,11 @@ public abstract class SimpleExpression<T> extends DslExpression<T> {
      */
     public BooleanExpression notIn(T... right) {
         if (right.length == 1) {
-            return ne(right[0]);
+            if (right[0] != null) {
+                return ne(right[0]);
+            } else {
+                return Expressions.booleanOperation(Ops.NOT_IN, mixin, ConstantImpl.create(Collections.singleton(null)));
+            }
         } else {
             return Expressions.booleanOperation(Ops.NOT_IN, mixin, ConstantImpl.create(ImmutableList.copyOf(right)));
         }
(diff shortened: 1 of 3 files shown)


jwgmeligmeyling:
Keep in mind that even `IN` comparisons with `NULL` will be unreliable and error-prone, which is why EQ/NEQ doesn't accept it in the first place.

Its really always better to use `isNull()` and `isNotNull()` where-ever possible.

In fact, the better change would probably be to iterate through the collections used for `in`/`notIn` and throw an Exception when any of the values is `null`, equivalently to the handling for `eq`/`neq`.

ajacob:
Thank you @jwgmeligmeyling for your fast reply !

I am aware of the gotchas using `NULL`, but I wonder if it's a good thing to make QueryDSL opinionated and decide whether a query should or should not be sent to the datastore.

Both
```sql
select * from cat where name in (null);
```
and
```sql
select * from cat where name not in (null);
```
return an empty set using Postgres or Mysql. As a QueryDSL end user making applications, I tend to think it's reasonable for the query to be ran and let the datastore decide what to do with it. If an empty result set should be sent or an error be thrown is a matter of the datastore implementation and I find it acceptable.

It's true that following this logic, even `eq(T right)` should not throw an `IllegalArgumentException` when null is passed !

jwgmeligmeyling:
`eq(null)` also suffers from the problem that one can't instantiate a constant expression of null (the reason this logic was implemented initially: https://bugs.launchpad.net/querydsl/+bug/720078 ). That problem doesn't occur here.

I don't really have a use case for this still. It seems rather error prone to me. I'd assume the `null` result would be treated `falsy`, so rows with `name` null would still be eliminated in the result. That may be unexpected. Not everyone expects `null != null`, so people might expect `null in (null)` to be `true` which it isn't.

ajacob:
> eq(null) also suffers from the problem that one can't instantiate a constant expression of null (the reason this logic was implemented initially: https://bugs.launchpad.net/querydsl/+bug/720078 ). That problem doesn't occur here.

I was wondering if it could be done with `Expressions.<T>nullExpression()` :thinking: 

Another option would be to silently remove `null`s from collection / varargs for `in` and `notIn` (instead of adding a behavior to throw an `IllegalArgumentException`)

What do you think about this option ?

I'm also wondering of options regarding `eq(null)`:
- Silently ignore the call (instead of throwing an `IllegalArgumentException`)
- Silently convert to `isNull`

But both options introduce non-negligible behavior change, I'm far more confortable wih a change for `in` and `notIn` !

johnktims:
Hey @ajacob. Our CI was broken when you branched from `master`. Would you mind rebasing on `master`? Sorry for the inconvenience.

jwgmeligmeyling:
@ajacob You should be able to run the test locally using `mvn -Pquickbuild,all clean install`. There are quite a few test failures that need to be resolved.

jwgmeligmeyling:
Hmm I cant explain all test failures. @johnktims  any idea whats going on here?

johnktims:
This is the first error I see: https://github.com/querydsl/querydsl/pull/2577/checks?check_run_id=718172294#step:6:1475

I get the same failure locally.

ajacob:
@jwgmeligmeyling If I run `mvn -Pquickbuild,all clean install` in local I'm getting a `BUILD SUCCESS` :thinking:

johnktims:
`-Pquickbuild` skips the tests. https://github.com/querydsl/querydsl/blob/master/pom.xml#L726

ajacob:
`mvn -Pall clean verify` ends with a `BUILD SUCCESS` also, after running tests (~4 times longer)

johnktims:
Just out of curiosity, what does `mvn --version` show?

johnktims:
This is bizarre. If test your branch on macOS, it fails with the same error in the first Github Actions build.

If I use Docker to run it on linux, it passes.
```
docker run --rm -i maven:3-jdk-8 bash <<-EOT
git clone https://github.com/querydsl/querydsl.git
cd querydsl/
git checkout -b ajacob-SimpleExpression-in-notin-unique-null master
git pull git://github.com/ajacob/querydsl.git SimpleExpression-in-notin-unique-null
mvn clean package
EOT
```

jwgmeligmeyling:
QueryDSL has always been opinionated about using nulls for constants. It's up to the application to handle nulls. These errors prevent bugs because they challenge users to deal with their nulls, instead of relying on comparisons with possibly unexpected results.

We're targeting Java (or JVM) programmers that do not necessarily know better than `null == null`. I find this a somewhat undesirable patch.

I am not necessarily against it. I too suffered the inconvenience from code breaking in production because it didn't handle a null that would have been totally fine had it ended up in the query. But there will always be two flavours here:

1. Do I want to use SQL null? I.e. `expr = null` yields `false`?
2. Do I want to rewrite `expr = x` to `expr is null` if `x` is a Java `null`?

I don't think we can choose for either of these because it will be application-specific. Forcing either one of these upon users is a dangerous assumption to make IMO.

jwgmeligmeyling:
if (age == null) {
    builder.tryAnd(QUser.user.id.isNull());
} else {
    builder.tryAnd(QUser.user.id.eq(age));
}

jiayp:
The age is a param from front, should not add the query condition when it's value is null.
The current implement is 
````
if (!ObjectUtils.isEmpty(age)) {
      builder.and(QUser.user.age.eq(age));
    }
````
But I want simplified into
····
tryAnd(age, QUser.user.id.eq(age));
····
Three line to one line and can chained.

jiayp:
And I write my custom TryBooleanBuilder,but when I use `tryAnd(age, QUser.user.id.eq(age));` and the age is null then exception throwed,because the null check is in `eq` method not in BooleanBuilder.

jiayp:
What I want is do not add the eq(null) predicate to boolenbuilder when a is null.So in my case, It safe.But I can not do it like this because the `eq` will thow exception when I call it.
I will try this way:`builder.tryAnd(QUser.user.age, "eq", paramAge).

PR #2578, opened 2020-05-07
[fixes #2566] fix: fixed hibernate handler iterate & trasnform conditional statement

### LINKED ISSUE #2566: HibernateHandler.iterate does not use scrolling (Hibernate 5.4 JPA)
I've upgraded from Hibernate 5.1.10.Final + Querydsl 4.2.2 to Hibernate 5.4.14.Final + Querydsl 4.3.1. Everything seems to work except results iteration. The implementation in HibernateHandler does not use scrolling and always fetches the whole list.

I found that if i change the condition `query instanceof NativeQuery` to `query instanceof org.hibernate.query.Query` (and the `unwrap` call under it), it stats working properly again.

Unfortunately `HibernateHandler` is package protected and I haven't found a way to register my own `QueryHandler`. So I hacked the temporary solution with Javassist:

```kotlin
        val pool = ClassPool.getDefault()
        val hibernateHandler = pool.get("com.querydsl.jpa.HibernateHandler")

        val iterate = hibernateHandler.getDeclaredMethod("iterate")
        iterate.insertBefore("""
            if ($1 instanceof org.hibernate.query.Query) {
                return shim.FixedHibernateHandler.iterate($$);
            }
        """)

        val transform = hibernateHandler.getDeclaredMethod("transform")
        transform.insertBefore("""
            if ($1 instanceof org.hibernate.query.Query) {
                return shim.FixedHibernateHandler.transform($$);
            } 
        """)

        hibernateHandler.toClass()
```

```java
package shim;

public final class FixedHibernateHandler {

    public static <T> CloseableIterator<T> iterate(Query query, FactoryExpression<?> projection) {
        org.hibernate.query.Query<?> hQuery = query.unwrap(org.hibernate.query.Query.class);
        ScrollableResults results = hQuery.scroll(ScrollMode.FORWARD_ONLY);
        CloseableIterator<T> iterator = new ScrollableResultsIterator<>(results);
        if (projection != null) {
            iterator = new TransformingIterator<>(iterator, projection);
        }
        return iterator;
    }

    @SuppressWarnings("deprecation")
    public static boolean transform(Query query, FactoryExpression<?> projection) {
        query.unwrap(org.hibernate.query.Query.class).setResultTransformer(transformer);
        return true;
    }
```

heesuk-ahn:
## Problem

HibernateHandler.iterate does not use scrolling issue.

Fixes #2566 

## Description

The main subjects that use Hibernate handlers are `JpaSQLQuery` and `JpaQuery`.

Using JpaSQLQuery, when you call `entityManager.createNaitiveQuery`, you get a `NativeQueryImpl`. 

On the other hand, when using JpaQuery, you get `QueryImpl` by calling` entityManager.createQuery`.

iterate func and transform func should be usable in both implementations. (NativeQueryImpl and QueryImpl)

So, I changed the instance check with `org.hibernate.query.Query`, the parent of both implementations.

### DIFF
--- a/querydsl-jpa/src/main/java/com/querydsl/jpa/HibernateHandler.java
+++ b/querydsl-jpa/src/main/java/com/querydsl/jpa/HibernateHandler.java
@@ -52,8 +52,8 @@ class HibernateHandler implements QueryHandler {
     @SuppressWarnings("unchecked")
     @Override
     public <T> CloseableIterator<T> iterate(Query query, FactoryExpression<?> projection) {
-        if (query instanceof NativeQuery) {
-            NativeQuery hQuery = (NativeQuery) query;
+        if (query instanceof org.hibernate.query.Query) {
+            org.hibernate.query.Query hQuery = (org.hibernate.query.Query) query;
             ScrollableResults results = hQuery.scroll(ScrollMode.FORWARD_ONLY);
             CloseableIterator<T> iterator = new ScrollableResultsIterator<T>(results);
             if (projection != null) {
@@ -73,9 +73,9 @@ class HibernateHandler implements QueryHandler {
     @SuppressWarnings("deprecation")
     @Override
     public boolean transform(Query query, FactoryExpression<?> projection) {
-        if (query instanceof NativeQuery) {
+        if (query instanceof org.hibernate.query.Query) {
             ResultTransformer transformer = new FactoryExpressionTransformer(projection);
-            query.unwrap(NativeQuery.class).setResultTransformer(transformer);
+            query.unwrap(org.hibernate.query.Query.class).setResultTransformer(transformer);
             return true;
         } else {
             return false;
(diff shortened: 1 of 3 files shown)


querydsl-jpa/src/main/java/com/querydsl/jpa/HibernateHandler.java
```diff
@@ -52,8 +52,8 @@ public boolean createNativeQueryTyped() {
     @SuppressWarnings("unchecked")
     @Override
     public <T> CloseableIterator<T> iterate(Query query, FactoryExpression<?> projection) {
-        if (query instanceof NativeQuery) {
-            NativeQuery hQuery = (NativeQuery) query;
+        if (query instanceof org.hibernate.query.Query) {
```
jwgmeligmeyling, line 55: Shouldnt we just always attempt to unwrap and not do any instanceof checks at all?

jwgmeligmeyling:
I have provided an alternative fix #2604 that also works if the query/EM is proxied by for example a J2EE application server.

Issue #2504, opened 2019-12-19
when groupBy with two fields,fetchCount throw exception


DannySnow:
here is my code:

```
query.groupBy(_entity.field1,_entity.field2).fetchCount()
```

then the exception:

```
java.lang.IllegalArgumentException: org.hibernate.hql.internal.ast.QuerySyntaxException: expecting CLOSE, found ',' near line 1, column 56 [select count(distinct entity.field1, entity.field2)
from com.demo.core.domain.model.entity entity
where entity.field3 = ?1

	at org.hibernate.internal.ExceptionConverterImpl.convert(ExceptionConverterImpl.java:133)
	at org.hibernate.internal.ExceptionConverterImpl.convert(ExceptionConverterImpl.java:157)
	at org.hibernate.internal.ExceptionConverterImpl.convert(ExceptionConverterImpl.java:164)
	at org.hibernate.internal.AbstractSharedSessionContract.createQuery(AbstractSharedSessionContract.java:670)
	at org.hibernate.internal.AbstractSessionImpl.createQuery(AbstractSessionImpl.java:23)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.springframework.orm.jpa.SharedEntityManagerCreator$SharedEntityManagerInvocationHandler.invoke(SharedEntityManagerCreator.java:305)
	at com.sun.proxy.$Proxy119.createQuery(Unknown Source)
	at com.querydsl.jpa.impl.AbstractJPAQuery.createQuery(AbstractJPAQuery.java:101)
	at com.querydsl.jpa.impl.AbstractJPAQuery.fetchCount(AbstractJPAQuery.java:81)
	at com.demo.core.infrastructure.repository.impl.queryRepositoryImpl.query(queryRepositoryImpl.java:91)
	at com.demo.core.infrastructure.repository.impl.queryRepositoryImplTest.query(queryRepositoryImplTest.java:25)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[...]

Issue #2053, opened 2016-11-18
Fetching `QueryResults` from a query having both `groupBy` and `having` clause results in a `QuerySyntaxException`


uvpoblotzki:
Fetching `QueryResults` from a query having both `groupBy` and `having` clause results in a `QuerySyntaxException`. Here are two test cases, the first one executes without error the second results in the above described error: 

```
    @Test
    public void groupByHavingFetch() {
        QCustomer customer = QCustomer.customer;
        JPAQuery<Customer> customerQuery = new JPAQuery<Customer>(entityManager, HQLTemplates.DEFAULT).from(customer);

        customerQuery.groupBy(customer.lastName).having(customer.id.count().gt(1)).fetch();
    }

    @Test
    public void groupByHavingFetchResults() {
        QCustomer customer = QCustomer.customer;
        JPAQuery<Customer> customerQuery = new JPAQuery<Customer>(entityManager, HQLTemplates.DEFAULT).from(customer);

        customerQuery.groupBy(customer.lastName).having(customer.id.count().gt(1)).fetchResults();
    }
```
The test in error generates following error:

```
java.lang.IllegalArgumentException: org.hibernate.hql.internal.ast.QuerySyntaxException: unexpected token: having near line 3, column 1 [select count(distinct customer.lastName)
from Customer customer
having count(customer.id) > ?1]

	at org.hibernate.internal.ExceptionConverterImpl.convert(ExceptionConverterImpl.java:131)
	at org.hibernate.internal.ExceptionConverterImpl.convert(ExceptionConverterImpl.java:155)
	at org.hibernate.internal.ExceptionConverterImpl.convert(ExceptionConverterImpl.java:162)
	at org.hibernate.internal.AbstractSharedSessionContract.createQuery(AbstractSharedSessionContract.java:662)
	at org.hibernate.internal.AbstractSessionImpl.createQuery(AbstractSessionImpl.java:23)
	at com.querydsl.jpa.impl.AbstractJPAQuery.createQuery(AbstractJPAQuery.java:101)
	at com.querydsl.jpa.impl.AbstractJPAQuery.fetchResults(AbstractJPAQuery.java:211)
	at GroupByHavingTest.groupByHavingFetchResults(GroupByHavingTest.java:49)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[...]

PR #2580, opened 2020-05-07
[fixes #2579] fix: added `forCount` conditional statement for skip transform logic when countQuery

### LINKED ISSUE #2579: [BUG & Solve] JpaSQLQuery `fetchResults` illegal statement Exception
## Bug Solve Demo link
https://github.com/heesuk-ahn/querydsl-4.3.1-fetchresult-issue-debug

## Problem

If `fetchResults` is called using` JpaSQLQuery`, illegalArgumentException occurs.

## Description

When calling `fetchResults`, the` count` query is called beforehand.

If `ResultTransform` is set to receive the result value of the query as a desired object, illegalException occurs.

Why is this error occurring?

The reason is that you want to convert the result value of the count query that is executed in advance to `ResultTransform`.

To reproduce this error condition, I created the `projections.bean` test for FactoryExpression as below.

```
public QueryResults<BookName> getBooksWithProjectionFromJPASQL(int offset, int limit) {
        SQLTemplates oracleTemplate = OracleTemplates.builder().printSchema().build();
        JPAHelper jpaHelper = new JPAHelper(emf.createEntityManager(), oracleTemplate);

        Map<String, Expression<?>> bindings = new HashMap<>();
        bindings.put("name", book.name);

        return jpaHelper.query()
                        .select(Projections.bean(BookName.class, bindings))
                        .from(book)
                        .offset(offset)
                        .limit(limit)
                        .fetchResults();
    }
```

When fetchResults is called, the following code is executed.

```
 private Query createQuery(boolean forCount) {
         NativeSQLSerializer serializer = (NativeSQLSerializer) serialize(forCount);
         String queryString = serializer.toString();
         logQuery(queryString, serializer.getConstantToAllLabels());
         Expression<?> projection = queryMixin.getMetadata().getProjection();
         Query query;

         .
         .
         .
         . [!] code line 161 - 165       
         if (projection instanceof FactoryExpression) {
             if (!queryHandler.transform(query, (FactoryExpression<?>) projection)) {
                 this.projection = (FactoryExpression<?>) projection;
             }
         }
 
         return query;
     }
```

Let's look at the logic of code line 161 of the createQuery function.

What code lines 161-165 mean is applying a `ResultTransform` for projection

__But the problem here is that the forCount query must `skip` this logic.__

If skip is not performed, ResultTransform is applied to forCount query, and `illegalArgumentException` occurs.

## My Suggest

To solve this problem, I simply added `!ForCount` to conditional statement.

## Integration Test

After changing the code as below, I created a querydsl-jpa snapshot jar and added it to the external lib.

```
[Fixed]

.
.
. code line 161
 if (!forCount && projection instanceof FactoryExpression) { <---- added !forCount
    if (!queryHandler.transform(query, (FactoryExpression<?>) projection)) {
        this.projection = (FactoryExpression<?>) projection;
    }
 }
```

And I could confirm that the below test works normally.

```
    @Test
    public void should_Get_Books_With_Projection_From_JPASQL() {
        final int offset = 0;
        final int limit = 10;

        executeQueryForInsert(emf.createEntityManager(), firstBook);

        QueryResults<BookName> bookNames = bookRepository.getBooksWithProjectionFromJPASQL(offset, limit);

        assertThat(bookNames.getResults().get(0).getName()).isEqualTo(bookName);
    }
```

heesuk-ahn:
## solve demo link

* https://github.com/heesuk-ahn/querydsl-4.3.1-fetchresult-issue-debug

## Check Issue

* #2579

### DIFF
--- a/querydsl-jpa/src/test/java/com/querydsl/jpa/JPASQLBase.java
+++ b/querydsl-jpa/src/test/java/com/querydsl/jpa/JPASQLBase.java
@@ -13,7 +13,9 @@
  */
 package com.querydsl.jpa;
 
-import static org.junit.Assert.assertEquals;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
 
 import javax.persistence.EntityManager;
 import javax.persistence.Query;
@@ -27,6 +29,8 @@ import org.junit.runner.RunWith;
 
 import com.querydsl.core.Target;
 import com.querydsl.core.testutil.ExcludeIn;
+import com.querydsl.core.types.Expression;
+import com.querydsl.core.types.Projections;
 import com.querydsl.jpa.domain.Cat;
 import com.querydsl.jpa.domain.Color;
 import com.querydsl.jpa.domain.QCat;
@@ -34,6 +38,8 @@ import com.querydsl.jpa.domain.sql.SAnimal;
 import com.querydsl.jpa.sql.JPASQLQuery;
 import com.querydsl.jpa.testutil.JPATestRunner;
 import com.querydsl.sql.SQLTemplates;
+
+import static org.junit.Assert.assertEquals;
 
 @RunWith(JPATestRunner.class)
 public class JPASQLBase extends AbstractSQLTest implements JPATest {
@@ -49,6 +55,9 @@ public class JPASQLBase extends AbstractSQLTest implements JPATest {
     private final SQLTemplates templates = Mode.getSQLTemplates();
 
     private EntityManager entityManager;
+
+    private final SAnimal cat = new SAnimal("cat");
+    private final QCat catEntity = QCat.cat;
 
     @Override
     protected JPASQLQuery<?> query() {
@@ -73,11 +82,22 @@ public class JPASQLBase extends AbstractSQLTest implements JPATest {
         }
     }
 
+    private <T> void insertEntitiesForTest(final List<T> entities) {
+        for (T entity: entities) {
+            entityManager.persist(entity);
+        }
+        entityManager.flush();
+    }
+
+    private <T> void removeEntitiesForTest(final List<T> entities) {
+        for (T entity: entities) {
+            entityManager.remove(entity);
+        }
+        entityManager.flush();
+    }
+
     @Test
     public void entityQueries_createQuery() {
-        SAnimal cat = new SAnimal("cat");
-        QCat catEntity = QCat.cat;
-
         Query query = query().from(cat).select(catEntity).createQuery();
         assertEquals(6, query.getResultList().size());
     }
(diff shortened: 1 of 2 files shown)


johnktims:
Would you mind adding the tests you wrote in the issue? Also, It looks like you branched from `master` when the CI was broken so you'll need to rebase. Sorry for the hassle.

heesuk-ahn:
@johnktims You're welcome :)

I will do it as soon as possible.

TODO)

* should add test code for this issue
* rebase code

If you have any other thoughts, please comment. :)

jwgmeligmeyling:
While adding a test, please also have a look how it handles queries with multiple group by clauses or a having clause. For JPA (non-SQL) queries, this causes various issues (#2504, #2053).

heesuk-ahn:
@jwgmeligmeyling Thanks for the good comments.

I also added a test for group and having. :)

heesuk-ahn:
humm..

It says that `start test database` is broken. Do you know anything about this issue?

(https://github.com/querydsl/querydsl/pull/2580/checks?check_run_id=809917861)

@johnktims @jwgmeligmeyling

johnktims:
@heesuk-ahn It occasionally times out. I restarted the build and the test databases started correctly this time.

querydsl-jpa/src/test/java/com/querydsl/jpa/JPASQLBase.java
```diff
@@ -86,10 +90,49 @@ public void entityQueries_createQuery() {
...
+    }
+
+    @Test
+    public void should_get_black_cat_count_by_using_group_by_and_having() {
+        final long expectedBlackCatCount = 6L;
+
+        final long actualBlackCatCount = query().from(cat)
+                                                .select(catEntity.name.count())
+                                                .groupBy(catEntity.color)
+                                                .having(catEntity.name.count().goe(1))
```
jwgmeligmeyling, line 132: This having clause is not entropy reducing, so even if its handled correcly, the query gives the proper result.

heesuk-ahn: How should I test it? Can you give me a comment~? 🤔

jwgmeligmeyling: `catEntity.name.count()` can never be 0. It is minimally 1. But it can be higher than 1. Not sure if that occurs in your test data. The point is, `catEntity.name.count() >= 1` will never filter any rows. Therefore you can't know whether the having clause was properly handled. It would be interesting to see what happens for `catEntity.name.count() >= 2`. It should return a non empty result (the ones   with `count() >= 2`), and eliminate the results with `count() = 1`). The latter for example doesn't happen in JPA currently. It doesn't handle the having clause at all. Therefore count queries with a HAVING clause are fundamentally broken in JPQL (not JPA SQL), and it would be interesting to see how this behaves in JPA SQL.

heesuk-ahn: Thanks for sharing your thoughts. :)

I have created two additional test data to check the having clause more accurately. (tabbyColorCatFoo, tabbyColorCatBar)

The total test data were 6 black cats and 2 Tabby cats.

And I wrote the query as below.

```
final long actualTabbyCatCount = query().from(cat)
                                                                      .select(catEntity.name.count())
                                                                      .groupBy(catEntity.color)
                                                                      .having(catEntity.name.count().eq(2L))
                                                                      .fetchOne();

assertEquals(expectedTabbyCatCount, actualTabbyCatCount); //true
```

jwgmeligmeyling:
Thanks for your changes @heesuk-ahn , I'll review it tomorrow!

heesuk-ahn:
if you'll execuse me, any update :) ~?

Issue #2582, opened 2020-05-11
[querydsl-sql-codegen] Serialize keys in ABC sorted order so the same file will be generated always


balazs-zsoldos:
Current situation:

If we re-generate metadata classes, primary keys, foreign keys and inverse foreign keys will be generated in "random" order in the class as they are stored in HashMaps.

In a project where the generated files are part of the committed source and the files are re-generated from time-to-time, such metadata classes will change where there is no functional change just because the order of the keys are different in the file. This causes unnecessary changes in the git commits.

Suggestion: Generate metadata classes in the way that keys are ABC sorted, so the generated files will be always the same if the database is not changed.

PR #2583, opened 2020-05-11
#2582 keys should be sorted in metadata classes

balazs-zsoldos:
Please review this PR. I modified the tests to check if foreign keys and inverse foreign keys are sorted in ABC order. However, I am not sure that all databases would support multiple primary keys on one table, so it is not tested.


johnktims:
This looks good to me. I'll let the others take a look though. Thanks for the contribution.

balazs-zsoldos:
Any chance that this could be pulled in before the next release?

johnktims:
Thanks for the contribution, @balazs-zsoldos !

PR #2298, opened 2018-04-21
Don't add @NotNull on columns with default value. #2294

jrodalo:
If a column on the DB has a default value it shouldn't add a @NotNull annotation on the exported Bean.

Issue #2294, opened 2018-04-19
[QueryDSL SQL] Don't generate javax.validation.NotNull for primary keys


jrodalo:
Hi, 

I couldn't find any issues about this problem. I can't use "`validationAnnotations`" if I want to use `@Valid` to validate a new instance (with id=null).

I'll try to explain the problem ;)

### Expected behaviour

If the column is a Primary Key or has a `default` value it shouldn't add a `@NotNull` annotation on the field.

### Actual behaviour

QueryDSL's MetaDataExporter adds the `@NotNull` annotation on all columns that are marked as not nullable on the table.

```java
if (validationAnnotations) {
    if (nullable == DatabaseMetaData.columnNoNulls) {
        property.addAnnotation(new NotNullImpl());
    }
    ...
}
```

https://github.com/querydsl/querydsl/blob/master/querydsl-sql-codegen/src/main/java/com/querydsl/sql/codegen/MetaDataExporter.java#L314

### Steps to reproduce

1. Create a table with a primary key

```sql
CREATE TABLE "users" (
  id INTEGER PRIMARY KEY DEFAULT nextval('users_id_seq') NOT NULL,
  name VARCHAR(255)
);
```

2. Configure querydsl-maven-plugin to generate validation annotations

```xml
<plugin>
    <groupId>com.querydsl</groupId>
    <artifactId>querydsl-maven-plugin</artifactId>
    <version>4.2.1</version>
    ...
    <configuration>
        <exportBeans>true</exportBeans>
        <validationAnnotations>true</validationAnnotations>
    </configuration>
    ...
</plugin>
```

3. Generate sources and check generated code (ID field has `@NotNull`)

```java
@Generated("com.querydsl.codegen.BeanSerializer")
public class Users {

    @NotNull
    private Integer id;
    
    @Size(max=255)
    private String name;
    
    ...
}
```

4. Try to use `@Valid` in a Controller to insert a new user (I'm using Spring here)

```java
@Controller
public class TestController {

    @PostMapping("/test")
    public String insert(@ModelAttribute @Valid User user, BindingResult result) {
    
        result.hasErrors() // == true (Field error in object 'user' on field 'id': rejected value [null]
        
        return "view";
    }
}
```

### Environment

- QueryDSL SQL 4.2.1
- Spring Boot 2.0.1
- PostgreSQL 9.3
- Maven
- Java 8

jrodalo:
[...]

PR #2602, opened 2020-05-29
Don't add @NotNull on columns with default value. #2294

johnktims:
https://github.com/querydsl/querydsl/pull/2298 + tests


kartoffelsup:
But if the value can't be null, why is NotNull wrong then?

PR #2604, opened 2020-05-31
Hibernate 5 Support

### LINKED ISSUE #2328: Left Join with condition on right table not working
I have two tables - `user` and `booking`. Each user may have many bookings (one-to-many relationship). 

    user:                          booking:
    
        id | name    |               id | country    | user_id | price   |
        -------------|               ------------------------------------|
         1 | Alice   |                1 | Italy      | 1       | 2000    |
         2 | Bob     |                2 | France     | 1       | 2500    |
                                      3 | Spain      | 1       | 3000    |

I want to select all users and all bookings where booking's price is greater than 2000 using `Query DSL`. If a user doesn't have any bookings or bookings don't match the condition I still want to select this user. 

First, let's have a look at how it would look like using a simple SQL left join query:

    SELECT u.*, b.* FROM user u LEFT JOIN booking b ON u.id = b.user_id AND b.price > 2000

The above query should provide the following result:

        id | name    | id    | country    | user_id | price   |
        -------------|----------------------------------------|
         1 | Alice   | 2     | France     | 1       | 2500    |
         1 | Alice   | 3     | Spain      | 1       | 3000    |
         2 | Bob     | null  | null       | null    | null    |
      

Now I want to do it using `JPA` with `Query DSL`

**JPA-related stuff:**

    @Entity
    public class User {
    
        @Id
        private Long id;
        private String name;
    
        @OneToMany(cascade = ALL, fetch = EAGER, orphanRemoval = true, mappedBy = "user")
        private List<Booking> bookings;
    
        // getters and setters
    }

    @Entity
    public class Booking {
    
        @Id
        private Long id;
        private String name;
        private Integer price;
    
        @ManyToOne(fetch = LAZY)
        @JoinColumn(name = "user_id")
        private User user;
    
        // getters and setters
    }

**Query DSL:**

    public List<User> getUsersAndBookings() {
        QUser user = QUser.user;
        QBooking booking = QBooking.booking;
    
        JPAQuery<User> jpaQuery = new JPAQuery(entityManager);

        List<User> result = jpaQuery.from(user).leftJoin(user.bookings, booking).on(booking.price.gt(2000)).fetchJoin().fetch();
    
        return result;
    }

In fact, this code is not working and I get the following exception:

    org.hibernate.hql.internal.ast.QuerySyntaxException: with-clause not allowed on fetched associations; use filters [select user from com.example.demo.entity.User user left join fetch user.bookings as booking with booking.price > ?1]

The problem is that the condition clause is specified in `on` method - `on(booking.price.gt(2000))`. 

After some research I found that this condition should be specified in `where` method and should look like this:

    List<User> result = jpaQuery.from(user).leftJoin(user.bookings, booking).where(booking.price.gt(2000)).fetchJoin().fetch();

This works, but not how I would expect it to work, since it doesn't return ALL users, it returns only one user (Alice), which has some bookings, matching the condition clause. Basically, it just filters the merged table (result table after left join operation) and that's not what I'm looking for.

I want to retrieve all users, and if there are no any bookings for a specific user, then just have `null` instead of booking list for this user.
 

Please help, been struggling for hours without any success. 

<hr/>
Versions used:

- Spring Boot 2.0.2   
- Spring Data JPA 2.0.7  
- Hibernate 5.2.16.Final   
- QueryDSL 4.1.4

### LINKED ISSUE #2566: HibernateHandler.iterate does not use scrolling (Hibernate 5.4 JPA)

jwgmeligmeyling:
The last portion of Hibernate 5 support.

Supersedes #2578
Fixes #2566, the remaining issue for Hibernate 5.x support after #2264 (in #2510)
Also kind of fixes #2328, as the issue there was that the legacy `WITH` join type was still used for Hibernate, while this currently should use `ON`.

### DIFF
--- a/querydsl-jpa/src/main/java/com/querydsl/jpa/HQLTemplates.java
+++ b/querydsl-jpa/src/main/java/com/querydsl/jpa/HQLTemplates.java
@@ -28,8 +28,9 @@ import com.querydsl.core.types.Ops;
  * HQLTemplates extends {@link JPQLTemplates} with Hibernate specific extensions
  *
  * @author tiwe
- *
+ * @deprecated Most likely you want to use {@link Hibernate5Templates} instead
  */
+@Deprecated
 public class HQLTemplates extends JPQLTemplates {
 
     private static final QueryHandler QUERY_HANDLER;
@@ -86,16 +87,22 @@ public class HQLTemplates extends JPQLTemplates {
 
     @Override
     public boolean wrapElements(Operator operator) {
+        // For example: JPaIntegration.docoExamples98_12
         return wrapElements.contains(operator);
     }
 
     @Override
     public String getTypeForCast(Class<?> cl) {
-        return typeNames.get(cl);
+        String typeName = typeNames.get(cl);
+        if (typeName == null) {
+            return super.getTypeForCast(cl);
+        }
+        return typeName;
     }
 
     @Override
     public String getExistsProjection() {
+        // TODO Required / supported just for Hibernate?
         return "1";
     }
 
--- a/querydsl-jpa/src/main/java/com/querydsl/jpa/Hibernate5Templates.java
+++ b/querydsl-jpa/src/main/java/com/querydsl/jpa/Hibernate5Templates.java
@@ -1,0 +1,32 @@
+package com.querydsl.jpa;
+
+/**
+ * Hibernate5Templates extends {@link JPQLTemplates} with Hibernate specific extensions
+ *
+ * @author Jan-Willem Gmelig Meyling
+ *
+ */
+public class Hibernate5Templates extends HQLTemplates {
+
+    public static final Hibernate5Templates DEFAULT = new Hibernate5Templates();
+
+    public Hibernate5Templates() {
+    }
+
+    public Hibernate5Templates(char escape) {
+        super(escape);
+    }
+
+    @Override
+    public boolean wrapConstant(Object constant) {
+        // HHH-6913 is fixed in 5.0, default to JPA behaviour
+        return false;
+    }
+
+    @Override
+    public boolean isWithForOn() {
+        // Hibernate supports the on-clause since 5.0, and the ON clause is actually mandatory for entity joins
+        return true;
+    }
+
+}
--- a/querydsl-jpa/src/main/java/com/querydsl/jpa/HibernateHandler.java
+++ b/querydsl-jpa/src/main/java/com/querydsl/jpa/HibernateHandler.java
@@ -15,6 +15,7 @@ package com.querydsl.jpa;
 
 import java.util.Iterator;
 
+import javax.persistence.PersistenceException;
 import javax.persistence.Query;
 
 import org.hibernate.ScrollMode;
@@ -32,7 +33,7 @@ import com.querydsl.core.types.FactoryExpression;
  * @author tiwe
  *
  */
-class HibernateHandler implements QueryHandler {
+public class HibernateHandler implements QueryHandler {
 
     @Override
     public void addEntity(Query query, String alias, Class<?> type) {
@@ -52,15 +53,15 @@ class HibernateHandler implements QueryHandler {
     @SuppressWarnings("unchecked")
     @Override
     public <T> CloseableIterator<T> iterate(Query query, FactoryExpression<?> projection) {
-        if (query instanceof NativeQuery) {
-            NativeQuery hQuery = (NativeQuery) query;
-            ScrollableResults results = hQuery.scroll(ScrollMode.FORWARD_ONLY);
+        try {
+            org.hibernate.query.Query unwrappedQuery = query.unwrap(org.hibernate.query.Query.class);
+            ScrollableResults results = unwrappedQuery.scroll(ScrollMode.FORWARD_ONLY);
             CloseableIterator<T> iterator = new ScrollableResultsIterator<T>(results);
             if (projection != null) {
                 iterator = new TransformingIterator<T>(iterator, projection);
             }
             return iterator;
-        } else {
+        } catch (PersistenceException e) {
             Iterator<T> iterator = query.getResultList().iterator();
             if (projection != null) {
                 return new TransformingIterator<T>(iterator, projection);
@@ -73,11 +74,11 @@ class HibernateHandler implements QueryHandler {
     @SuppressWarnings("deprecation")
     @Override
     public boolean transform(Query query, FactoryExpression<?> projection) {
-        if (query instanceof NativeQuery) {
+        try {
             ResultTransformer transformer = new FactoryExpressionTransformer(projection);
-            query.unwrap(NativeQuery.class).setResultTransformer(transformer);
+            query.unwrap(org.hibernate.query.Query.class).setResultTransformer(transformer);
             return true;
-        } else {
+        } catch (PersistenceException e) {
             return false;
         }
     }
--- a/querydsl-jpa/src/main/java/com/querydsl/jpa/impl/JPAProvider.java
+++ b/querydsl-jpa/src/main/java/com/querydsl/jpa/impl/JPAProvider.java
@@ -18,7 +18,13 @@ import java.util.Map;
 import javax.persistence.EntityManager;
 
 import com.google.common.collect.Maps;
-import com.querydsl.jpa.*;
+import com.querydsl.jpa.BatooTemplates;
+import com.querydsl.jpa.DataNucleusTemplates;
+import com.querydsl.jpa.EclipseLinkTemplates;
+import com.querydsl.jpa.HQLTemplates;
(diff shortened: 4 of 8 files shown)


querydsl-jpa/src/main/java/com/querydsl/jpa/Hibernate5Templates.java
```diff
@@ -0,0 +1,30 @@
...
+    public Hibernate5Templates() {
+    }
+
+    public Hibernate5Templates(char escape) {
+        super(escape);
+    }
+
+    @Override
+    public boolean isWithForOn() {
+        return false;
```
jwgmeligmeyling, line 22: Hibernates WITH clause was HQL's precesessor of JPA's ON clause. Hibernate 5.x implements JPA 2.1 which introcued the ON clause for JPA. Nowadays, the WITH clause has limited functionality over the ON clause. For example entity joins require the ON clause instead. So currently, Hibernate users would actually be better of using the standard `JPATemplates` than the `JPQLTemplates`. 

This aims to set that straight.

querydsl-jpa/src/main/java/com/querydsl/jpa/HibernateHandler.java
```diff
@@ -52,15 +53,15 @@ public boolean createNativeQueryTyped() {
     @SuppressWarnings("unchecked")
     @Override
     public <T> CloseableIterator<T> iterate(Query query, FactoryExpression<?> projection) {
-        if (query instanceof NativeQuery) {
-            NativeQuery hQuery = (NativeQuery) query;
-            ScrollableResults results = hQuery.scroll(ScrollMode.FORWARD_ONLY);
+        try {
+            org.hibernate.query.Query unwrappedQuery = query.unwrap(org.hibernate.query.Query.class);
```
jwgmeligmeyling, line 57: Remove instance of checks,  just attempt to unwrap. This is because the query may be wrapped in various layers of proxies.

Also, we should unwrap a `Query` here and not a `NativeQuery` because this code is also used for regular queries.

querydsl-jpa/src/main/java/com/querydsl/jpa/HibernateHandler.java
```diff
@@ -73,11 +74,11 @@ public boolean createNativeQueryTyped() {
     @SuppressWarnings("deprecation")
     @Override
     public boolean transform(Query query, FactoryExpression<?> projection) {
-        if (query instanceof NativeQuery) {
+        try {
             ResultTransformer transformer = new FactoryExpressionTransformer(projection);
-            query.unwrap(NativeQuery.class).setResultTransformer(transformer);
+            query.unwrap(org.hibernate.query.Query.class).setResultTransformer(transformer);
```
jwgmeligmeyling, line 79: Remove instance of checks,  just attempt to unwrap. This is because the query may be wrapped in various layers of proxies.

Also, we should unwrap a `Query` here and not a `NativeQuery` because this code is also used for regular queries.

querydsl-jpa/src/main/java/com/querydsl/jpa/impl/JPAProvider.java
```diff
@@ -39,9 +46,23 @@ private static void addMapping(String className, JPQLTemplates templates) {
     }
 
     static {
+        boolean hibernate5;
+
+        try {
+            String version = Class.forName("org.hibernate.Session").getPackage().getImplementationVersion();
+            String[] versionParts = version.split("\\.");
+            int major = Integer.parseInt(versionParts[0]);
+//            hibernate5 = major >= 5;
```
jwgmeligmeyling, line 55: Should be fixed, but I want to see how this affects the test suite. So I will enable this in a separate commit.

querydsl-jpa/src/main/java/com/querydsl/jpa/impl/JPAProvider.java
```diff
@@ -50,16 +71,23 @@ private static void addMapping(String className, JPQLTemplates templates) {
...
         templatesByName.put("datanucleus", DataNucleusTemplates.DEFAULT);
     }
 
     public static JPQLTemplates getTemplates(EntityManager em) {
         // detect by delegate
         for (Map.Entry<Class<?>, JPQLTemplates> entry : mappings.entrySet()) {
-            if (entry.getKey().isAssignableFrom(em.getDelegate().getClass())) {
-                return entry.getValue();
+            try {
+                if (em.unwrap(entry.getKey()) != null) {
```
jwgmeligmeyling, line 84: `unwrap` is preferred over `getDelegate` (even mentioned in the Javadoc of `getDelegate` ). I am keeping the old method as a fallback in the catch clause.

querydsl-jpa/src/main/java/com/querydsl/jpa/HibernateHandler.java
```diff
@@ -32,7 +33,7 @@
  * @author tiwe
  *
  */
-class HibernateHandler implements QueryHandler {
+public class HibernateHandler implements QueryHandler {
```
idosal, line 36: I must be missing something - why is this public now?

querydsl-jpa/src/main/java/com/querydsl/jpa/HQLTemplates.java
```diff
@@ -28,8 +28,9 @@
  * HQLTemplates extends {@link JPQLTemplates} with Hibernate specific extensions
  *
  * @author tiwe
- *
+ * @deprecated Most likely you want to use {@link Hibernate5Templates} instead
```
idosal, line 31: Do we even still support Hibernate 4? I recall an earlier PR already removed compatibility. If so, we might as well replace this implementation.

querydsl-jpa/src/main/java/com/querydsl/jpa/HibernateHandler.java
jwgmeligmeyling: Its not necessairy. The issue reporter mentioned not being able to work around the issueX, because HibernateHandler couldnt be extended or instantiated.  opening it up for extension allows users to use the Hibernate handler with custom Templates as well (for example Templates with custom functions). For that use case the class can easily be copied as well.

The change is unrelated; so if you dont feel like exposing this class I happily revert the change :)

querydsl-jpa/src/main/java/com/querydsl/jpa/HQLTemplates.java
jwgmeligmeyling: I wasnt sure if we just target 5 or dropped support for 4 alltogether. When replacing the template we will really break hibernate 4 compatibility.

Notice that on Hibernate 5 the right templates for Hibernate 5 are already selected in JpaProvider. So I think it wont harm to keep the old Hibernate 4 templates in there.

querydsl-jpa/src/main/java/com/querydsl/jpa/HibernateHandler.java
idosal: Sounds good to me.

PR #2610, opened 2020-06-08
[#2468] R2DBC support

### LINKED ISSUE #2468: R2DBC support
R2DBC (Reactive Relational Database Connectivity) is for reactive programming and containing concise SPI. Looks like it will be very well integrated with the DSL-style frameworks.

See:

- R2DBC SPI: https://github.com/r2dbc/r2dbc-spi
- JavaDoc: https://r2dbc.io/spec/0.8.0.M8/api

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/78112375-r2dbc-support?utm_campaign=plugin&utm_content=tracker%2F303568&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F303568&utm_medium=issues&utm_source=github).
</bountysource-plugin>

riba2101:
This addes R2DBC support (pre-alpha) to QueryDsl.

Fixes #2468


jwgmeligmeyling:
Thanks @riba2101 , I'll have a look tomorrow into your compilation issues

riba2101:
There are no compilation issues, r2dbc tests are failing, will take a look tomorrow

jwgmeligmeyling:
Ah sorry, I misread your commit message. I'll have look regardless 😄

riba2101:
so...fixed locally everything -> could you pls provide a detailed report + tell me what versions you run so that i can do a similar setup locally

querydsl-core/src/main/java/com/querydsl/core/ReactiveFetchable.java
```diff
@@ -0,0 +1,46 @@
...
+     * Consuming the result produces {@link NullPointerException} if the value is null.
+     *
+     * @return first result
+     */
+    Mono<T> fetchFirst();
+
+    /**
+     * Get the projection as a unique result.
+     * Consuming the result produces {@link NullPointerException} if the value is null.
+     * Consuming the result produces {@link IndexOutOfBoundsException} if there is more than one matching result
```
jwgmeligmeyling, line 33: For consistency, this case should really throw `NonUniqueResultException`

querydsl-core/src/main/java/com/querydsl/core/ReactiveFetchable.java
```diff
@@ -0,0 +1,46 @@
...
+     * Get the first result of the projection.
+     * Consuming the result produces {@link NullPointerException} if the value is null.
+     *
+     * @return first result
+     */
+    Mono<T> fetchFirst();
+
+    /**
+     * Get the projection as a unique result.
+     * Consuming the result produces {@link NullPointerException} if the value is null.
```
jwgmeligmeyling, line 32: Instead of throwing a `NullPointerException`, you should return `Mono.empty()`

querydsl-core/src/main/java/com/querydsl/core/ReactiveFetchable.java
```diff
@@ -0,0 +1,46 @@
...
+     * Get the projection as a typed Flux.
+     * Consuming the result produces {@link NullPointerException} if a value is null.
+     *
+     * @return result
+     */
+    Flux<T> fetch();
+
+    /**
+     * Get the first result of the projection.
+     * Consuming the result produces {@link NullPointerException} if the value is null.
```
jwgmeligmeyling, line 24: Instead of throwing a `NullPointerException`, you should return `Mono.empty()`

querydsl-core/src/main/java/com/querydsl/core/ReactiveFetchable.java
```diff
@@ -0,0 +1,46 @@
...
+ * {@code ReactiveFetchable} defines default projection methods for {@link Query} implementations.
+ * All Querydsl query implementations should implement this interface.
+ *
+ * @param <T> result type
+ */
+public interface ReactiveFetchable<T> {
+
+    /**
+     * Get the projection as a typed Flux.
+     * Consuming the result produces {@link NullPointerException} if a value is null.
```
jwgmeligmeyling, line 16: Instead of throwing a `NullPointerException`, you should return `Flux.empty()`

querydsl-core/src/main/java/com/querydsl/core/ReactiveFetchableQuery.java
```diff
@@ -0,0 +1,41 @@
...
+public interface ReactiveFetchableQuery<T, Q extends ReactiveFetchableQuery<T, Q>> extends SimpleQuery<Q>, ReactiveFetchable<T> {
+
+    /**
+     * Change the projection of this query
+     *
+     * @param <U>
+     * @param expr new projection
+     * @return the current object
+     */
+    <U> ReactiveFetchableQuery select(Expression<U> expr);
```
jwgmeligmeyling, line 21:
```suggestion
    <U> ReactiveFetchableQuery<U, ?> select(Expression<U> expr);
```

querydsl-core/src/main/java/com/querydsl/core/ReactiveFetchableQuery.java
```diff
@@ -0,0 +1,41 @@
...
+
+    /**
+     * Change the projection of this query
+     *
+     * @param exprs new projection
+     * @return the current object
+     */
+    ReactiveFetchableQuery select(Expression<?>... exprs);
```
jwgmeligmeyling, line 29:
```suggestion
    ReactiveFetchableQuery<Tuple, ?> select(Expression<?>... exprs);
```

querydsl-core/src/main/java/com/querydsl/core/support/ReactiveFetchableQueryBase.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    @Override
+    public final Mono<T> fetchFirst() {
+        return limit(1).fetchOne();
+    }
+
+    @Override
+    public Mono<T> fetchOne() {
+        return fetch().singleOrEmpty();
```
jwgmeligmeyling, line 31: This actually implements the correct behaviour, make sure to update the Javadoc in the interface.

querydsl-r2dbc/src/main/java/com/querydsl/r2dbc/H2Templates.java
```diff
@@ -0,0 +1,85 @@
...
+import com.querydsl.core.types.Ops;
+
+import java.sql.Types;
+
+/**
+ * {@code H2Templates} is an SQL dialect for H2
+ *
+ * @author tiwe
+ */
+public class H2Templates extends SQLTemplates {
```
jwgmeligmeyling, line 25: H2 templates is on the classpath (you're depending on `querydsl-sql` ). Why is there a need to also ship it here?

I would maybe like to see the `H2RdbcTemplates` extend the actual `H2Templates`. At least then we do not have to worry about the two template variants diverging too much leading to bugs.

But is it really necessary? I am fine with extending the `SQLTemplates`  API in `querydsl-sql`  with a couple of methods that you might need for proper R2dbc support.

querydsl-r2dbc/src/main/java/com/querydsl/r2dbc/JDBCTypeMapping.java
```diff
@@ -0,0 +1,174 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * {@code JDBCTypeMapping} defines a mapping from JDBC types to Java classes.
+ *
+ * @author tiwe
+ */
+final class JDBCTypeMapping {
```
jwgmeligmeyling, line 36: Just use the one from `querydsl-dsl` . Make it `public` if you have to.

querydsl-r2dbc/src/main/java/com/querydsl/r2dbc/JavaTypeMapping.java
```diff
@@ -0,0 +1,162 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * {@code JavaTypeMapping} provides a mapping from Class to Type instances
+ *
+ * @author tiwe
+ */
+class JavaTypeMapping {
```
jwgmeligmeyling, line 30: Just use the one from `querydsl-dsl` . Make it `public` if you have to.

querydsl-r2dbc/src/main/java/com/querydsl/r2dbc/H2Templates.java
riba2101: the reason behind this is that the underlying super `SQLTemplates` differs because of a single line `com.querydsl.r2dbc.SQLTemplates#serializeInsert:919` (`R2DBCInsertBatch` vs `SQLInsertBatch`), yet again didn't know how invasive I'm allowed to be

riba2101:
Yeah, there are def Javadoc inconsistencies - sry for that, was kinda speed running through the code. 

I totally agree with you about the duplicate code problems(101%). Hate them, and it is a maintainability nightmare. I would love to see improvements in extensibility for such a case where the signature of a method differs from the base in the version 5. 

The problem was that I didn't know how invasive I can be :) 

This being said, there are only a couple of places where I could introduce other changes(will do them) without introducing massive changes to the querydsl-sql Or would need to refactor half of SQL to make it extensible. What's your take on this?

jwgmeligmeyling:
> . I would love to see improvements in extensibility for such a case where the signature of a method differs from the base in the version 5.

@johnktims are you fine with making some of the classes duplicated here `public`  instead?

johnktims:
Oh yeah. I’d much rather make the classes `public` than copy them.

querydsl-r2dbc/src/main/java/com/querydsl/r2dbc/JavaTypeMapping.java
riba2101: checked this -> so there are types for r2dbc that changed and this registry is for those types (notice the import `import com.querydsl.r2dbc.types.*;`

querydsl-r2dbc/src/main/java/com/querydsl/r2dbc/JDBCTypeMapping.java
riba2101: see next, here we found that Blob/Clob has a special type for r2dbc - im 99% sure this will change when we change/map to correct r2dbc types (for example r2dbc heavily depends on java.time api)

querydsl-core/src/main/java/com/querydsl/core/support/ReactiveFetchableQueryBase.java
riba2101: yeah the Javadoc wasn't changed - did this

querydsl-r2dbc/pom.xml
```diff
@@ -0,0 +1,226 @@
...
+        </dependency>
+
+        <dependency>
+            <groupId>io.projectreactor</groupId>
+            <artifactId>reactor-test</artifactId>
+            <version>${reactor.version}</version>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
```
mp911de, line 87: Nit: Please rather import `r2dbc-bom` for dependency management. This way, you don't require to specify individual driver versions.

riba2101: yeah, did it already locally -> this is a heavy WIP :)

riba2101:
Hey guys,

I'm still at this but lately didn't have time at all.
I've kinda hit a hard wall -> R2DBC does not support java.util.Date/Time/java.sql.Date...(except Postgres) and to do that I would need to add to BeanPath the create methods for java.time API which comes in java 8, so would need to upgrade the java version which would be a major change? 

what now? halt this and wait for querydsl 5? do the java upgrade? get a beer? 

BR

mp911de:
R2DBC uses JSR-310 types. You should be able to wrap the `Result`/`Row` to return plain old dates. In a similar fashion, wrapping `Statement` to pass plain old dates as arguments. Your wrappers would then handle the conversion.

riba2101:
If I do it like that querydsl would only support plain old dates, or we would have a double conversion (i did this). Then stumbled on the next problem(s):

1. the whole date support of querydsl is not supported, but mapping is working
2. Drivers don't have an ObjectCodec but rather throw a IllegalArgumentException - "codec not found"
3. Drivers don't support partial SQL dates, only the full format - for example if I say `YEAR(some_year_col)` and get back 1980, the driver(h2 for example) throws an exception that it can't parse a date because it expects the ISO format or falls back to the yyyyMMdd format(this is a problem with the driver)

mp911de:
Can you file a bug report against R2DBC H2? The next service release for r2dbc-h2 is scheduled for next week so we can include fixes until then.

riba2101:
yeah ofc, wanted to do it tomorrow. unfortunately, it is not the only bug I've found and am collecting them (the r2dbc support is close to stable now).

riba2101:
hi guys, 

so, added a lot of stuff, found various r2dbc (driver) bugs -> will open them shortly
there are a lot of changes i know, but for the sake of pulling this trough, added an example in r2dbc

known issue: posgress is swallowing inserts

jwgmeligmeyling:
Hi @riba2101, any new status updates? Anything where we can help?

riba2101:
Hi @jwgmeligmeyling.

actually it is "done". There is an example spring project. PR needs to be checked and I need to open various R2DBC bugs.
For the time being Postgres support is disabled because some inserts get swallowed up.

Meaningful docs need to be written, but tbh I'm kinda flooded with work

jwgmeligmeyling (changes requested):
* Branch should be up to date with develop
* `maven-enforcer-plugin` should be reenabled.
   * This isn't going to cause issues, because the current develop targets a 5.x release which allows for breaking API changes.
* All the unnecesairy visibility changes should be undone, PR should be reduced to the bare minimum
* R2DBC dependencies should be removed from core, interfaces that rely on it should be moved to the r2dbc module.

riba2101:
hmmm the tests are tunning for 5h already, no output, can someone check?

aoudiamoncef:
Hi @jwgmeligmeyling @riba2101,

IMHO it could be better to make  `querydsl-core` generic using [Reactive Streams](https://www.reactive-streams.org/) in order to have multiple [implementations ](https://en.wikipedia.org/wiki/Reactive_Streams) like Reactor, RxJava..etc

riba2101:
this has been my point and a lot of ground work was already been done, thus the size of the pr, but then again, the version 5 requires another update and since this is already slow, i wanted to see the changes first in the version 5 and get a go ahead for such invasive changes.

would be also nice to get help since its already huge

aboutZZ:
Any updates of this PR?

riba2101:
as already mentioned, im kinda super busy lately and would appreciate help!

chenzhenjia:
Is querydal no longer maintained?

PR #2627, opened 2020-07-19
Prevent ExtendedBeanSerializer from generating 2 toString methods

### LINKED ISSUE #2467: ExtendedBeanSerializer generates a second toString if addToString is set to true
BeanSerializer generates toString if configured to do so (addToString is true), while ExtendedBeanSerializer always does it, in line with the JavaDoc. 

If ExtendedBeanSerializer is used with addToString=true - which can happen when someone moves to ExtendedBeanSerializer from BeanSerializer, this results 2 toString methods in the generated beans and failed compilation.

ilitosh:
fixes #2467

### DIFF
--- a/querydsl-sql-codegen/src/main/java/com/querydsl/sql/codegen/ExtendedBeanSerializer.java
+++ b/querydsl-sql-codegen/src/main/java/com/querydsl/sql/codegen/ExtendedBeanSerializer.java
@@ -54,7 +54,6 @@ public class ExtendedBeanSerializer extends BeanSerializer {
 
         StringBuilder anyColumnIsNull = new StringBuilder();
         StringBuilder columnEquals = new StringBuilder();
-        StringBuilder toString = new StringBuilder();
         List<String> properties = new ArrayList<String>();
         for (PrimaryKeyData pk : primaryKeys) {
             for (String column : pk.getColumns()) {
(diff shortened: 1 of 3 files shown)


ilitosh:
CUBRID tests are failing, not related to the changes..

querydsl-sql-codegen/src/main/java/com/querydsl/sql/codegen/ExtendedBeanSerializer.java
```diff
@@ -112,4 +112,9 @@ protected void bodyEnd(EntityType model, CodeWriter writer) throws IOException {
 
     }
 
+    @Override
+    // In this class toString is added within bodyEnd, override with empty to ensure the base class does not add another
+    // toString
+    protected void addToString(EntityType model, CodeWriter writer) {
```
jwgmeligmeyling, line 118: This seems a bit hacky to me, can't we move the code from the `bodyEnd()` to this overridden `addToString`? Is the implementation different from the default anyway? The toString in `bodyEnd()` doesn't really seem that sophisticated. Rather more a debugging leftover...

jwgmeligmeyling:
Small remark, I'd like to see the code cleaned up a bit. Don't bother about the CUBRID tests. They fail every now and then.

querydsl-sql-codegen/src/main/java/com/querydsl/sql/codegen/ExtendedBeanSerializer.java
ilitosh: I tried doing this initially, but unfortunately `toString` is being built along with collecting properties / pks within `bodyEnd`. Extracting this would either result duplication of this logic or keeping the state between the method calls. Agree this solution is hacky but best I could come up without rewriting it too much and preserving the format at the same time.

The implementation is different, yes, `ExtendedBeanSerializer` only includes the primary keys into `toString`. And I cannot move this up to `BeanSerializer` either as it should not be aware of PKs.

toString example using the unit test:
```
BeanSerializer: id = 1, name = X, sub_id = ##
ExtendedBeanSerializer: DomainClass#1;##
```

jwgmeligmeyling:
I pushed a change, I think I like this approach better.

PR #2632, opened 2020-08-11
serialize BEFORE_FILTERS flags for update queries

mqs24d:
Enable developers to use Postgres update x set ...=... from syntax with BEFORE_FILTERS flag - for example:
```
 db.update(table)
      .set(table.f1, table2.f2)
      .addFlag(Position.BEFORE_FILTERS, table.getTableName() + " " + table + " from " + table2.getTableName() + " " + table2)
      .where(
        table.fkField.eq(table2.pkField)
        ...
      )
      .execute();
```

### DIFF
--- a/querydsl-sql/src/main/java/com/querydsl/sql/SQLSerializer.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/SQLSerializer.java
@@ -651,6 +651,7 @@ public class SQLSerializer extends SerializerBase<SQLSerializer> {
         }
         skipParent = false;
 
+        serialize(Position.BEFORE_FILTERS, metadata.getFlags());
         if (metadata.getWhere() != null) {
             serializeForWhere(metadata);
         }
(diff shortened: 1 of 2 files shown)


jwgmeligmeyling:
The code looks logical. However, your example uses `AFTER_SELECT` and your implementation uses `BEFORE_FILTERS`. Is this an oversight? Also, could you add a simple rendering test case?

mqs24d:
> The code looks logical. However, your example uses `AFTER_SELECT` and your implementation uses `BEFORE_FILTERS`. Is this an oversight? Also, could you add a simple rendering test case?

Sorry thats just a typo in the example - i fix it right away. But the code is correct! I try my best to add a testcase in the next days.

jwgmeligmeyling:
Ok! If it has a test included I'll merge it, I've put it on the planning for inclusion for the next milestone.

mqs24d:
i just added the test. When is the next milestone planned?

PR #2636, opened 2020-08-13
[#2380] Added configuration for gradle incremental annotation processing

### LINKED ISSUE #2380: support for incremental annotation processing in gradle 4.7+
Since version 4.7 gradle allows annotation processors to opt-in for incremental annotation processing. Without it any change in a project with a dependency on querydsl results in full rebuild, including the generation of Q classes. Adapting this would reduce build times considerably.

[Relevant  documentation ](https://docs.gradle.org/4.7-rc-2/userguide/java_plugin.html#sec:incremental_annotation_processing)

The gradle team is providing support to implement this for popular projects relying on annotation processors. Originated in this [ticket](url).

ebbing:
We are using this change for six months in a quite big project and it works fine for us.

Fixes #2380

### DIFF
--- a/querydsl-apt/src/apt/jpa/META-INF/gradle/incremental.annotation.processors
+++ b/querydsl-apt/src/apt/jpa/META-INF/gradle/incremental.annotation.processors
@@ -1,0 +1,1 @@
+com.querydsl.apt.jpa.JPAAnnotationProcessor,isolating


natros:
that's only for JPA

ebbing:
that's right

Am Do., 13. Aug. 2020 um 09:52 Uhr schrieb Filipe Sousa <
notifications@github.com>:

miensol:
@idosal @johnktims @pslootweg @robertandrewbain @Shredder121 @timowest Any chance you'll have a time to review the changes?

jwgmeligmeyling:
Many of our contributors are not Gradle experts. I see concerns that this only works for JPA. I have no confirmation that it will still generate code for SQL or any of our other packages at all.

For this to be integrated, we need test cases or at least some sort of proof that everything still works. And not just for JPA, but also the other packages. Currently we don't build with gradle.

As it is, this PR is unacceptable IMO.

jwgmeligmeyling (changes requested):
Needs test

natros:
Mapstruct enabled annotation incremental compilation and they were able to test with maven using gradle-test-kit and gradle-tooling-api. 

https://github.com/mapstruct/mapstruct/blob/master/integrationtest/src/test/java/org/mapstruct/itest/tests/GradleIncrementalCompilationTest.java

miensol:
@jwgmeligmeyling Would you mind sharing the reasons why you would prefer having the gradle annotation processing cache enabled for other processors?

I might be missing something so please clarify. Releasing this change, assuming it is tested properly, would improve the experience for gradle + jpa users. It would not harm SQL users.

jwgmeligmeyling:
> Would you mind sharing the reasons why you would prefer having the gradle annotation processing cache enabled for other processors?

I don't.

@natros pointed out that this only fixes the issue for JPA. The module in which this file is created, also supports JDO, JPA, Morphia and Spring Roo. I want to make sure that with these change these metamodels are not affected. It would be great if these generators also supported incremental annotation processing, but my goal here really is that they keep working and that JPA metamodel generation is not unintentionally triggered for projects that do not want to use QueryDSL JPA.

If there are tests or some example project that shows that both SQL and JPA metamodel generation just works fine with the change applied, thats enough for me. But as a non-gradle user, I need some example to verify this change.

NicklasWallgren:
@ebbing Have you made the artifact of this PR available somewhere? I'm eager to try it out.

PR #2643, opened 2020-09-05
[#2025] querydsl-sql : Change Joda-Time to optional

### LINKED ISSUE #2025: querydsl-sql : Joda time should be optional
I am using Java 8 so I don't want Joda time in my classpath.
Indeed it messes with the imports: if I want to use a `java.time.LocalDate`, I have to be careful not to import `org.joda.time.LocalDate`.

Currently if I try to exclude Joda time from my classpath, I cannot initialize Querydsl because in `JavaTypeMapping`, the `CalendarType` is registered and it depends on Joda time. 
Moreover:
- still in `JavaTypeMapping`, Joda time types are registered,
- the `Configuration` class is final and reference `JavaTypeMapping`, so there is no way around.

In a first step, the modifications that needs to be done are:
- Refactor `AbstractDateTimeType` to remove the Joda time dependency,
- In `JavaTypeMapping`, do the joda time initialization the same way java 8 types are registered,
- Mark joda time as optional in the pom.xml :)

If that is ok, I can do the change.
However, it is a breaking change: libraries and projects depending on querydsl-sql and using joda time will have to explicitly add joda time in their pom.xml, gradle etc.
So the version on which this development will be made has to be decided.

daniel-shuy:
Fixes #2025 

This is a breaking change as it requires users who are using Joda-Time to explicitly declare it as a dependency.

As I mentioned in https://github.com/querydsl/querydsl/issues/2025#issuecomment-687620041, there are multiple approaches to replace Joda-Time's [DateTimeFormatter](https://www.joda.org/joda-time/apidocs/org/joda/time/format/DateTimeFormatter.html), each with their own tradeoffs. This PR uses the [SimpleDateFormat](http://docs.oracle.com/javase/6/docs/api/java/text/SimpleDateFormat.html) with `ThreadLocal` approach.


querydsl-sql/src/main/java/com/querydsl/sql/types/AbstractJodaTimeDateTimeType.java
```diff
@@ -0,0 +1,35 @@
...
+
+import java.util.Calendar;
+import java.util.TimeZone;
+
+/**
+ * Common abstract superclass for Type implementations for Joda-Time
+ *
+ * @param <T>
+ */
+public abstract class AbstractJodaTimeDateTimeType<T> extends AbstractType<T> {
```
jwgmeligmeyling, line 14: Why not:

```suggestion
public abstract class AbstractJodaTimeDateTimeType<T> extends AbstractDateTimeType<T> {
```

here?

jwgmeligmeyling, line 14: Why did you opt to create a new `AbstractJodaTimeDateTimeType` rather than modify the existing `AbstractDateTimeType` - which was, and still is, dependent on Joda date/time?

The change of signature for the DateTimeType implementation leads to a backwards compatiblity issue:

```
Class com.querydsl.sql.types.AbstractJodaTimeDateTimeType
4323
 Added Class , access abstract public super synchronized
4324
Class com.querydsl.sql.types.DateTimeType
4325
 Changed Class , access 
4326
Class com.querydsl.sql.types.LocalDateTimeType
4327
 Changed Class , access 
4328
Class com.querydsl.sql.types.LocalDateType
4329
 Changed Class , access 
4330
Class com.querydsl.sql.types.LocalTimeType
4331
 Changed Class , access 
4332
[WARNING] Rule 0: org.semver.enforcer.RequireBackwardCompatibility failed with message:
```

querydsl-sql/pom.xml
```diff
@@ -35,6 +35,7 @@
       <groupId>joda-time</groupId>
       <artifactId>joda-time</artifactId>
       <version>1.6</version>
+      <optional>true</optional>
```
jwgmeligmeyling, line 38: Do you see the possibility to bump the version of `joda-time` as well? In the issue some users also report version conflicts.

querydsl-sql/src/main/java/com/querydsl/sql/types/AbstractJodaTimeDateTimeType.java
daniel-shuy: I modified [AbstractDateTimeType](https://github.com/querydsl/querydsl/pull/2643/files#diff-9255a775bc0f53143b864924632350eb) to no longer depends on Joda time. I created `AbstractJodaTimeDateTimeType` to replace `AbstractDateTimeType` as the parent class for the Joda time converters.

The change of signature for `DateTimeType` is not actually a breaking change, as `DateTimeType` now extends `AbstractJodaTimeDateTimeType`, which itself also extends `AbstractDateTimeType`. Even though the semver enforcer plugin flags it as a breaking change, we will need to increment the major version anyways as this PR itself is a breaking change (users that use Joda time will have to explicitly declare it as a dependency).

querydsl-sql/pom.xml
daniel-shuy: Sure, if the QueryDSL team approves it

daniel-shuy:
@idosal @johnktims @pslootweg @robertandrewbain @Shredder121 @timowest 
hi, could one of you please help to review the PR? Thanks

daniel-shuy:
@idosal @johnktims @pslootweg @robertandrewbain @Shredder121 @timowest bump

> 
> 
> @idosal @johnktims @pslootweg @robertandrewbain @Shredder121 @timowest
> hi, could one of you please help to review the PR? Thanks
@idosal @johnktims @pslootweg @robertandrewbain @Shredder121 @timowest bump

querydsl-sql/src/main/java/com/querydsl/sql/types/AbstractDateTimeType.java
```diff
@@ -38,14 +37,40 @@ protected static Calendar utc() {
         return (Calendar) UTC.clone();
     }
 
-    protected static final DateTimeFormatter dateFormatter = DateTimeFormat.forPattern("yyyy-MM-dd");
+    private static final ThreadLocal<SimpleDateFormat> dateFormatter = new ThreadLocal<SimpleDateFormat>() {
```
jwgmeligmeyling, line 40: We're setting the minimal JRE requirement to 8 in the upcoming QueryDSL release. As a result, I played with swapping out the `SimpleDateFormat` here with Java 8's `DateTimeFormatter`. However, that only works with `Temporals` and not simple `Date` objects. Converting a `Date` to a `Temporal` seems rather involved.

@daniel-shuy WDYT? Should we swap the implementation out for  `DateTimeFormatter`?

jwgmeligmeyling (approved):
The changes are excellent. Thanks very much for this often requested feature 👍🏻

querydsl-sql/src/main/java/com/querydsl/sql/types/AbstractDateTimeType.java
daniel-shuy: > We're setting the minimal JRE requirement to 8 in the upcoming QueryDSL release

Nice! I've rebased this PR onto the `java-8-2` branch.

> As a result, I played with swapping out the SimpleDateFormat here with Java 8's DateTimeFormatter. However, that only works with Temporals and not simple Date objects. Converting a Date to a Temporal seems rather involved.

`java.util.Date` can be easily converted to `java.time.Instant` using [Date#toInstant()](https://docs.oracle.com/javase/8/docs/api/java/util/Date.html#toInstant--). `Calendar` also has an equivalent [Calendar#toInstant()](https://docs.oracle.com/javase/8/docs/api/java/util/Calendar.html#toInstant--) method.

`java.sql.Timestamp`, `java.sql.Date` and `java.sql.Time` also have [Timestamp#toLocalDateTime()](https://docs.oracle.com/javase/8/docs/api/java/sql/Timestamp.html#toLocalDateTime--), [Date#toLocalDate()](https://docs.oracle.com/javase/8/docs/api/java/sql/Date.html#toLocalDate--) and [Time#toLocalTime()](https://docs.oracle.com/javase/8/docs/api/java/sql/Time.html#toLocalTime--) to convert to `LocalDateTime`, `LocalDate` and `LocalTime` respectively.

> WDYT? Should we swap the implementation out for DateTimeFormatter?

I've added a commit that replaces `SimpleDateFormat` with Java 8's `DateTimeFormatter`, you can take a look. If we decide not to use it I can simply drop the commit. One of the advantages is that it allows [AbstractJSR310DateTimeType](https://github.com/querydsl/querydsl/pull/2643/commits/e19073171abfcf4caaf6894eb7eb652a53f04b8b#diff-4cc25c794652fea9373b3da85194ed1817806a7637a43a1f9d76d8e96af9dccc) to extend [AbstractDateTimeType](https://github.com/querydsl/querydsl/pull/2643/commits/e19073171abfcf4caaf6894eb7eb652a53f04b8b#diff-3237b2c366f99970fac048ab8af05274f5e12b4377789411a8d0d1aceaf3dac9).

jwgmeligmeyling: Thanks! I agree we should include these changes. I also like how `AbstractJSR310DateTimeType` can now extend `AbstractDateTimeType`.

I see the tests is failing. The CUBRID tests regularly fail (Ugh), so you don't need to check those. The other ones I still have to look into. Unless you're already working on them?

daniel-shuy: Sorry, I've fixed it. Turns out `Instant` has to be converted to `LocalDateTime` first for it to be formatted using `DateTimeFormatter`.

Some tests are still failing, but I think they are existing issues in the `java-8-2` branch.

jwgmeligmeyling: Thanks! Ill look into these and merge the PR tomorrow.

jwgmeligmeyling:
Thanks for your contribution! 👍🏻

PR #2657, opened 2020-10-03
issue 2460: use Generated annotation with CLASS retention

### LINKED ISSUE #2242: Annotation processing breaks with Java 9
Related issue:
https://github.com/google/auto/issues/560

Solution is to use type mirrors correctly:
* javax.annotation.Generated for pre java9
* javax.annotation.processing.Generated for java9 onwards.

### LINKED ISSUE #2460: Improvement: @Generated annotation to mark code that should be ignored by JaCoCo
Hi,
It would be nice to have an "@Generated" annotation with class retention policy that can be ignored by code coverage tools like [JaCoCo](https://www.jacoco.org/jacoco/). This should work similar to how it's done for [Immutables](https://github.com/jacoco/jacoco/issues/656) and [Lombok](https://github.com/jacoco/jacoco/pull/495).
Thanks in advanced.

### LINKED ISSUE #2612: Add support for Jakarta annotations + Jakarta persistence (in jakarta.* packages)
Currently, when building a project with QueryDSL-APT 4.3.1 and jakarta.annotation-api 2.0.0-RC1, no Q-classes are generated (with an error message like "java.lang.NoClassDefFoundError: javax/annotation/Generated"). It still works with jakarta.annotation-api 1.3.5.

With Jakarta EE, many types in the javax.* package are moved to the jakarta.* packages.
This includes the Annotation types in javax.annotations, and it looks like [javax.annotation.Generated](https://jakarta.ee/specifications/annotations/1.3/apidocs/index.html?javax/annotation/Generated.html) is now [jakarta.annotation.Generated](https://jakarta.ee/specifications/annotations/2.0/apidocs/index.html?jakarta/annotation/Generated.html).

The same happens when using jakarta.persistence-api 3.0.0-RC2 (instead of 2.2.3), here the problem is javax.persistence.Entity → jakarta.persistence.Entity.

rob-valor:
I added the `@Generated` annotation to the querydsl-core module because only the core and jpa modules are needed in an application using QueryDsl.

I also wanted to add `<property name="illegalClasses" value="javax.annotation.Generated"/>` to [IllegalImport](https://github.com/querydsl/querydsl/blob/182de7fe5d7521ff34c9a87670ec295bb791f782/src/main/resources/querydsl_checks_base.xml#L43) but I guess the version of checkstyle is to low because it is a [documented property](https://checkstyle.sourceforge.io/config_imports.html#IllegalImport_Examples) but failed to be accepted.

Fixes #2460
Fixes #2612
Fixes #2242


rob-valor:
Not clear why that build fails on missing `@Param`in Javadoc which was added in 2013

jwgmeligmeyling:
Hi @rob-valor ,

I want to integrate this patch, but I would like the transition as smooth as possible, as the latest releases have caused enough trouble for our users.

Do you see any way of making the used `Generated` annotation parameterizable so that the default annotation remains `javax.annotation.Generated`? (At least until a next major release, but hopefully we can integrate this sooner).

rob-valor:
@jwgmeligmeyling Since this is the first time OSS (code) contribution for me and I changed a lot of files that I have to "undo", what is the best approach for the current pull request? Do I just delete my fork and start over again? (Don't know how Github and this pull request handles this). A rollback of this current change would pollute the git history, no?

idosal:
Hey @rob-valor, thanks for the contribution. The PR will be squashed so the commit history is irrelevant. You can choose whichever option is most comfortable. :)

rob-valor:
@idosal I have some questions:

1. Is there any objection to _write_ the annotation as fully qualified (like `@javax.annotation.Generated("...")`)? This makes injection of class as String in the `Serializer`easier than having to parse the String for import and use simple class name as annotation on the class.
2. I would just add a boolean to the `AptOptions` to use the _`CLASS` retained annotation_ and if not set use `javax.annotation.Generated` for java8 and earlier and `javax.annotation.processing.Generated` for java9 and newer (depending on which one can be loaded with a `Class.forName`). Is that ok?
3. about `BeanSerializer`, `ExtendedBeanSerializer` and `GroovyBeanSerializer` because they can be also instantiated in [AbstractModule#get(Class, String)](https://github.com/querydsl/querydsl/blob/ce90e59c6fd4a5cba4cb3b587b6f3cc34d5a3f95/querydsl-codegen/src/main/java/com/querydsl/codegen/AbstractModule.java#L95):
I'm adding the _annotation_ class as a fully qualified class name (String) because instantiation in that `AbstractModule.get` method does not support injecting a `Class` (it binds only _instances_ of a specific class). Those 3 serializers seem to be instantiated with the default constructor. To support that annotation class to be injected in the serializer constructor I need a constructor with a String parameter. Now `BeanSerializer` and `GroovyBeanSerializer` already have an overloaded constructor with a String parameter `javadocSuffix` which doesn't seem to be used by the framework. I could add a constructor with 2 parameters and add the default value of `javadocSuffix` or `propertyAnnotations` (depending on which one used as 2nd parameter) as binding in [CodegenModule#configure](https://github.com/querydsl/querydsl/blob/ce90e59c6fd4a5cba4cb3b587b6f3cc34d5a3f95/querydsl-codegen/src/main/java/com/querydsl/codegen/CodegenModule.java#L57). Is that ok?

jwgmeligmeyling:
Hi @rob-valor , I don't have any objections to your suggestion. Perhaps the approach suggested in 3 will work for 2 as well (for consistency). All's fine!

rob-valor:
I think this should be _it_. (but again with that javadoc failing :disappointed:)

There is however `SpatialRefSys` and `QSpatialRefSys` which have also the `javax.annotation.Generated` annotation. I'm not sure why its added to the querydsl source code but **I think those should not be there**.  Also it looks like these are added by `BeanSerializer` and `MetaDataSerializer` which have dynamic class name for annotation now.

jwgmeligmeyling:
Thanks @rob-valor , I'll have a look and integrate it ASAP.

I'll fix the Javadoc issues myself. Thats not a major issue.

There's no need to port this to the Groovy and Scala processors yet. Its obviously nice to have but these modules are much less frequently used.

querydsl-apt/src/main/java/com/querydsl/apt/DefaultConfiguration.java
```diff
@@ -255,7 +255,9 @@ public DefaultConfiguration(
         } else {
             variableNameFunction = DefaultVariableNameFunction.INSTANCE;
         }
-        module.bind(CodegenModule.VARIABLE_NAME_FUNCTION_CLASS, variableNameFunction);
```
jwgmeligmeyling, line 258: This line is incorrectly removed.

querydsl-codegen/src/main/java/com/querydsl/codegen/GeneratedAnnotationResolver.java
```diff
@@ -0,0 +1,49 @@
...
+     *
+     * @return the Generated annotation class from java. Never {@code null}.
+     */
+    public static String resolveDefault() {
+        return DEFAULT_GENERATED_ANNOTATION_CLASS;
+    }
+
+    private static String resolveJavaDefault() {
+        try {
+            return Class.forName("javax.annotation.processing.Generated").getName();
```
jwgmeligmeyling, line 42: As an extra option, we could also try `jakarta.annotation.Generated`.

querydsl-apt/src/main/java/com/querydsl/apt/DefaultConfiguration.java
```diff
@@ -255,7 +255,9 @@ public DefaultConfiguration(
...
+
+        String generatedAnnotationClass = GeneratedAnnotationResolver.resolve(options.get(QUERYDSL_GENERATED_ANNOTATION_CLASS));
+        module.bind(CodegenModule.GENERATED_ANNOTATION_CLASS, generatedAnnotationClass);
```
jwgmeligmeyling, line 260: I think we could actually use type `Class<? extends Annotation>` here and make it ourselves a little easier in the codgen changes.

querydsl-core/src/main/java/com/querydsl/core/annotations/Generated.java
```diff
@@ -0,0 +1,42 @@
...
+ * It uses <code>CLASS</code> retention to allow byte code analysis tools like
+ * Jacoco detect this code as being generated.
+ * <p/>
+ * It is now used instead of <code>javax.annotation.Generated</code>.
+ */
+
+@Documented
+@Retention(CLASS)
+@Target(TYPE)
+public @interface Generated {
```
jwgmeligmeyling, line 35: This should either become a last resort default or move to a test package as a proof of concept.

jwgmeligmeyling:
No need to work on the comments. I am already integrating this in my java-11 support branch which will make it into the 5.x.x release.

querydsl-codegen/src/main/java/com/querydsl/codegen/ProjectionSerializer.java
```diff
@@ -58,7 +79,6 @@ protected void intro(EntityType model, CodeWriter writer) throws IOException {
 
         // imports
         writer.imports(NumberExpression.class.getPackage());
-        writer.imports(ConstructorExpression.class, Generated.class);
```
jwgmeligmeyling, line 61: `ConstructorExpression` is removed here accidentally.

rob-valor:
To be clear: I don't have to take any action for these review remarks?

jwgmeligmeyling:
No you don't, I already addressed them. I'll push as soon as all the tests pass.

PR #2658, opened 2020-10-04
Fix Flaky tests in SearchQueryTest

jonathan-stein:
If `mvn test` is run consecutively in the querydsl-hibernate-search without `mvn clean` in between, the following unit tests will fail:

1. SearchQueryTest#count
2. SearchQueryTest#listResult
3. SearchQueryTest#paging

This occurs because the actual directory created by the unit tests is `target/lucene` not `target/lucene3`.

### DIFF
--- a/querydsl-hibernate-search/src/test/java/com/querydsl/hibernate/search/AbstractQueryTest.java
+++ b/querydsl-hibernate-search/src/test/java/com/querydsl/hibernate/search/AbstractQueryTest.java
@@ -38,7 +38,7 @@ public abstract class AbstractQueryTest {
     @BeforeClass
     public static void setUpClass() throws IOException {
         FileUtils.delete(new File("target/derbydb"));
-        FileUtils.delete(new File("target/lucene3"));
+        FileUtils.delete(new File("target/lucene"));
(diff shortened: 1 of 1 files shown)


jwgmeligmeyling (approved):

jwgmeligmeyling:
Thanks! 👍

jonathan-stein:
Hi, I saw that this change was later reverted. I saw that some of the CUBRID tests were failing, and was wondering if anyone ever figured out how that was related to this change?

jwgmeligmeyling:
This test was reverted because it turned out to break CI. Turns out the `lucene3` folder does need to be removed - at least in some scenario's. I didn't have time to further look into it.

The CUBRID tests just fail from time to time, it seems to be because the database containers don't boot up properly. I am actually considering to drop the CUBRID tests because they are incredibly annoying.

jonathan-stein:
Would it be a good idea to delete both `lucene` and `lucene3` in these tests then?

jwgmeligmeyling:
I cant find the failing test anymore, Ill reapply the patch. Maybe it were just these annoying cubrid tests after all...

PR #2669, opened 2020-10-21
Remove ImmutableList usage in *Template constructors

yorlov:


jwgmeligmeyling:
I am working to eliminate Guava as a dependency altogether (#2324), I'll merge in this patch as part of it.

jwgmeligmeyling (approved):
I am working to eliminate Guava as a dependency altogether (#2324),

querydsl-core/src/main/java/com/querydsl/core/types/PredicateTemplate.java
```diff
@@ -32,7 +31,7 @@
     @Nullable
     private transient volatile Predicate not;
 
-    protected PredicateTemplate(Template template, ImmutableList<?> args) {
+    protected PredicateTemplate(Template template, List<?> args) {
         super(Boolean.class, template, args);
```
jwgmeligmeyling, line 35:
```suggestion
        super(Boolean.class, template, Collections.unmodifiableList(new ArrayList<>(args)));
```

jwgmeligmeyling: Currently we use `ImmutableList` as an efficient way to ensure immutability of expression types. Using a plain `List`, we can only ensure this by making an unmodifiyable copy (to guard against copies of the original as well). This takes a performance penalty.

jwgmeligmeyling (changes requested):
After giving this second thoughts: we shouldn't apply this patch.

Templates are expressions. Expressions ought to be immutable. Lots of assumptions are being made on this immutability in QueryDSL. Expressions rely on immutable collections to guarantee their immutability. The only way of ensuring this with regular collections is by making a local copy, which adds a performance penalty.

Lastly, there is little need for users to interact with ` PredicateTemplate` directly. Templates are a pretty staightforward thing and I can't imagine a case where you'd want to extend it. Instead, you probably want to extend `ExpressionBase`.

Also for simply instantiating Template expressions, you should use the static factory methods in `Expressions`. There have variants that accept a regular List as argument, i.e. `Expressions.stringTemplate(String, List)`.

What was your use case for these overloads?

PR #2672, opened 2020-10-26
Target Java 8 for compilation

jwgmeligmeyling:
_This pull request is a work in progress. I am also awaiting the very first CI results. I expect a couple of issues still._

# Changelog

### Breaking changes

* Java 8 minimal requirement. If you still rely on Java <7, please use the latest 4.x.x version.
* Removed bridge method that were in place for backwards compatibility of legacy API's. This may lead to some breaking API changes.

### New features

* Added `Fetchable#stream()` which returns a `Stream<T>`. _Make sure that the returned stream is always closed to free up resources, for example using try-with-resources. It does **not** suffice to rely on a terminating operation on the stream for this (i.e. `forEach`, `collect`)._

### Dependency updates

* `cglib` to 3.3.0 for Java 8+ support
* `asm` to 9.0 for Java 8+ support
* DataNucleus 5.2.x for Java 8+ support
  * JDO now uses `org.datanucleus:javax.jdo` instead of `javax.jdo:jdo-api`. Seems more widely used and also got more recently updated.

### Plans

- [x] Require Java 8
- [x] Optimize code for Java 8 (use new API's)
- [x] Switch out Guava's Optional with Java 8's `Optional`.
- [x] Add `Fetchable#stream`
- [x] Add tests for `Fetchable#stream` (in particular Hibernate and plain JPA)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -61,13 +61,10 @@
     <guava.version>18.0</guava.version>
     <codegen.version>0.6.8</codegen.version>
     <mysema.lang.version>0.2.4</mysema.lang.version>
-    <cglib.version>2.2.2</cglib.version>    
-    <findbugs.version>1.3.2</findbugs.version>
+    <cglib.version>3.3.0</cglib.version>
     <slf4j.version>1.6.1</slf4j.version>
-    <surefire.version>2.18</surefire.version>
+    <surefire.version>2.22.2</surefire.version>
     <animal-sniffer.version>1.18</animal-sniffer.version>
-
-    <jdo.version>3.0.1</jdo.version>
     <morphia.version>1.3.2</morphia.version>
     
     <!-- Import-Package definitions for maven-bundle-plugin -->
@@ -141,6 +138,11 @@
         <groupId>javax.inject</groupId>
         <artifactId>javax.inject</artifactId>
         <version>1</version>
+      </dependency>
+      <dependency>
+        <groupId>org.datanucleus</groupId>
+        <artifactId>javax.jdo</artifactId>
+        <version>3.2.0-m13</version>
       </dependency>
     </dependencies>
   </dependencyManagement>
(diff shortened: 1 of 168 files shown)


querydsl-collections/src/test/java/com/querydsl/collections/NumberTest.java
```diff
@@ -15,8 +15,8 @@
     @Test
     public void sum() throws Exception {
         NumberPath<BigDecimal> num = Expressions.numberPath(BigDecimal.class, "num");
-        CollQuery<?> query = CollQueryFactory.from(num, Arrays.asList(new BigDecimal("1.6"), new BigDecimal("1.3")));
+        CollQuery<BigDecimal> query = CollQueryFactory.<BigDecimal> from(num, Arrays.<BigDecimal> asList(new BigDecimal("1.6"), new BigDecimal("1.3")));
```
jwgmeligmeyling, line 18: `CollQueryFactory` suffers from https://stackoverflow.com/a/32294457/2104280.

This is technically not a breaking API change: users that already were on Java 8, struggled with the exact same issue in their own codebase alrady.

querydsl-core/pom.xml
```diff
@@ -48,17 +48,6 @@
     </dependency>
     
      <!-- backwards compatibility -->
-    <dependency>
-      <groupId>com.infradna.tool</groupId>
```
jwgmeligmeyling, line 52: `bridge-method-annotation` is unmaintained and doesn't support Java 8 bytecode.

The bridge methods are also not needed anymore: its safe to remove backwards compatibility methods in a major release.

pom.xml
```diff
@@ -221,12 +223,12 @@
...
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-assembly-plugin</artifactId>
-          <version>2.5.4</version>
+          <version>3.3.0</version>
         </plugin>
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-compiler-plugin</artifactId>
-          <version>3.3</version>
+          <version>3.8.1</version>
```
jwgmeligmeyling, line 231: This wasn't causing issues, but I don't think it hurts to keep this up-to-date for the best compatiblility.

pom.xml
```diff
@@ -221,12 +223,12 @@
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-assembly-plugin</artifactId>
-          <version>2.5.4</version>
+          <version>3.3.0</version>
```
jwgmeligmeyling, line 226: This wasn't causing issues, but I don't think it hurts to keep this up-to-date for the best compatiblility.

PR #2673, opened 2020-10-26
Make Alias.* thread-safe

### LINKED ISSUE #2671: QueryDSL not thread-safe, Alias.$(alias.getId()) returns null
I have been using QueryDSL in various projects over the years. Now I have two concrete projects that are struggling with a severe bug issue that QueryDSL:
As it seems `Alias` stuff has a concurrency problem leading to `Alias.$` function to return `null` where it should not.
That bug only occurrs in complex multi-threaded batch processing situations and is not so easy to reproduce.

General code pattern of the affected query was something like this (nothing really exciting or exotic):
```java
FooEntity foo = Alias.alias(FooEntity.class);
BarEntity bar = Alias.alias(BarEntity.class);
JPAQuery<FooEntity> query = new JPAQuery<FooEntity>(getEntityManager()).from(Alias.$(foo));
query.join(Alias.$(bar)).on(Alias.$(bar.getFoo().getId()).eq(Alias.$(foo.getId())));
...
```

We could even reproduce the bug locally and see in the debugger what is going on:
* In that situation in debugger we could reproducibly see that `Alias.$(foo.getId())` is returning `null`
* We stepped into `foo.getId()` and saw that the is entering this if:
https://github.com/querydsl/querydsl/blob/2bf234caf78549813a1e0f44d9c30ecc5ef734e3/querydsl-core/src/main/java/com/querydsl/core/alias/PropertyAccessInvocationHandler.java#L81
* It also gets the dummy return value (42) from the `propToObj` map (used as cache):
https://github.com/querydsl/querydsl/blob/2bf234caf78549813a1e0f44d9c30ecc5ef734e3/querydsl-core/src/main/java/com/querydsl/core/alias/PropertyAccessInvocationHandler.java#L82
* However, the `propToExpr` map (separate map used as cache for expressions) does not contain the corresponding entry (what seems to be an inconsistent state to me) leading to this putting `null` in the `current` `ThreadLocal` (actually an assertion for not `null` should be added here producing an exception with additional contextual information):
https://github.com/querydsl/querydsl/blob/2bf234caf78549813a1e0f44d9c30ecc5ef734e3/querydsl-core/src/main/java/com/querydsl/core/alias/PropertyAccessInvocationHandler.java#L87
* To further analyze this, we put a conditional breakpoint in `PropertyAccessInvocationHandler` with the expression `propToExpr.size() != propToObj.size()` and came to the conclusion that two different threads where accessing the same CG-Lib proxy instance. As these objects are internally using regular `Map`s they are not thread-safe and IMHO this is by design.
* So we digged our code where we could share Alias CG-Lib proxy objects accross threads. The analysis of our entire code-base gave us the result that we do this nowhere: We always regularily create these `alias` instances and their sub-objects in local variables and never in members, statics, etc.
* Now we digged in QueryDSL and made this observation:
https://github.com/querydsl/querydsl/blob/2bf234caf78549813a1e0f44d9c30ecc5ef734e3/querydsl-core/src/main/java/com/querydsl/core/alias/Alias.java#L349
This code is calling `AliasFactory` on a single shared instance:
https://github.com/querydsl/querydsl/blob/2bf234caf78549813a1e0f44d9c30ecc5ef734e3/querydsl-core/src/main/java/com/querydsl/core/alias/Alias.java#L58
The call leads to this code-line in question:
https://github.com/querydsl/querydsl/blob/2bf234caf78549813a1e0f44d9c30ecc5ef734e3/querydsl-core/src/main/java/com/querydsl/core/alias/AliasFactory.java#L112
As it seems you are using a cache for these proxy-objects shared accross all threads.
https://github.com/querydsl/querydsl/blob/2bf234caf78549813a1e0f44d9c30ecc5ef734e3/querydsl-core/src/main/java/com/querydsl/core/alias/AliasFactory.java#L48
As the Key can easily match accross different threads, I come to the conclusion that QueryDSL is simply not thread-safe and therefore buggy.
Any hints where I could be wrong or confirmations of my observation or anything else what could help?
Thanks for making QueryDSL and thanks in advance for taking care of this issue.

jwgmeligmeyling:
Fixes #2671

Replaces the use of the LoadingCache, so also helps with the removal of Guava.

### DIFF
--- a/querydsl-core/src/main/java/com/querydsl/core/alias/Alias.java
+++ b/querydsl-core/src/main/java/com/querydsl/core/alias/Alias.java
@@ -140,8 +140,16 @@ public final class Alias {
      * @param <D>
      * @return expression
      */
+    @SuppressWarnings("unchecked")
     public static <D> CollectionPath<D, SimpleExpression<D>> $(Collection<D> arg) {
-        return aliasFactory.getCurrentAndReset();
+        final Object current = aliasFactory.getCurrentAndReset();
+        if (arg instanceof CollectionPath) {
+            return (CollectionPath<D, SimpleExpression<D>>) arg; //NOSONAR
+        } else if (arg instanceof ManagedObject) {
+            return (CollectionPath<D, SimpleExpression<D>>) ((ManagedObject) arg).__mappedPath();
+        } else {
+            return (CollectionPath<D, SimpleExpression<D>>) current;
+        }
     }
 
     /**
@@ -212,8 +220,16 @@ public final class Alias {
      * @param <D>
      * @return expression
      */
+    @SuppressWarnings("unchecked")
     public static <D> ListPath<D, SimpleExpression<D>> $(List<D> arg) {
-        return aliasFactory.getCurrentAndReset();
+        final Object current = aliasFactory.getCurrentAndReset();
+        if (arg instanceof ListPath) {
+            return (ListPath<D, SimpleExpression<D>>) arg; //NOSONAR
+        } else if (arg instanceof ManagedObject) {
+            return (ListPath<D, SimpleExpression<D>>) ((ManagedObject) arg).__mappedPath();
+        } else {
+            return (ListPath<D, SimpleExpression<D>>) current;
+        }
     }
 
     /**
@@ -234,8 +250,16 @@ public final class Alias {
      * @param <V>
      * @return expression
      */
+    @SuppressWarnings("unchecked")
     public static <K, V> MapPath<K, V, SimpleExpression<V>> $(Map<K, V> arg) {
-        return aliasFactory.getCurrentAndReset();
+        final Object current = aliasFactory.getCurrentAndReset();
+        if (arg instanceof MapPath) {
+            return (MapPath<K, V, SimpleExpression<V>>) arg; //NOSONAR
+        } else if (arg instanceof ManagedObject) {
+            return (MapPath<K, V, SimpleExpression<V>>) ((ManagedObject) arg).__mappedPath();
+        } else {
+            return (MapPath<K, V, SimpleExpression<V>>) current;
+        }
     }
 
     /**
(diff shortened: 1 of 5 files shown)


querydsl-core/src/main/java/com/querydsl/core/alias/Alias.java
```diff
@@ -140,8 +140,16 @@
...
      * @return expression
      */
+    @SuppressWarnings("unchecked")
     public static <D> CollectionPath<D, SimpleExpression<D>> $(Collection<D> arg) {
-        return aliasFactory.getCurrentAndReset();
+        final Object current = aliasFactory.getCurrentAndReset();
+        if (arg instanceof CollectionPath) {
+            return (CollectionPath<D, SimpleExpression<D>>) arg; //NOSONAR
+        } else if (arg instanceof ManagedObject) {
+            return (CollectionPath<D, SimpleExpression<D>>) ((ManagedObject) arg).__mappedPath();
```
jwgmeligmeyling, line 149: Prefer `__mappedPath()` over the thread local context, for aliases that are shared between threads (i.e. static variables).

querydsl-core/src/main/java/com/querydsl/core/alias/ManagedObject.java
```diff
@@ -22,10 +22,10 @@
...
 
     /**
-     * Returns the underlying {@link EntityPath} instance of this proxy
+     * Returns the underlying {@link Path} instance of this proxy
      *
-     * @return underlying {@link EntityPath} instance
+     * @return underlying {@link Path} instance
      */
-    EntityPath<?> __mappedPath();
+    Path<?> __mappedPath();
```
jwgmeligmeyling, line 29: This was faulty. The `Path` may also be a `SetPath`, `MapPath`, `ListPath` or `CollectionPath`.

querydsl-core/src/main/java/com/querydsl/core/alias/AliasFactory.java
```diff
@@ -42,30 +39,11 @@
...
-
-    private final LoadingCache<Pair<Class<?>,Expression<?>>, ManagedObject> proxyCache =
-        CacheBuilder.newBuilder().build(
-            new CacheLoader<Pair<Class<?>,Expression<?>>,ManagedObject>() {
-                @Override
-                public ManagedObject load(Pair<Class<?>, Expression<?>> input) {
-                    return (ManagedObject) createProxy(input.getFirst(), input.getSecond());
-                }
-            });
+    private final ConcurrentHashMap<Class<?>, Map<Expression<?>, ManagedObject>> proxyCache = new ConcurrentHashMap<>();
```
jwgmeligmeyling, line 42: The ConcurrentHashMap's size is bounded, as it uses classes as key.

PR #2676, opened 2020-10-27
Replace Guava API's with Java 8 ones

### LINKED ISSUE #2324: Get rid of guava

jwgmeligmeyling:
Use standard Java where possible. This makes it easier to ever migrate off of Guava.
- [x] Replace `Lists`, `Sets` and `Maps` with standard copy-constructors. Utility methods like `Lists.newArrayList()` where primarily a thing when Java didn't have diamond operators, which were added in Java 7.
- [x] Replace `Charsets` which `StandardCharsets`, which is part of the JRE since Java 8.
- [x] Replace Guava `Objects.equal` with Java 8's `Objects.equals`
- [x] Replace Guava's `Objects.hashCode` with Java 8's `Objects.hash`
- [x] Replace Guava's `Function` with Java 8's `Function
- [x] Replace `com.google.common.primitives.Primitives`
- [x] Replace `com.google.common.collect.Iterables` (good candidate for Java 8's Stream API?)
- [x] Replace `com.google.common.io.Files`
- [x] Replace `com.google.common.collect.ImmutableList`
- [x] Replace `com.google.common.collect.ImmutableSet`
- [x] Replace `com.google.common.collect.Multimap`
- [x] Replace `com.google.common.io.Resources`
- [x] Replace `com.google.common.base.CharMatcher`
- [x] Replace transitive Guava dependency through codegen 
    - [x] Include codegen as module to QueryDSLs repository through a subtree merge
- [x] Replace transitive Guava dependency through Reflections
    - [x] Upgrade Reflections dependency
- [x] Remove Guava dependency from Maven descriptors 

Fixes #2324


jwgmeligmeyling:
I am reducing the scope of work within this PR slightly. I don't think we should ever "get rid" of Guava.

Currently our expressions are immutable. It is mandatory for these expressions to be immutable, because the way they are used, cached, cloned, etc. Expressions depend on immutable collections to enforce their immutability. In plain Java 8 there are no immutable collection types. There is an unmodifiable collection wrapper, but to prevent changes to the original list from outside, you should also copy a list received from a parameter to a new list. This has to happen between every boundary (for example `Expressions.*` utility methods, to expressions, to mixins). This means a lot of copying overhead.

Which means that we either have to sacrifice safety (non-immutable collections lead to non-immutable expressions, which will lead to hard to debug issues for our users), or sacrifice performance (from all the copying overhead).

I don't see a way to succesfully pull this off without going to Java 11, which is going to be too recent for now.

So unfortunately, it seems Guava will stay around for a while. I'll eliminate as much Guava as I possibly can to make it more feasibile in the future.

I'll also bump the dependency and try to get shaded artefacts working so that any compatibility issues are addressed for those concerned.

PR #2679, opened 2020-10-28
 Return Optional from Fetchable#fetchOne and fetchFirst

jwgmeligmeyling:
Originally these changes were part of the `java-8` branch, but I've decided to maintain a seperate PR for this as I want to discuss the feature in more detail due to its breaking nature (for example for `spring-data`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,26 +1,20 @@
 # Changelog
 
-### 5.0.0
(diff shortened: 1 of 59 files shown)


jwgmeligmeyling:
Closing this PR due to its breaking nature on existing projects and integrations (verified with Spring Data).

Its just not worthwhile.

Issue #2303, opened 2018-05-03
Consider optimising AbstractSQL::newInstance (becomes a bottleneck for small queries)


rbalamohan:
AbstractSQLQuery::fetch becomes a hotspot when queryDSL is used for large number of queries and most of these queries execute in millisecond latencies. When profiling, it was observed that following codepath in "AbstractSQLQuery::fetch" becoming a bottleneck.

<img width="1075" alt="querydsl" src="https://user-images.githubusercontent.com/7969713/39597811-f11e7560-4f59-11e8-91ff-a9cce7782794.png">

`
 		FactoryExpression<T> fe = (FactoryExpression<T>) expr;
                        while (rs.next()) {
                            if (getLastCell) {
                                lastCell = rs.getObject(fe.getArgs().size() + 1);
                                getLastCell = false;
                            }
                            rv.add(newInstance(fe, rs, 0));
                        }
`

For every record retrieved, this ends up creating a new FactoryExpression. There are 2 issues that came up. 

1. newInstance() iterating through the expressions and populating the arguments. This can be improved without looking at the list and checking the map for type related information later.

2. When creating FactoryExpression (e.g FactoryExpressionUtils::newInstance(Object... a)), it ends up invoking "compress()" method which can be optimised as "inner.getArgs" does not change for every record. 

Since the query runs for couple of milliseconds and returns 1000s of rows, above 2 functions become hotspots. It would be good to optimise these methods so that querydsl can handle smaller queries a lot more efficiently.

Shredder121:
Thanks for looking into this.
This sounds interesting to pursue.

rbalamohan:
Thanks @Shredder121 . 

I have posted a PR here https://github.com/querydsl/querydsl/pull/2305. 

It would be awesome, if you could review. :)

jwgmeligmeyling:
Change was reverted due to a regression, because there is no support for it yet it's removed from the 5.x milestone.

PR #2680, opened 2020-10-28
#2303 - Optimising AbstractSQL::newInstance

idosal:
Original issue: #2303


jwgmeligmeyling:
@idosal  is this revert still required?

idosal:
> @idosal is this revert still required?

I need to rerun the tests. If everything passes I think we can introduce the optimization.

Issue #2621, opened 2020-06-26
[suggestion] how about using docker for querydsl test environment?


heesuk-ahn:
In the current querydsl test environment, vagrant is used.

This environment is good, but it requires a number of pre-requisites. (VirtualBox, Puppet, Vagrant..)

If we provide various databases as docker compose, wouldn't it be simpler to build a test environment?

for example,

```
version: '2'
services:
  mysql:
    image: mysql:5.5
    environment:
      MYSQL_USER: querydsl
      MYSQL_PASSWORD: querydsl
      MYSQL_DATABASE: querydsl
      MYSQL_ROOT_PASSWORD: querydsl
    volumes:
      - /var/lib/mysql
    ports:
      - "3306:3306"
  postgre-sql:
    image: postgres:9.1
    ports:
      - "5433:5432"
    environment:
      POSTGRES_USER: querydsl
      POSTGRES_PASSWORD: querydsl
      POSTGRES_DATABASE: querydsl
      POSTGRES_ROOT_PASSWORD: querydsl
    volumes:
      - /var/lib/postgresql/data
.
.
.
(Other necessary database environments...)
```

jwgmeligmeyling:
It's possible. Usually, I just test against the database of my preference (which runs in Docker) and rely on CI for all various databases that are supported.

idosal:
I usually just run the docker-compose file we use in CI locally.

jwgmeligmeyling:
Ah here it is: https://github.com/querydsl/querydsl/blob/master/travis/docker-compose.yml

I guess we could mention it in the README

heesuk-ahn:
Oh, docker-compose already exists! :) 

I'll change it to an issue that mentions the docker-compose environment in the Readme.

johnktims:
I’ve been meaning to update the readme and move the `docker-compose.yml` file to the root of the project. Thanks for the reminder.

PR #2682, opened 2020-10-30
Move docker-compose to root

idosal:
#2621


jwgmeligmeyling (approved):
Changes look great!

Would love to see the section "Vagrant/Puppet setup" replaced in the `README.md` as well. From the original issue we found that even though the containers were already there, users had trouble finding them. Of course we can update the README in a separate PR just fine.

idosal:
I agree. I'll take care of the README tomorrow.

PR #2685, opened 2020-10-31
[#2684] Provide Kotlin example project

### LINKED ISSUE #2684: Provide Kotlin example project
On Stackoverflow we see more and more people attempting to use QueryDSL with Kotlin.

Its actually possible to use Java APT processors with Kotlin, so this should be supported out of the box.

We should provide an example configuration to get users started.

jwgmeligmeyling:
Fixes #2684

In response to https://stackoverflow.com/questions/64615494/querydsl-setup-for-kotlin-with-maven


querydsl-examples/querydsl-example-kotlin-jpa/src/main/kotlin/com/querydsl/examples/kotlin/entity/ExampleEntity.kt
```diff
@@ -0,0 +1,20 @@
+package com.querydsl.examples.kotlin.entity
```
jwgmeligmeyling, line 1: TODO: Add license headers to added files.

PR #2692, opened 2020-11-03
Update doc book reference page

idosal:


querydsl-docs/src/main/docbook/en-US/Querydsl_Reference.xml
```diff
@@ -39,6 +39,14 @@
...
             <surname>Bain</surname>
           </author>
+          <author>
+            <firstname>Ido</firstname>
+            <surname>Salomon</surname>
+          </author>
+          <author>
+            <firstname>Jan-Willem</firstname>
+            <surname>Gmelig Meyling</surname>
+          </author>
```
jwgmeligmeyling, line 49: We should also update `<developers>` in `pom.xml`.

jwgmeligmeyling (approved):

Issue #2395, opened 2018-12-14
PreparedStatement not closed when using fetchOne() or fetchFirst()


rklerke:
hi
As far as i can see the preparedStatement in AbstractSQLQuery initialized in l. 364 and used in l. 370 is never closed in case of exception is thrown when it is being used in l. 370. It should be closed in the finally clause which is already present.

thanks for a great framework!

best regards
Rasmus

jwgmeligmeyling:
The `PreparedStatement` and the `ResultSet` are associated with the `SQLListenerContextImpl` which takes care of closing these resources. Closing happens automatically after the `Iterator` is exhausted in `SQLResultIterator`. In case of an exception, the context is closed by the catch-block.

PR #2698, opened 2020-12-07
Fix AbstractSQLQuery

NikitaKochkurov:
Fix bug https://github.com/querydsl/querydsl/issues/2395?fbclid=IwAR0A5V2Krws9E-BTPyIxl5ttLzMHQu2tzYo_UIleUSWzxnar2ptl10uDPPo

### DIFF
--- a/querydsl-sql/src/main/java/com/querydsl/sql/AbstractSQLQuery.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/AbstractSQLQuery.java
@@ -410,6 +410,8 @@ public abstract class AbstractSQLQuery<T, Q extends AbstractSQLQuery<T, Q>> exte
             throw configuration.translate(queryString, constants, e);
         } catch (RuntimeException e) {
             logger.error("Caught " + e.getClass().getName() + " for " + queryString);
+            onException(context, e);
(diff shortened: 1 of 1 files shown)


querydsl-sql/src/main/java/com/querydsl/sql/AbstractSQLQuery.java
```diff
@@ -347,7 +348,8 @@ protected Configuration getConfiguration() {
         SQLListenerContextImpl context = startContext(connection(), queryMixin.getMetadata());
         String queryString = null;
         List<Object> constants = ImmutableList.of();
-
+        PreparedStatement stmt = null;
+        ResultSet rs = null;
         try {
```
jwgmeligmeyling, line 353: Please use `try-with-resources`.

jwgmeligmeyling: `try-with-resources` was not appropiate here, because in the happy flow the resources need to be closed lazily through the `SQLResultIterator`. The proper fix was to properly invoke `endContext` in the catch block for non-SQL runtime errors, as the context is responsible for freeing the resources in either scenario.

PR #2709, opened 2021-01-01
Get rid of javax-inject dependency in jpa and sql

jwgmeligmeyling:
I'd like to get rid of the dependency to `javax-inject` in QueryDSL code. This dependency might collide with other means of dependency injection in application (for example Spring Autowired) and it also may result in having javax-inject several times on the classpath because there are various forks of the API currently. Just from a framework point of view I also really don't like how we attempt to hook into the DI of the application.

So my suggestion is to replace all occurences of `Provider` with Java 8's `Supplier` instead. 

@johnktims @robertandrewbain @Shredder121 @idosal Would love to have some feedback on this. Do you think this change is likely to break a lot of applications? Or do you think these overloads are relatively rarely used and its safe to make this somewhat breaking change?


querydsl-examples/querydsl-example-sql-guice/pom.xml
```diff
@@ -92,8 +92,8 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
         <configuration>
-          <source>1.7</source>
-          <target>1.7</target>
+          <source>1.8</source>
```
robertandrewbain, line 95: Should this be 8, rather than 1.8, as it is below in querydsl-example-sql-spring/pom.xml?

robertandrewbain (changes requested):

querydsl-examples/querydsl-example-sql-guice/pom.xml
jwgmeligmeyling: Yes, thanks for pointing it out!

PR #2710, opened 2021-01-01
JPQL: Null instead of Tuple with null value is returned

### LINKED ISSUE #2048: JPQL: Null instead of Tuple with null value is returned
**Problem:** 
When I execute JPQL where only one column is selected and value on some of the rows is NULL in database, QueryDSL returns list of tuples where some of the items are NULL.
When I execute JPQL where two columns are selected and value on some of the rows are NULL for both columns in database, QueryDSL returns list where all items are instances of Tuple.

**Expected behaviour:**
Do not return null instead of Tuple having 1 null value.

**Environment:**
QueryDsl 3.6.x, 3.7.x
EclipseLink 2.6.2
jtds 1.3.1

**Code to fix:**
com.mysema.query.jpa.impl.AbstractJPAQuery::getResultList
instead of adding null to result when "o" is null, create instance with 1 null value.
rv.add(null); -> rv.add(projection.newInstance((new Object[] {null}));

**Logic behind:**
My observation is that javax.persistence.Query implementation (EclipseLink in my case) returns either list of type of single column (e.g. List<String>) if only one column is selected, otherwise, array of objects is returned (List<Object[]>). However, this should not be the case, when mapping to Tuple is expected.

**Current workaround:**
Select at least 2 columns.

ladislavlencucha:
Fix issue 2048 where single field selection which contains null value is not processed correctly (null instead of Tuple instance in 3.x and Tuple with null array in 4.x)

Fixes #2048

### DIFF
--- a/querydsl-collections/src/main/java/com/querydsl/collections/CollQuerySerializer.java
+++ b/querydsl-collections/src/main/java/com/querydsl/collections/CollQuerySerializer.java
@@ -210,8 +210,9 @@ public final class CollQuerySerializer extends SerializerBase<CollQuerySerialize
     public Void visit(FactoryExpression<?> expr, Void context) {
         visitConstant(expr);
         append(".newInstance(");
+        append("new Object[] {");
         handle(", ", expr.getArgs());
-        append(")");
+        append("})");
         return null;
     }
 
--- a/querydsl-collections/src/test/java/com/querydsl/collections/CollQueryStandardTest.java
+++ b/querydsl-collections/src/test/java/com/querydsl/collections/CollQueryStandardTest.java
@@ -13,17 +13,29 @@
  */
 package com.querydsl.collections;
 
-import static org.junit.Assert.*;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
 
 import java.util.Arrays;
 import java.util.Date;
 import java.util.List;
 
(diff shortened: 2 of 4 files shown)


jwgmeligmeyling:
The fix in `QTuple` is only partial. The underlying problem is that `QTuple` should never be invoked with `null`, but always with an `Object[]`.

In my commit I've fixed this for `CollQuerySerializer`.

However, the original issue was caused by the fact that for single valued tuples EclipseLink returns the plain value instead of an `Object[]`. Its normally the `EclipseLinkHandler` and specifically the `TransformingIterator` that should take care of this.

So we need to verify whether the issue still exists in Eclipselink and whether that issue is caused by an issue in `EclipseLinkHandler` or `TransformingIterator`, or that because of some reason the `EclipseLinkHandler` is not used while it should have.

Therefore the provided reproducer demonstrated a different issue than what would have caused #2048.

jwgmeligmeyling:
Because EclipseLink has no equivalent of result transformers (in Hibernate) the transformation for the factory expression is instead handled by `AbstractJPAQuery#getResultList`. The code there seemed faulty. I've fixed it so my reproducer for Eclipselink passes, I am now awaiting CI results to determine if I broke anything else.

jwgmeligmeyling:
I notice that now too 😉  Let's see what CI has to say about it.

ladislavlencucha:
@jwgmeligmeyling what do you think, ready for review?

jwgmeligmeyling:
Merged! Thanks for your contribution!

PR #2726, opened 2021-01-24
[#2612] Support Jakarta EE through Eclipse Transformer

### LINKED ISSUE #2612: Add support for Jakarta annotations + Jakarta persistence (in jakarta.* packages)

jwgmeligmeyling:
Allow Jakarta EE API's to be used through a transformed artefact. ln order to use the transformed modules, use the `jakarta` classifier for the dependencies to `querydsl-jpa`, `querydsl-apt` and `querydsl-jpa-codegen`.

- [x] Would be nice if we can somehow embed `src/apt/jpa/META-INF/services/javax.annotation.processing.Processor` in the produced jar...

Fixes #2612

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ A huge thanks goes out to all contributors that made this release possible in th
 * [#2106](https://github.com/querydsl/querydsl/issues/2106) - Support NullsLast ordering in `querydsl-collections`.
 * [#2404](https://github.com/querydsl/querydsl/issues/2404) - Upgrade of JTS / Geolatte in `querydsl-spatial`
 * [#2320](https://github.com/querydsl/querydsl/issues/2320) - Make Spatial support available to `HibernateDomainExporter` and `JPADomainExporter`. 
+* [#2612](https://github.com/querydsl/querydsl/issues/2612) - Support jakarta.* packages for new Jakarta EE releases (available through the`jakarta` classifiers for Maven)
 
 #### Bugfixes
 
--- a/querydsl-apt/pom.xml
+++ b/querydsl-apt/pom.xml
@@ -149,7 +149,25 @@
         <configuration>
           <useSystemClassLoader>false</useSystemClassLoader>
         </configuration>
-      </plugin>  
+      </plugin>
+
+      <plugin>
+        <groupId>org.eclipse.transformer</groupId>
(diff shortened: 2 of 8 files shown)


gitmotte:
Is the transformer solution a temporary solution? It does not compile/test against the new jakarta-API, am i right?

gitmotte:
Another downside of this solution is, that you need to add the classifier on each use of the library, it's not done with adding the classifier within dependencyManagement

jwgmeligmeyling:
We'll see where the ecosystem is in a couple of years. So far the adoption of these new APIs is lacking. I wouldn't be surprised to see JPA get superseded by a complete new API's before this name change flies off. In the mean time, backwards compatability is more important for us than forward compatability with tons of users still on Hibernate 5 and even Hibernate 4. And even Java EE application servers use Eclipse's Transformers, even on runtime to rely on the transformation.

gitmotte:
Yes, i know about the api-compatibility, the automatic transformation won't work otherwise ...

jwgmeligmeyling:
> Is the transformer solution a temporary solution? It does not compile/test against the new jakarta-API, am i right?

Created #2898 for that.

> Another downside of this solution is, that you need to add the classifier on each use of the library, it's not done with adding the classifier within dependencyManagement

We might switch to a separate artefact instead of a classifier as part of #2898. However, QueryDSL currently works best/easiest when classifiers are provided anyway.

PR #2727, opened 2021-01-24
Java 9 Automatic Module Names

jwgmeligmeyling:


michael-simons:
I wonder what your plan's are about releasing a patch with that change? It would be nice for other dependencies that want to provide valid `module-info.java`. 

(https://github.com/neo4j-contrib/cypher-dsl in that case).

Can you share information about your release plans?

jwgmeligmeyling:
I got a bit distracted with infra work at my current gig but I want to release somewhere by the end of this month.

michael-simons:
Great to hear, Jan-Willem. Thanks for your feedback.

jwgmeligmeyling:
Can you verify that the current `5.0.0-SNAPSHOT`  (master) has all the changes you need? I've only been able to do a limited amount of testing with JPMS.

michael-simons:
I'll ping you back. Let me see how far I come.

michael-simons:
So, here's what I did:

- Checkout 5.0.0 SNAPSHOT of QueryDSL, build and installed it locally
- Checkout my Cypher-DSL Snapshot, updated to 5.0.0 of yours, same procedure as above
- Created a demo (attached here)

-> Works

I am afraid however that a major upgrade won't make it into things like Spring Data (which where our common target lies).
As you the changes in this commit here only affects the automatic module names (which are very well chosen), I wonder if you see a chance creating a 4.4.x patch release containing those? That would be incredibly helpful. It wouldn't affect the overall runtime of your library at all.

If you need help cherry picking that commit or testing, I can offer it. Cheers and thanks already.

[mp.zip](https://github.com/querydsl/querydsl/files/6198022/mp.zip)

michael-simons:
2nd feedback: I back ported this commit here to the `QUERYDSL_4_4_0` and it builds just fine. The tests in the SQL module fails (but they do so without my changes, too).

So, if you want, I'd be more than happy to create a PR with a back port against a branch or tag that you could release as a patch release. 

Please let me know what you think.

michael-simons:
Aaah why not just fork and push it: https://github.com/michael-simons/querydsl/commits/QUERYDSL_4_4_1 

Such a patch would be really useful.

PR #2737, opened 2021-01-25
Type nullif and coalesce in expression hierarchy

### LINKED ISSUE #1376: Performing coalesce on NumberPath in from or where clause

jwgmeligmeyling:
Closes #1530
Fixes #1376

### DIFF
--- a/querydsl-core/src/main/java/com/querydsl/core/types/dsl/ComparableExpression.java
+++ b/querydsl-core/src/main/java/com/querydsl/core/types/dsl/ComparableExpression.java
@@ -409,5 +409,36 @@ public abstract class ComparableExpression<T extends Comparable> extends Compara
         return nullif(ConstantImpl.create(other));
     }
 
+    /**
+     * Create a {@code coalesce(this, exprs...)} expression
+     *
+     * @param exprs additional arguments
+     * @return coalesce
+     */
+    @Override
+    @SuppressWarnings({"unchecked", "rawtypes"})
+    public ComparableExpression<T> coalesce(Expression<?>... exprs) {
+        Coalesce<T> coalesce = new Coalesce<T>(getType(), mixin);
+        for (Expression expr : exprs) {
+            coalesce.add(expr);
+        }
+        return coalesce.getValue();
+    }
+
+    /**
+     * Create a {@code coalesce(this, args...)} expression
+     *
+     * @param args additional arguments
+     * @return coalesce
+     */
+    @Override
+    @SuppressWarnings({"unchecked"})
+    public ComparableExpression<T> coalesce(T... args) {
+        Coalesce<T> coalesce = new Coalesce<T>(getType(), mixin);
+        for (T arg : args) {
+            coalesce.add(arg);
+        }
+        return coalesce.getValue();
+    }
 
 }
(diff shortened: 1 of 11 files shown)


jwgmeligmeyling:
Kind of analog to #1530, see if we can incorporate these changes.

jwgmeligmeyling:
Hi @robertandrewbain , as you've reviewed #1530 before (and submitted #1376), could you have a look at these changes?

As you see I anticipate this as a breaking change. However, I think this change is for the better for the sake of consistency and clean API. Furthermore, with this major release we can finally introduce a few small breaking changes.

AFbenHassine (approved):
This is great, I'd be glad to be able to use comparable methods on coalesce and maintain the type.

querydsl-core/src/main/java/com/querydsl/core/types/dsl/ComparableExpression.java
```diff
@@ -387,4 +387,58 @@ public BooleanExpression loeAny(SubQueryExpression<? extends T> right) {
...
+    }
+
+    /**
+     * Create a {@code coalesce(this, exprs...)} expression
+     *
+     * @param exprs additional arguments
+     * @return coalesce
+     */
+    @Override
+    @SuppressWarnings({"unchecked", "rawtypes"})
```
F43nd1r, line 419: Is this necessary? What about `Expression<T>... exprs`?
That applies to the whole inheritance hierarchy.

jwgmeligmeyling:
@F43nd1r are you comfortable with this minor breaking change?

F43nd1r:
> @F43nd1r are you comfortable with this minor breaking change?

Yes, with demostrated use case breaking changes are acceptable, and there is no better time to break binary compat than a major release.

querydsl-core/src/main/java/com/querydsl/core/types/dsl/ComparableExpression.java
jwgmeligmeyling: Good suggestion, I'll check if it introduces other binary incompatibilities and otherwise apply the patch.

PR #2776, opened 2021-02-22
Fix the expression for week truncation for MySQL, Oracle and SQL Server

badgersow:
QueryDSL has problems with generating expressions which truncate the date to the start of the week.
This PR fixes them for:
* MySQL
* Oracle
* SQL Server

Please see internal comments for explanation of what's wrong with current implementation.

I've added unit tests for new statements, but I haven't found a way to add a proper integration test which checks things on real DBMS and actually truncates dates. Please point me in the right direction.

### DIFF
--- a/querydsl-sql/src/main/java/com/querydsl/sql/MySQLTemplates.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/MySQLTemplates.java
@@ -121,7 +121,7 @@ public class MySQLTemplates extends SQLTemplates {
 
         add(Ops.DateTimeOps.TRUNC_YEAR,   "str_to_date(concat(date_format({0},'%Y'),'-1-1'),'%Y-%m-%d')");
         add(Ops.DateTimeOps.TRUNC_MONTH,  "str_to_date(concat(date_format({0},'%Y-%m'),'-1'),'%Y-%m-%d')");
-        add(Ops.DateTimeOps.TRUNC_WEEK,   "str_to_date(concat(date_format({0},'%Y-%u'),'-2'),'%Y-%u-%w')");
+        add(Ops.DateTimeOps.TRUNC_WEEK,   "str_to_date(concat(date_format({0},'%Y-%u'),'-1'),'%Y-%u-%w')");
         add(Ops.DateTimeOps.TRUNC_DAY,    "str_to_date(date_format({0},'%Y-%m-%d'),'%Y-%m-%d')");
         add(Ops.DateTimeOps.TRUNC_HOUR,   "str_to_date(date_format({0},'%Y-%m-%d %k'),'%Y-%m-%d %k')");
         add(Ops.DateTimeOps.TRUNC_MINUTE, "str_to_date(date_format({0},'%Y-%m-%d %k:%i'),'%Y-%m-%d %k:%i')");
--- a/querydsl-sql/src/main/java/com/querydsl/sql/OracleTemplates.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/OracleTemplates.java
@@ -135,7 +135,7 @@ public class OracleTemplates extends SQLTemplates {
 
         add(Ops.DateTimeOps.TRUNC_YEAR, "trunc({0}, 'year')");
         add(Ops.DateTimeOps.TRUNC_MONTH, "trunc({0}, 'month')");
-        add(Ops.DateTimeOps.TRUNC_WEEK, "trunc({0}, 'w')");
+        add(Ops.DateTimeOps.TRUNC_WEEK, "trunc({0}, 'day')");
         add(Ops.DateTimeOps.TRUNC_DAY, "trunc({0}, 'dd')");
         add(Ops.DateTimeOps.TRUNC_HOUR, "trunc({0}, 'hh')");
         add(Ops.DateTimeOps.TRUNC_MINUTE, "trunc({0}, 'mi')");
--- a/querydsl-sql/src/main/java/com/querydsl/sql/OracleTemplates.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/OracleTemplates.java
@@ -135,7 +135,7 @@ public class OracleTemplates extends SQLTemplates {
 
         add(Ops.DateTimeOps.TRUNC_YEAR, "trunc({0}, 'year')");
         add(Ops.DateTimeOps.TRUNC_MONTH, "trunc({0}, 'month')");
-        add(Ops.DateTimeOps.TRUNC_WEEK, "trunc({0}, 'day')");
+        add(Ops.DateTimeOps.TRUNC_WEEK, "trunc({0}, 'iw')");
         add(Ops.DateTimeOps.TRUNC_DAY, "trunc({0}, 'dd')");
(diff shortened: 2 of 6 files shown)


querydsl-sql/src/main/java/com/querydsl/sql/MySQLTemplates.java
```diff
@@ -121,7 +121,7 @@ public MySQLTemplates(char escape, boolean quote) {
 
         add(Ops.DateTimeOps.TRUNC_YEAR,   "str_to_date(concat(date_format({0},'%Y'),'-1-1'),'%Y-%m-%d')");
         add(Ops.DateTimeOps.TRUNC_MONTH,  "str_to_date(concat(date_format({0},'%Y-%m'),'-1'),'%Y-%m-%d')");
-        add(Ops.DateTimeOps.TRUNC_WEEK,   "str_to_date(concat(date_format({0},'%Y-%u'),'-2'),'%Y-%u-%w')");
```
badgersow, line 124: This always used to truncate to Tuesday. I've changed it to Monday.
[Official documentation](https://dev.mysql.com/doc/refman/8.0/en/date-and-time-functions.html) confirms it (see the section on `%u`).

querydsl-sql/src/main/java/com/querydsl/sql/OracleTemplates.java
```diff
@@ -135,7 +135,7 @@ public OracleTemplates(char escape, boolean quote) {
 
         add(Ops.DateTimeOps.TRUNC_YEAR, "trunc({0}, 'year')");
         add(Ops.DateTimeOps.TRUNC_MONTH, "trunc({0}, 'month')");
-        add(Ops.DateTimeOps.TRUNC_WEEK, "trunc({0}, 'w')");
+        add(Ops.DateTimeOps.TRUNC_WEEK, "trunc({0}, 'day')");
```
badgersow, line 138: This looks like a typo. Please see [official Oracle docs](https://docs.oracle.com/cd/B19306_01/server.102/b14200/functions230.htm#i1002084)

querydsl-sql/src/main/java/com/querydsl/sql/SQLServerTemplates.java
```diff
@@ -151,7 +151,7 @@ protected SQLServerTemplates(Set<String> keywords, char escape, boolean quote) {
         // truncates timestamps by replacing suffix
         add(Ops.DateTimeOps.TRUNC_YEAR,   "CONVERT(DATETIME, CONVERT(VARCHAR(4), {0}, 120) + '-01-01')");
         add(Ops.DateTimeOps.TRUNC_MONTH,  "CONVERT(DATETIME, CONVERT(VARCHAR(7), {0}, 120) + '-01')");
-        // TODO week
+        add(Ops.DateTimeOps.TRUNC_WEEK, "cast(floor(cast(dateadd(dd, 2 - datepart(dw, {0}), {0}) as float)) as datetime");
```
badgersow, line 154: It doesn't follow existing pattern because it wasn't possible to implement it that way. New implementation makes use of internal `date` to `float` conversion, which is not the ideal thing, but is not likely to change. With proper integration test coverage it is better than no implementation at all (I hope).

querydsl-sql/src/main/java/com/querydsl/sql/OracleTemplates.java
jwgmeligmeyling: Seems you're right based on the documentation. Most databases use "w" or "week" here. What a poor choice on Oracle's side to use "day" here...

querydsl-sql/src/main/java/com/querydsl/sql/MySQLTemplates.java
jwgmeligmeyling: Is monday consistant with the start of ISO weeks? Or should that then be sunday?

I vagely remember that most databases follow ISO, where QueryDSL tends to follow the old Java Calendar API, which has a slightly different definition of weeks.

querydsl-sql/src/main/java/com/querydsl/sql/SQLServerTemplates.java
jwgmeligmeyling: It's better than nothing 🤷🏻 

At [Blaze-Persistence](https://github.com/Blazebit/blaze-persistence/blob/master/core/impl/src/main/java/com/blazebit/persistence/impl/function/trunc/week/MSSQLTruncWeekFunction.java) we use `DATEADD(WEEK, DATEDIFF(WEEK, 0, ?1), 0)` from https://stackoverflow.com/a/32955740/2104280 for SQL Server. Perhaps a bit cleaner 😄

querydsl-sql/src/main/java/com/querydsl/sql/MySQLTemplates.java
badgersow: According to the docs, it's always going to be Monday (regardless of DB config), which is following ISO standard. I've tested this on several dates and different start of the week setups. 
From what I see, DB are not consistent on following the ISO and some of them even have multiple functions which treat the config differently. The question we should ask is how QueryDSL should attempt to behave?
In my opinion (and my particular use case) it's best to follow ISO regardless of locale. It makes the result predictable and the same across DB. Having solid truncation to Monday is better than inconsisted behaviour based on DB quirks.
I'm open for suggestions, but this is what I'm following currently in this PR (ISO behaviour).

querydsl-sql/src/main/java/com/querydsl/sql/OracleTemplates.java
badgersow: I've changed this to 'iw', which always truncates to ISO week start (Monday). Oracle docs are extremely confusing, but my testing shows that 'iw' always truncates to Monday regardless of the territory config (unlike 'day')

querydsl-sql/src/main/java/com/querydsl/sql/SQLServerTemplates.java
badgersow: Thank you for good suggestion. I've played with it and it seems not to work on some inputs

```
set datefirst 1;
select DATEADD(WEEK, DATEDIFF(WEEK, 0, cast('2021-02-21' as datetime)), 0);
==> 2021-02-22 00:00:00.000
```

I've adopted a similar approach which seems to work correctly and always truncates to Monday.

```
select DATEADD(WEEK, DATEDIFF(WEEK, 0, {0} - 1), 0);
```

The only difference - we subtract one day from the input. There is an [article](http://www.silota.com/docs/recipes/sql-server-date-parts-truncation.html) describing why this works. It not trivial, but it worked on my tests (on different `datefirst` values it always truncates to Monday).

badgersow:
Hey @jwgmeligmeyling . Thanks a lot for rewiew. I've changed the code to always follow ISO and replied to your comments.

badgersow:
@jwgmeligmeyling Can you please review this PR once again or if you are busy, maybe you can advice who else I can ask for review? Thanks in advance for your time.

jwgmeligmeyling:
Yes my plans are to review this in the weekend and incorperate this in the 5.0 release!

Issue #2507, opened 2019-12-20
TRUE/FALSE BooleanExpressions can't be used in where clause for oracle directly


7h3kk1d:
The following code typechecks and compiles but does not provide valid Oracle SQL since `1` is not a valid boolean expression in oracle. This becomes more of a problem when you're trying to fold together `BooleanExpression`s using `BooleanExpression::and` and `BooleanExpression::or`

```java
new SQLQuery<>(new OracleTemplates())
            .select(Expressions.constant(1))
            .where(TRUE); // Produces select ? from dual where 1
```

jwgmeligmeyling:
This is not specific to Oracle, and not really a bug.

QueryDSL is a DSL that allows you to write SQL queries with code. If you use it to generate an invalid SQL query, the query will be invalid.

`select 1 from dual where 1 or 1` is apparently invalid in Oracle, so `query.select(Expressions.ONE).where(Expressions.FALSE.or(Expressions.TRUE))` is invalid as well.

QueryDSL should not go through a round of hoops to transform the QueryDSL under the hood to compensate for this. Then we open the door for thousands of more of these issues to be opened, suggesting the same for PostgreSQL, SQL Server, MySQL, JPQL etc. And then users that are confused their query gets unintentionally rewritten leading to a poorer QEP.

So I am sorry, although I do understand why its for some dialects that `Expression.FALSE` is a `Predicate` instead of just an `Expression<Boolean>` and therefore is a valid parameter for `.where(...)`, I am not particularly enthusiastic about this change. We want QueryDSL to remain transparent in query serialization.

PR #2786, opened 2021-03-04
#2507 allow using boolean template with Oracle

IlyaNerd:


IlyaNerd:
yeah, i dont see a way to fix it for Oracle, as need additional context - if TRUE is used as predicate or as expression. selecting true in oracle doesnt make sense from user standpoint, but god knows what can be found in the wild :)

PR #2796, opened 2021-04-07
created bill of materials for querydsl

lpandzic:


jwgmeligmeyling (approved):

jwgmeligmeyling:
Change looks good.

I am wondering though: what problem does this solve?

I am familiar with bill of material POMs for other projects. However with QueryDSL I usually only need to declare one or two dependencies and the rest comes transitively.

I think it would be nice if we create some documentation for this change describing the preferred way to manage QueryDSL dependencies.

lpandzic:
It's mostly for people on the "bleeding edge".
Example situation:
1. project uses spring and spring-boot-dependencies bom
2. querydsl releases 5.0.0
3. user wants to use querydsl 5.0.0 and is forced to redefine every querydsl dependency in his dependency management due to the fact that spring boot bom has old dependencies.

I had this exact situation with jackson and it really helps that you only need to redefine jackson bom in front of spring boot bom for it to override the spring boot one.

I agree that this should be documented.

jwgmeligmeyling:
Thanks! 👍🏻

PR #2805, opened 2021-04-10
Backport of the automatic modules names.

michael-simons:
Hi. This is a back port of d651911bd592526e34f70a50778954810e88f11b into the 4.x branch. It would be really beneficial to release a 4.4.1 patch release so that QueryDSL can be used in modular world without relying on file based generated modules names. This doesn't change anything when run on the class path and doesn't add any further Java modules system capabilities to QueryDSL.

Thank you for considering this.

### DIFF
--- a/querydsl-apt/pom.xml
+++ b/querydsl-apt/pom.xml
@@ -134,6 +134,19 @@
         </configuration>
       </plugin>
        -->
+
+      <plugin>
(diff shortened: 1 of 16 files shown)


jwgmeligmeyling:
Querydsl 5.0.0.M1 is out and is the first version of Querydsl to support Java 9 out-of-the-box. We do not intend any further 4.x releases.

PR #2794, opened 2021-04-05
Kotlin codegen support

F43nd1r:
Fixes #1828 
Supersedes #2730

PR #2815, opened 2021-05-02
Java 11 compile

F43nd1r:
Started work on #2779 as noted in https://github.com/querydsl/querydsl/pull/2794#issuecomment-830772378

Main points:
- Replace org.reflections:reflections with io.github.classgraph:classgraph, because the old library isn't updated anymore.
- Tests were expecting the old `javax.annotation.Generated`, generify.
- Update several dependency versions, in case of ECJ that includes a group change.


F43nd1r:
I'm trying to fix more of the tests, but I'm unsure on how to run them locally. The vagrant setup mentioned in the readme doesn't seem to work anymore. Is there a method/setup I can use?

F43nd1r:
@jwgmeligmeyling I'm no scala expert. Accessing `scala.Enumeration$Value` in generated sources works with jdk 8, but not with 11.

```
Running com.querydsl.scala.GenericExporterTest
target/gen1/com/querydsl/scala/ext/QSomeEntity.scala:24: error: type Enumeration$Value is not a member of package scala
  val someScalaEnum = createComparable[scala.Enumeration$Value]("someScalaEnum")
                                             ^
Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 6.777 sec <<< FAILURE! - in com.querydsl.scala.GenericExporterTest
Export(com.querydsl.scala.GenericExporterTest)  Time elapsed: 6.776 sec  <<< FAILURE!
java.lang.AssertionError: Compilation failed
	at com.querydsl.scala.CompileTestUtils$.assertCompileSuccess(CompileTestUtils.scala:36)
	at com.querydsl.scala.CompileTestUtils$.assertCompileSuccess(CompileTestUtils.scala:27)
	at com.querydsl.scala.GenericExporterTest.Export(GenericExporterTest.scala:20)
```

lpandzic:
The version defined in [scala module pom.xml](https://github.com/querydsl/querydsl/blob/master/querydsl-scala/pom.xml#L22) [doesn't support jdk 11](https://docs.scala-lang.org/overviews/jdk-compatibility/overview.html).

lpandzic:
Try upgrading to 2.11.2

F43nd1r:
@lpandzic this PR used `2.12.12` even before I started working on it. I'll try the downgrade anyway.

F43nd1r:
Works for now. 
Long term this dependency should be upgraded to a version not 5+ years old, but I guess that's valid for most dependencies in here.

lpandzic:
> 
> 
> Works for now.
> Long term this dependency should be upgraded to a version not 5+ years old, but I guess that's valid for most dependencies in here.

I agree, but the focus now should be getting 5.0.0 released.

querydsl-codegen-utils/src/main/java/com/querydsl/codegen/utils/SimpleCompiler.java
```diff
@@ -47,36 +47,8 @@ protected static boolean isSureFireBooter(URLClassLoader cl) {
...
-                }
-            }
-            return paths.stream().collect(Collectors.joining(File.pathSeparator));
-        } catch (UnsupportedEncodingException e) {
-            throw new CodegenException(e);
-        } catch (IOException e) {
-            throw new CodegenException(e);
-        }
+    public static String getClassPath(ClassLoader cl) {
+        return new ClassGraph().overrideClassLoaders(cl).getClasspath();
```
jwgmeligmeyling, line 51: A key design decision in QueryDSL 5 was keeping the transitive dependencies to a bare minimum. Some folks in the issue tracker mentioned even that frameworks should impose no further dependencies at all.

I know this dependency is just used in code gen and won't leak into the users application in _most_ cases. That is, unless the wrong dependency scope is used for the apt dependencies or a module with a runtime dependency on codegen is used (like `Alias.*`).

Was it necessary to update this code for Java 11?

jwgmeligmeyling (approved):

querydsl-codegen-utils/src/main/java/com/querydsl/codegen/utils/SimpleCompiler.java
F43nd1r: Classloaders aren't `URLClassLoader`s anymore. 

Alternative would be to copy/paste from that library (or any other source), but updating is necessary.

jwgmeligmeyling: Oh well, the code relied on a dependency anyway, that we now got rid of 😄

querydsl-core/pom.xml
```diff
@@ -98,4 +98,21 @@
       </plugin>
     </plugins>
```
jwgmeligmeyling, line 99: Can you verify that the introduced dependency for `classgraph` is in `provided` scope here? (as it is brought transitively with the dependency on codegen)

src/main/resources/querydsl_checks_base.xml
```diff
@@ -72,7 +69,7 @@
     <module name="WhitespaceAround"/>
 
     <module name="ModifierOrder"/>
-    <module name="RedundantModifier"/>
+<!--    <module name="RedundantModifier"/>-->
```
jwgmeligmeyling, line 73: Ah yes, I see I still had to reapply a few rules here. Somehow they started giving issues after the update.

Should we first merge this and then update the other PR?

querydsl-core/pom.xml
F43nd1r: Did you intend to comment somewhere else? Core doesn't depend on codegen

src/main/resources/querydsl_checks_base.xml
F43nd1r: Yes, this first, the other will be simpler to resolve

querydsl-core/pom.xml
F43nd1r: classgraph scope is `test` for all modules except codegen and codegen-utils

jwgmeligmeyling:
I noticed that coverage report on Java 11 gives issues on the master branch: https://github.com/querydsl/querydsl/runs/2729469140 . It is not enabled on the other branches, hence the issue.

We might want to look into it some day.

PR #2883, opened 2021-06-03
Fix various issues caused by invalid parameter assumptions

### LINKED ISSUE #1413: Incorrect parameter values with Hibernate custom types.

### LINKED ISSUE #1429: Reusing of constants in JPQL generation causes issues with hibernate query caching

### LINKED ISSUE #2326: Use of Hibernate 5.3 and HibernateQuery results in "Could not locate named parameter"
The use of querydsl-jpa's HibernateQuery with Hibernate 5.3 will thrown an exception when you use parameters:
```
java.lang.IllegalArgumentException: Could not locate named parameter [1], expecting one of []

	at org.hibernate.query.internal.ParameterMetadataImpl.getNamedParameterDescriptor(ParameterMetadataImpl.java:218)
	at org.hibernate.query.internal.ParameterMetadataImpl.getQueryParameter(ParameterMetadataImpl.java:187)
	at org.hibernate.query.internal.QueryParameterBindingsImpl.getBinding(QueryParameterBindingsImpl.java:188)
	at org.hibernate.query.internal.AbstractProducedQuery.setParameter(AbstractProducedQuery.java:494)
	at org.hibernate.query.internal.AbstractProducedQuery.setParameter(AbstractProducedQuery.java:107)
	at com.querydsl.jpa.hibernate.HibernateUtil.setValue(HibernateUtil.java:91)
	at com.querydsl.jpa.hibernate.HibernateUtil.setConstants(HibernateUtil.java:79)
	at com.querydsl.jpa.hibernate.AbstractHibernateQuery.createQuery(AbstractHibernateQuery.java:105)
	at com.querydsl.jpa.hibernate.AbstractHibernateQuery.createQuery(AbstractHibernateQuery.java:97)
	at com.querydsl.jpa.hibernate.AbstractHibernateQuery.fetch(AbstractHibernateQuery.java:174)
	at com.querydsl.example.jpa.repository.TweetRepository.findAllWithHibernateQuery(TweetRepository.java:30)
	at com.google.inject.persist.jpa.JpaLocalTxnInterceptor.invoke(JpaLocalTxnInterceptor.java:66)
	at com.querydsl.example.jpa.repository.TweetRepositoryTest.find_list_by_predicate(TweetRepositoryTest.java:44)
```

I am not the only person experiencing this error, as @jpimag experienced the same issue:
https://github.com/querydsl/querydsl/pull/2283#issuecomment-399247423

To reproduce this issue, I modified the querydsl-example-jpa-guice example contained within the QueryDSL repository. You can find the patch here:
https://gist.github.com/andygoossens/5eed865ccfe84857f7e99ddb2dfb5790
Then just run test class `com.querydsl.example.jpa.repository.TweetRepositoryTest` and check whether it fails.

You can easily compare the different behaviour between Hibernate versions by altering the `<hibernate.version>` parameter in the POM file (part of the "hibernate5" profile). e.g. Hibernate version 5.2.17.Final works fine, but version 5.3.1.Final fails. Don't forget to activate the necessary Maven profile(s) of course! :-)

It looks like there is a mixup between named query parameters and positional query parameters. `getConstantToLabel()` could contain a constant with index "1" (as a String and meant to be used as `?1`) but QueryDSL is calling `Query<R> setParameter(String name, Object val);` meant for named parameters (like `:example`). Hibernate does not find the named parameter "1" and fails.

### LINKED ISSUE #2816: Generated JPA query with incorrect argument binding indexes
## Observed vs. expected behavior

Given an `JPA` entity with two boolean fields, one of them annotated with a custom type `@Type`, when querying by those two fields providing the same boolean value (`true, true` or `false, false`) the generated query ends up using `?1` for both arguments (as they are considered a constant).

The generated query is
```sql
select user
from User user
where user.active = ?1 and user.admin = ?1  -- note the two arguments as ?1
```
when the expected would be
```sql
select user
from User user
where user.active = ?1 and user.admin = ?2
```

## Steps to reproduce

Here is a sample application to reproduce the problem, with some tests and loggers enabled to show the differences between the generated query and the expected one.

https://github.com/didiez/hibernate-querydsl-type-binding-bug

## Environment

Querydsl version: **`4.4.0`**

Querydsl module: **`querydsl-jpa`**

Database: tested against `oracle 11g` and `h2` (probably database agnostic)

JDK: 11

## Additional details

Might be related to the changes applied to fix #2326 (PR #2354)

jwgmeligmeyling:
Reverts #2354
Fixes #2326
Fixes #2816
Fixes #1413
Fixes #1429
Closes #2000
Affects https://github.com/Blazebit/blaze-persistence/pull/1071

### DIFF
--- a/querydsl-core/src/main/java/com/querydsl/core/support/SerializerBase.java
+++ b/querydsl-core/src/main/java/com/querydsl/core/support/SerializerBase.java
@@ -13,19 +13,31 @@
  */
 package com.querydsl.core.support;
 
+import com.querydsl.core.JoinFlag;
+import com.querydsl.core.QueryFlag;
+import com.querydsl.core.types.Constant;
+import com.querydsl.core.types.Expression;
+import com.querydsl.core.types.FactoryExpression;
+import com.querydsl.core.types.Operation;
+import com.querydsl.core.types.Operator;
+import com.querydsl.core.types.Ops;
+import com.querydsl.core.types.ParamExpression;
+import com.querydsl.core.types.Path;
+import com.querydsl.core.types.PathType;
+import com.querydsl.core.types.Template;
+import com.querydsl.core.types.TemplateExpression;
+import com.querydsl.core.types.Templates;
+import com.querydsl.core.types.Visitor;
+import org.jetbrains.annotations.NotNull;
+
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.EnumSet;
-import java.util.HashMap;
+import java.util.IdentityHashMap;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-
-import com.querydsl.core.JoinFlag;
-import com.querydsl.core.QueryFlag;
-import com.querydsl.core.types.*;
-
-import static java.util.Collections.unmodifiableMap;
 
 /**
  * {@code SerializerBase} is a stub for Serializer implementations which serialize query metadata to Strings
@@ -46,8 +58,9 @@ public abstract class SerializerBase<S extends SerializerBase<S>> implements Vis
 
     private String anonParamPrefix = "_";
 
-    private Map<Object, String> constantToNamedLabel;
-    private Map<Object, Integer> constantToNumberedLabel;
+    protected final List<Object> constants = new LinkedList<>();
+
+    protected final Map<Object, String> constantToLabel = new IdentityHashMap<>();
 
     @SuppressWarnings("unchecked")
     private final S self = (S) this;
@@ -79,35 +92,8 @@ public abstract class SerializerBase<S extends SerializerBase<S>> implements Vis
         return constantPrefix;
     }
 
-    @Deprecated
     public Map<Object, String> getConstantToLabel() {
-        return getConstantToAllLabels();
-    }
-
-    public Map<Object, String> getConstantToNamedLabel() {
-        if (constantToNamedLabel == null) {
-            constantToNamedLabel = new HashMap<Object, String>(4);
-        }
-        return constantToNamedLabel;
-    }
-
-    public Map<Object, Integer> getConstantToNumberedLabel() {
-        if (constantToNumberedLabel == null) {
-            constantToNumberedLabel = new HashMap<Object, Integer>(4);
-        }
-        return constantToNumberedLabel;
-    }
-
-    public Map<Object, String> getConstantToAllLabels() {
-        Map<Object, String> namedLabels = getConstantToNamedLabel();
-        Map<Object, Integer> numberedLabels = getConstantToNumberedLabel();
-
-        Map<Object, String> allLabels = new HashMap<Object, String>(namedLabels);
-        for (Map.Entry<Object, Integer> entry : numberedLabels.entrySet()) {
-            allLabels.put(entry.getKey(), entry.getValue().toString());
-        }
-
-        return unmodifiableMap(allLabels);
+        return constantToLabel;
     }
 
     protected int getLength() {
@@ -223,13 +209,37 @@ public abstract class SerializerBase<S extends SerializerBase<S>> implements Vis
     }
 
     public void visitConstant(Object constant) {
-        if (!getConstantToAllLabels().containsKey(constant)) {
-            final String constLabel = constantPrefix + (getConstantToNamedLabel().size() + 1);
-            getConstantToNamedLabel().put(constant, constLabel);
-            append(constLabel);
-        } else {
-            append(getConstantToNamedLabel().get(constant));
-        }
+        final String constantLabel = getConstantToLabel().computeIfAbsent(constant, this::getConstantLabel);
+        constants.add(constant);
+        serializeConstant(constants.size(), constantLabel);
+    }
+
+    /**
+     * Serialize the constant as parameter to the query. The default implementation writes the
+     * label name for the constants. Some dialects may replace this by indexed based or
+     * positional parameterization.
+     * Dialects may also use this to prefix the parameter with for example ":" or "?".
+     *
+     * @param parameterIndex index at which this constant occurs in {@link #getConstants()}
+     * @param constantLabel label under which this constant occurs in {@link #getConstantToLabel()}
+     */
+    protected void serializeConstant(int parameterIndex, String constantLabel) {
+        append(constantLabel);
+    }
+
+    /**
+     * Generate a constant value under which to register a new constant in {@link #getConstantToLabel()}.
+     *
+     * @param value the constant value or parameter to create a constant for
+     * @return the generated label
+     */
+    @NotNull
+    protected String getConstantLabel(Object value) {
+        return constantPrefix + (getConstantToLabel().size() + 1);
+    }
+
+    public List<Object> getConstants() {
+        return constants;
     }
 
     @Override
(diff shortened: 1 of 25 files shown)


F43nd1r:
looks good, if everything runs through. But why are tests running four times on this now?

jwgmeligmeyling:
I think because I also push to the querydsl remote so I get both the push and pr builds, for both Java 8 and 11.

Its unrelated to the change.

jwgmeligmeyling:
@F43nd1r it has proven surprisingly difficult to implement this properly.

HQL historically had two types of parameters: legacy positional parameters (`?`) and named parameters (`:test`). JPA currently describes to types of parameters implementors need to support as well: indexed parameters (`?1`, `?2`) and named parameters (`:test`). Hibernate recently dropped support for legacy positional parameters in favour of indexed parameters. Indexed parameters and named parameters cannot be mixed within the same query.

QueryDSL has two types of parameters: named parameters and anonymous parameters. While constants are usually serialized as parameters as well (unless literal rendering is enabled, which is currently only supported for querydsl-sql).

All constants for which a parameter is generated are stored with a unique key value in `constantToLabel`. While you would expect this label to be used for name of a named parameter, this is not the case. In JPA we always use indexed parameters. And the same applies to `querydsl-sql`, where  `constantToLabel` is unused and a list `constants` is maintained instead. In both cases, even when named parameters are explicitly provided in the query (through `new Param(Object.class, "name")`), this name is lost in the final query. While it would be nice to copy these explicitly provided parameter names to the final query, this would require us to put generated parameter names for the indexed parameter names in the query as indexed and named parameters usually cannot be mixed.

The entire setup in `constantToLabel` is solely used for two purposes:
- "attribute" names for the `CollQuerySerializer`. `CollQuerySerializer` executes queries in-memory and therefore compiles some Java code and thus needs "variables" to store the constants in.
- parameter names for `JDOQLSerializer`, as JDO does only seem to support named parameters and not indexed parameter, and thus a name needs to be generated if one is missing.

IMO it does not make sense to deal with `constantToLabel` in `SerializerBase` if its only applicable to two very specific dialects and only causes trouble for the other dialects.

I am currently trying to set up a new structure for `SerializerBase` that is both more easily extendable but also doesn't break any of the current dialects.

querydsl-core/src/main/java/com/querydsl/core/support/SerializerBase.java
```diff
@@ -46,8 +58,9 @@
 
     private String anonParamPrefix = "_";
 
-    private Map<Object, String> constantToNamedLabel;
-    private Map<Object, Integer> constantToNumberedLabel;
+    protected final List<Object> constants = new LinkedList<>();
```
F43nd1r, line 61: Is there a need for this? Seems like the only difference to `constantToLabel` keys is order, is that relevant?
Seems like it should be possible to use a single data structure instead of keeping two in sync.

F43nd1r (commented):
Are all the fixed issues covered with old tests already? If not we could add some to prevent future regressions

querydsl-core/src/main/java/com/querydsl/core/support/SerializerBase.java
jwgmeligmeyling: Unfortunately it is not, and that is because `constantToLabel` works very differently across the various serializers for different dialects.

Constant to label is a mapping used primarily by JDO and CollQuerySerializer to create a unique label for unique objects and store them as a variable in the context of the query (so this is in fact not really a constant or a parameter at all).

As query parameters objects may occur different times using various mappings. Therefore the SQL serializer already maintained a list of constants instead of relying on constantToLabel. The proper fix was to use this for JPQL as well.

I did try to combine the two datastructures and it is possible as long as all labels are unique and the mapping is inverted to labelToConstant. However, this approach makes the implementation in SQL and JPQL overly complex and error prone (as this issues already show) and it also  defeats optimizations  that were deliberately made for JDO and CollQuerySerializer. In the end I just stumbled upon lots of overriden methods and serializer hacks to get all tests passing in all modules.

I am also worried that changing `constantToLabel` to `labelToConstant` is going to be a massive pain in the ass for integrators that implement different serializers based on this API or even use `getLabelToConstant()` for logging / parameter binding purposes  in their own applications.

This approach is simpler, more generic and only changes recently updated API's while leaving the API's that have been there for years in tact.

jwgmeligmeyling:
* #2326 is tested simply by testing the examples / tests against Hibernate 5.3.
* #2816 is caused by #1429
* #1413 is caused by #1429
* #1429 is  tested by the changes in the asserted queries in the JPA integration

So no new tests are required beyond the altered tests in this PR.

F43nd1r (approved):

PR #2884, opened 2021-06-03
Safe the environment, skip duplicate test runs

F43nd1r:
Related https://github.com/querydsl/querydsl/pull/2883#issuecomment-854234484

PR #2887, opened 2021-06-04
Fix ci precondition

F43nd1r:
Copy & paste error from action docu introduced in #2884

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
 
   build:
     needs: check_duplicate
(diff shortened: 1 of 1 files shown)


jwgmeligmeyling:
Screenshot 2021-06-04 at 14 03 07


Now there's no build at all 😅

F43nd1r:
Yeah, seems like the actions we're using is bugged. Both runs claim to be newer
```
Skip execution because the exact same files are concurrently checked in older https://github.com/querydsl/querydsl/actions/runs/906163660
```
```
Skip execution because the exact same files are concurrently checked in older https://github.com/querydsl/querydsl/actions/runs/906163770
```

PR #2890, opened 2021-06-04
Use patched skip duplicate actions action

F43nd1r:
I've forked and fixed the action, use my fork until https://github.com/fkirc/skip-duplicate-actions/pull/116 is merged.


F43nd1r:
Fast upstream response, is now merged.

jwgmeligmeyling:
`same_content_newer` seems to only check file paths, not contents. So if you amend a commit without adding new changes to a new file, the build will not execute. Perhaps it should check for commit SHA as well?

F43nd1r:
@jwgmeligmeyling do you have an example run where this happened?

PR #2895, opened 2021-06-07
Add one-arg overloads for coalesce

F43nd1r:
Related discussion: https://querydsl.slack.com/archives/C06KLG1BP/p1623048740053800?thread_ts=1622940119.038500&cid=C06KLG1BP

This has two advantages:
 - no "generic array creation" warnings for users with only one arg
 - no implicit array creation in this case, improving performance

### DIFF
--- a/querydsl-core/src/main/java/com/querydsl/core/types/dsl/BooleanExpression.java
+++ b/querydsl-core/src/main/java/com/querydsl/core/types/dsl/BooleanExpression.java
@@ -178,17 +178,44 @@ public abstract class BooleanExpression extends LiteralExpression<Boolean> imple
     }
 
     /**
+     * Create a {@code coalesce(this, expr)} expression
+     *
+     * @param expr additional argument
+     * @return coalesce
+     */
+    @Override
+    public BooleanExpression coalesce(Expression<Boolean> expr) {
+        Coalesce<Boolean> coalesce = new Coalesce<Boolean>(getType(), mixin);
+        coalesce.add(expr);
+        return coalesce.asBoolean();
+    }
+
+    /**
      * Create a {@code coalesce(this, exprs...)} expression
      *
      * @param exprs additional arguments
      * @return coalesce
      */
     @Override
+    @SuppressWarnings("unchecked")
     public BooleanExpression coalesce(Expression<Boolean>... exprs) {
         Coalesce<Boolean> coalesce = new Coalesce<Boolean>(getType(), mixin);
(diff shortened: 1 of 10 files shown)


querydsl-core/src/main/java/com/querydsl/core/types/dsl/BooleanExpression.java
```diff
@@ -177,13 +177,27 @@ public BooleanExpression nullif(Boolean other) {
...
+    }
+
     /**
      * Create a {@code coalesce(this, exprs...)} expression
      *
      * @param exprs additional arguments
      * @return coalesce
      */
     @Override
+    @SuppressWarnings("unchecked")
```
jwgmeligmeyling, line 200: Does this eliminate "SafeVarargs" warnings?

jwgmeligmeyling (approved):

querydsl-core/src/main/java/com/querydsl/core/types/dsl/BooleanExpression.java
F43nd1r: In our codebase yes, not on the user side. To completely get rid of all warnings on user side we'd have to remove `T`

jwgmeligmeyling: I don't think removing <T> will solve it. The problem is that the method cannot be final right?

Anyways, the other two overloads will probably be the most frequently used ones.

F43nd1r: `Expression<?>...` would remove the warning on that method, but allows wrong parameters to be passed in.

jwgmeligmeyling: I'm not really seeing why `Expression<?>...` would prevent heap pollution because the array on the heap can still be exposed as return variable, but I am willing to sacrifice a bit of input type safety in exhange for not having to bother explaining our users that our methods are in fact SafeVarargs but cant be annotated as such ;-)

F43nd1r: I'm not saying it makes sense, it just works like this. Reverted the type adding commit on this branch.

jwgmeligmeyling: Seems its been to long since I've read Effective Java to understand SafeVarargs properly 😉

PR #2911, opened 2021-06-16
Dokka for Kotlin modules

F43nd1r:
Dokka is the Kotlin equivalent of javadoc. It generates `*-javadoc.jar`s for maven central.

Also moved the kotlin version definition up to root, differing versions in the two modules don't make too much sense.


jwgmeligmeyling:
@johnktims with this patch we should be able to do the release

jwgmeligmeyling:
@F43nd1r Thanks very much for looking into this!

jwgmeligmeyling (approved):

PR #2916, opened 2021-06-17
Change visibility of package private methods in AbstractMongodbQuery

### LINKED ISSUE #2915: Expose `QueryMixin` to subclasses of `AbstractMongodbQuery`
## Why the new feature should be added

To allow customized construction of sort and accessing other metadata, it would make sense to increase the visibility of `getQueryMixin()` to `protected` in `AbstractMongodbQuery`.

jwgmeligmeyling:
Fixes #2915

@mp911de please have a look 😄


mp911de (approved):
Wow, that was quick. Thanks a lot, looks good.

Issue #2921, opened 2021-06-21
Dependency management clash with Spring boot


tstavinoha:
## Observed vs. expected behavior

When using Querydsl BOM to manage dependencies in Maven projects, if Querydsl BOM is listed before Spring Boot dependencies, it causes the scope modifier `provided` to be added to dependency `jakarta.persistence:jakarta.persistence-api`. This causes a usability issue, because it makes the users have to manually list a dependency which would ideally be completely transient and pulled into the project by Spring Boot starter. We need to have Querydsl in front of Spring Boot BOM in order to use newer version of Querydsl.

Would you consider removing the `provided` modifier from the parent project?
It was recently [adjusted](https://github.com/querydsl/querydsl/commit/356be9d752b4da64bce41875755c738925f0c83e#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8)

## Steps to reproduce

Create a minimal project setup, add a dependency to spring-boot-starter-data-jpa and dependency management in order that querydsl bom precedes spring boot BOM.

```
    <dependencyManagement>
        <dependencies>

            <dependency>
                <groupId>com.querydsl</groupId>
                <artifactId>querydsl-bom</artifactId>
                <version>5.0.0.M1</version>
                <type>pom</type>
                <scope>import</scope>
            </dependency>

            <!-- SPRING BOM -->
            <dependency>
                <groupId>org.springframework.boot</groupId>
                <artifactId>spring-boot-dependencies</artifactId>
                <version>2.5.1</version>
                <type>pom</type>
                <scope>import</scope>
            </dependency>

        </dependencies>
    </dependencyManagement>
    <dependencies>

        <dependency>
            <groupId>org.springframework.boot</groupId>
            <artifactId>spring-boot-starter-data-jpa</artifactId>
        </dependency>

    </dependencies>
```

Output of `dependency:tree` should be something like:
```
[INFO] --- maven-dependency-plugin:2.8:tree (default-cli) @ demo ---
[INFO] com.example:demo:jar:0.0.1-SNAPSHOT
[INFO] \- org.springframework.boot:spring-boot-starter-data-jpa:jar:2.5.1:compile
[INFO]    \- jakarta.persistence:jakarta.persistence-api:jar:2.2.3:provided (scope managed from compile)
```

## Environment

Querydsl version: 5.0.0.M1

Querydsl module: general

Database: n/a

JDK: 16

## Additional details
[...]

PR #2943, opened 2021-06-23
added flatten plugin to remove parent from generated pom.xml in bom

lpandzic:
Flattening bom in order to prevent issues like https://github.com/querydsl/querydsl/issues/2921 or that is - to have a clear and simple bom.

### DIFF
--- a/querydsl-bom/pom.xml
+++ b/querydsl-bom/pom.xml
@@ -54,6 +54,11 @@
             </dependency>
             <dependency>
                 <groupId>${project.groupId}</groupId>
+                <artifactId>querydsl-guava</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>${project.groupId}</groupId>
                 <artifactId>querydsl-sql</artifactId>
                 <version>${project.version}</version>
             </dependency>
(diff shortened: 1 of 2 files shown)


lpandzic:
@jwgmeligmeyling can you review please?

querydsl-bom/pom.xml
```diff
@@ -16,6 +16,10 @@
     <url>${project.homepage}</url>
     <packaging>pom</packaging>
 
+    <properties>
+        <flatten-maven-plugin.version>1.2.7</flatten-maven-plugin.version>
```
F43nd1r, line 20: This property shows up in the generated pom. Is that necessary?

F43nd1r (changes requested):
Please add the generated file to gitignore.

Also some modules are missing from bom (at least `guava` and `kotlin-codegen`), you can either address that here or I'll open another PR.

lpandzic:
> Please add the generated file to gitignore.
> 
> Also some modules are missing from bom (at least `guava` and `kotlin-codegen`), you can either address that here or I'll open another PR.

Added flattened-pom to gitignore.
Added missing modules.

querydsl-bom/pom.xml
lpandzic: Removed it from flattened pom.

F43nd1r (approved):

jwgmeligmeyling (approved):

lpandzic:
I don't have write access permission, can I get one or can someone merge the PR?

jwgmeligmeyling:
Thanks for looking into this! 👍🏻

Issue #2964, opened 2021-07-12
Collections: Failure of setting up classpath for javac in case of running in Spring Boot on Ubuntu and MacOS


viktorzub:
Minimal reproduction code: 
https://github.com/viktorzub/spring-boot-jar-querydsl-bug/tree/main/spring-boot-querydsl-bug

## Steps to reproduce

When run as Spring Boot application, it tries to compile the code generated at runtime, but fails (on MacOS and Ubuntu): 

```

/Q_01943648192_1275614662_1275614662_1195259493.java:3: error: package villuv.springboot.querydsl.bug.app does not exist
    public static Iterable<villuv.springboot.querydsl.bug.app.TheBean> eval(Iterable<villuv.springboot.querydsl.bug.app.TheBean> bean_, String a1) {
                                                                                                                  ^
/Q_01943648192_1275614662_1275614662_1195259493.java:3: error: package villuv.springboot.querydsl.bug.app does not exist
    public static Iterable<villuv.springboot.querydsl.bug.app.TheBean> eval(Iterable<villuv.springboot.querydsl.bug.app.TheBean> bean_, String a1) {
                                                             ^
/Q_01943648192_1275614662_1275614662_1195259493.java:4: error: package villuv.springboot.querydsl.bug.app does not exist
java.util.List<villuv.springboot.querydsl.bug.app.TheBean> rv = new java.util.ArrayList<villuv.springboot.querydsl.bug.app.TheBean>();
                                                 ^
/Q_01943648192_1275614662_1275614662_1195259493.java:4: error: package villuv.springboot.querydsl.bug.app does not exist
java.util.List<villuv.springboot.querydsl.bug.app.TheBean> rv = new java.util.ArrayList<villuv.springboot.querydsl.bug.app.TheBean>();
                                                                                                                          ^
/Q_01943648192_1275614662_1275614662_1195259493.java:5: error: package villuv.springboot.querydsl.bug.app does not exist
for (villuv.springboot.querydsl.bug.app.TheBean bean : bean_) {
                                       ^
/Q_01943648192_1275614662_1275614662_1195259493.java:7: error: package com.querydsl.collections does not exist
        if (com.querydsl.collections.CollQueryFunctions.equals(bean.getValue(), a1)) {
                                    ^
6 errors
   at com.querydsl.codegen.utils.JDKEvaluatorFactory.compile(JDKEvaluatorFactory.java:74 undefined) ~[codegen-utils-5.0.0.M1.jar!/:na]
[...]

PR #2968, opened 2021-07-14
Collections: Failure of setting up classpath for javac in case of run…

viktorzub:
…ning in Spring Boot on Ubuntu and MacOS

I have tried this, but it work's only on redhat linux, still doesn't work on macos

code taken from https://github.com/querydsl/codegen/pull/43/files
code should be focused to fix https://github.com/querydsl/querydsl/issues/2964

Can some one review and help to develop proper fix

### DIFF
--- a/querydsl-codegen-utils/src/main/java/com/querydsl/codegen/utils/CompiledJavaFileObject.java
+++ b/querydsl-codegen-utils/src/main/java/com/querydsl/codegen/utils/CompiledJavaFileObject.java
@@ -1,0 +1,132 @@
+/*
+ * Copyright 2018, The Querydsl Team (http://www.querydsl.com/team)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


jwgmeligmeyling:
I'll have a look tonight (CEST)

jwgmeligmeyling:
Apparently it also doesn't work on Ubuntu Linux, at least not on Java 11. (Hence the CI failure). I'll do some more testing locally.

viktorzub:
@jwgmeligmeyling any updates here? I also still not able to make it work at macos(

jwgmeligmeyling:
Could you try using the ecj compiler as a workaround?

jwgmeligmeyling:
This approach is still system dependent and has issues on Java 11.

I think we should deprecate the `JDKEvaluatorFactory` and instead encourage the use of `ECJEvaluatorFactory`. Under JRE's, this evaluator factory should be selected by default.

viktorzub:
@jwgmeligmeyling i was not able to make it work with ECJ compiller with java 11 also(

PR #2983, opened 2021-07-29
Fix #2982

### LINKED ISSUE #2982: "IllegalArgumentException: eq(null) is not allowed. Use isNull() instead" on collection with one element, which is null
## Observed vs. expected behavior

Actual: when passing collection with one parameter, which is null to the "in(T... right)" method, I get "IllegalArgumentException: eq(null) is not allowed. Use isNull() instead"

Expected: I don't get the above mentioned exception when passing collection with one element, which is null. I get results containing null in a field.

## Steps to reproduce

In order to reproduce this bug, just use in(T... right) method located here at the moment of writing:
https://github.com/querydsl/querydsl/blob/master/querydsl-core/src/main/java/com/querydsl/core/types/dsl/SimpleExpression.java#L205
with a collection containing only one element, which is a null value.
Doing so you will get "java.lang.IllegalArgumentException: eq(null) is not allowed. Use isNull() instead" as we're using return eq(right[0]); and we should use isNull() instead.

## Environment

Querydsl version: 4.4.0

Querydsl module: querydsl-core

Database: Oracle

JDK: 8

## Additional details

In my opinion it's a simple fix and I can create a PR right after this Bug report.

Chriserus:
This null check fixes bug I described in ticket #2982

### DIFF
--- a/querydsl-core/src/main/java/com/querydsl/core/types/dsl/SimpleExpression.java
+++ b/querydsl-core/src/main/java/com/querydsl/core/types/dsl/SimpleExpression.java
@@ -204,7 +204,7 @@ public abstract class SimpleExpression<T> extends DslExpression<T> {
      */
     public BooleanExpression in(T... right) {
         if (right.length == 1) {
-            return eq(right[0]);
+            return right[0] === null ? isNull() : eq(right[0]);
         } else {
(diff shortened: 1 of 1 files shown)


jwgmeligmeyling:
Closing as per the discussion on https://github.com/querydsl/querydsl/pull/2577 .

The required use of `isNull()` is very much intentional.

Issue #2979, opened 2021-07-27
OSGI manifest is missing for 5.0.0


jherkel:
## Observed vs. expected behavior

OSGI manifest is missing for querydsl 5.0.0

## Steps to reproduce
Open final JAR and check MANIFEST.MF.  
```
Querydsl 5.0.0 :
Manifest-Version: 1.0
Automatic-Module-Name: com.querydsl.core
Build-Jdk-Spec: 1.8
Created-By: Maven Jar Plugin 3.2.0
```

Querydsl 4.4.0:
```
Manifest-Version: 1.0
Bundle-Description: core module for querydsl
Bundle-License: http://www.apache.org/licenses/LICENSE-2.0.txt
Bundle-SymbolicName: com.querydsl.core
Archiver-Version: Plexus Archiver
Built-By: jkt
Bnd-LastModified: 1600524810352
Bundle-ManifestVersion: 2
Bundle-DocURL: http://www.querydsl.com
Bundle-Vendor: Querydsl
Import-Package: com.google.common.base;version="18.0.0",com.google.com
 mon.cache;version="18.0.0",com.google.common.collect;version="18.0.0"
 ,com.google.common.primitives;version="18.0.0",com.mysema.commons.lan
 g;version="[0.2,1)",javax.annotation,javax.lang.model,net.sf.cglib.pr
 oxy;resolution:=optional
Tool: Bnd-2.1.0.20130426-122213
Export-Package: com.querydsl.core;uses:="com.mysema.commons.lang,com.q
 uerydsl.core.types,javax.annotation";version="4.4.0",com.querydsl.cor
 e.alias;uses:="com.querydsl.core.types,com.querydsl.core.types.dsl,ja
 vax.annotation";version="4.4.0",com.querydsl.core.annotations;version
 ="4.4.0",com.querydsl.core.dml;uses:="com.querydsl.core,com.querydsl.
 core.types,javax.annotation";version="4.4.0",com.querydsl.core.group;
 uses:="com.mysema.commons.lang,com.querydsl.core,com.querydsl.core.ty
 pes,com.querydsl.core.types.dsl";version="4.4.0",com.querydsl.core.su
 pport;uses:="com.mysema.commons.lang,com.querydsl.core,com.querydsl.c
 ore.types,com.querydsl.core.types.dsl,javax.annotation";version="4.4.
 0",com.querydsl.core.types;uses:="com.google.common.base,com.google.c
[...]

PR #3016, opened 2021-09-06
[#2979] fix problem with generating OSGI manifests

jherkel:
I had to change some things in master pom : 
1) maven-checkstyle-plugin - I had a problem with a generated class within querydsl-osgi-itest-jpa maven subproject
2) maven-javadoc-plugin - I needed to append javax.interceptor/javax.interceptor-api because I  had a problem with generating javadocs with JDK8 when @Transactional annotation was used
3) Create additional 'osgi' profile


F43nd1r (approved):
Looks good to me, but please refrain from commiting unrelated whitespace changes, makes it harder to read the diff.

pom.xml
```diff
@@ -74,9 +76,7 @@
     <dokka.version>1.5.0</dokka.version>
 
     <!-- Import-Package definitions for maven-bundle-plugin -->
-    <osgi.import.package.root>
-      *
-    </osgi.import.package.root>
+    <osgi.import.package.root>*</osgi.import.package.root>
```
jwgmeligmeyling, line 79: Isn't `com.querydsl.*` a more sensible default then `*` here?

pom.xml
```diff
@@ -819,6 +836,16 @@
         <module>querydsl-maven-plugin</module>
       </modules>
     </profile>
+    
+    <profile>
+      <id>osgi</id>
```
jwgmeligmeyling, line 841: What is the purpose of this `osgi` module?

querydsl-jdo/pom.xml
```diff
@@ -146,6 +146,7 @@
         </executions>
         <configuration>
           <archive>
+            <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
```
jwgmeligmeyling, line 149: Are manifest entries (below) still added if a manifest file is provided?

jwgmeligmeyling (commented):
I have a few questions, besides that the changes look good.

Thanks for adding an example case as well, this will really help us to maintain OSGi support down the road!

pom.xml
jherkel: no, '*' is default for bnd and it means that bnd imports all referred packages

querydsl-jdo/pom.xml
jherkel: yes, you can check that final jar contains also Automatic-Module-Name. These missing OSGI headers can be fixed also that we change <packaging> from jar to bundle. (bundle also create a jar archive)

pom.xml
jherkel: I wanted to select out these integration tests for OSGI.

F43nd1r:
@jwgmeligmeyling questions seem answered, is this good to merge?

jwgmeligmeyling:
I reviewed it, but didn't have the change to test it (package and compare the manifests for each of the modules).

PR #3041, opened 2021-09-23
Use maven-compiler-plugin for annotation processing

### LINKED ISSUE #2091: Document how querydsl-apt classifiers can be used
I think it’s cleaner to have the `querydsl-apt` dependency only on the build plugin instead of using the `provided` scope. So instead of

```XML
<dependency>
  <groupId>com.querydsl</groupId>
  <artifactId>querydsl-apt</artifactId>
  <version>${querydsl.version}</version>
  <scope>provided</scope>
</dependency>
```
do this:
```XML
<build>
  <plugins>
    ...
    <plugin>
      <groupId>com.mysema.maven</groupId>
      <artifactId>apt-maven-plugin</artifactId>
      <version>1.1.3</version>
      <executions>
        <execution>
          <goals>
            <goal>process</goal>
          </goals>
          <configuration>
            <outputDirectory>target/generated-sources/java</outputDirectory>
            <processor>com.querydsl.apt.jpa.JPAAnnotationProcessor</processor>
          </configuration>
        </execution>
      </executions>
      <dependencies>
        <dependency>
          <groupId>com.querydsl</groupId>
          <artifactId>querydsl-apt</artifactId>
          <version>${querydsl.version}</version>
        </dependency>
      </dependencies>
    </plugin>
    ...
  </plugins>
```

This way the IDE doesn’t suggest classes from the `querydsl-apt` dependency which aren’t available at runtime.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/41714214-document-how-querydsl-apt-classifiers-can-be-used?utm_campaign=plugin&utm_content=tracker%2F303568&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F303568&utm_medium=issues&utm_source=github).
</bountysource-plugin>

devoto13:
...instead of the unsupported apt-maven-plugin.

The main advantage of this change is that Intelij IDEA would automatically pick up the annotation processor and run it when building inside IDE. Thus eliminating the need to manually run `mvn generate-sources` or `mvn compile` to get the generated classes every time after changing the branch. maven-compiler-plugin also properly cleans up the generated sources whenever a model is removed, unlike the apt-maven-plugin, where user has to manually go and remove the no-longer-needed Q-class from the build folder as otherwise compilation will fail.

Closes #2091

Creating the PR as a draft just to demonstrate the idea and get feedback. If you think that this is the right approach, I can extend it to all querydsl modules and update documentation.

### DIFF
--- a/querydsl-jpa/README.md
+++ b/querydsl-jpa/README.md
@@ -22,27 +22,34 @@ And now, configure the Maven APT plugin :
     <plugins>
       ...
       <plugin>
-        <groupId>com.mysema.maven</groupId>
-        <artifactId>apt-maven-plugin</artifactId>
-        <version>1.1.3</version>
-        <executions>
-          <execution>
-            <goals>
-              <goal>process</goal>
-            </goals>
-            <configuration>
-              <outputDirectory>target/generated-sources/java</outputDirectory>
-              <processor>com.querydsl.apt.jpa.JPAAnnotationProcessor</processor>
-            </configuration>
-          </execution>
-        </executions>
-        <dependencies>
-          <dependency>
-            <groupId>com.querydsl</groupId>
-            <artifactId>querydsl-apt</artifactId>
-            <version>${querydsl.version}</version>
-          </dependency>
-        </dependencies>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <version>3.8.1</version>
+        <configuration>
+            <annotationProcessors>
+                <annotationProcessor>com.querydsl.apt.jpa.JPAAnnotationProcessor</annotationProcessor>
+            </annotationProcessors>
+            <annotationProcessorPaths>
+                <path>
+                    <groupId>com.querydsl</groupId>
+                    <artifactId>querydsl-apt</artifactId>
+                    <version>${querydsl.version}</version>
+                    <classifier>jpa</classifier>
+                </path>
+                <path>
+                    <groupId>jakarta.persistence</groupId>
+                    <artifactId>jakarta.persistence-api</artifactId>
+                    <version>2.2.3</version>
+                </path>
+                <!-- Only necessary when using JDK 11+ -->
+                <path>
+                    <groupId>jakarta.annotation</groupId>
+                    <artifactId>jakarta.annotation-api</artifactId>
+                    <version>1.3.5</version>
+                </path>
+            </annotationProcessorPaths>
+        </configuration>
+        ...
       </plugin>
       ...
     </plugins>
--- a/querydsl-jpa/pom.xml
+++ b/querydsl-jpa/pom.xml
@@ -223,7 +223,7 @@
   </dependencies>
   
   <build>
-    <plugins> 
+    <plugins>
     
       <plugin>
         <groupId>org.apache.felix</groupId>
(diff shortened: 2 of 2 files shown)


devoto13:
Originally tested this in my own project, which is still on 4.3.1, but looks like with 5.0.0 `jakarta.annotation-api` is not needed at all.

querydsl-jpa/README.md
```diff
@@ -22,27 +22,34 @@ And now, configure the Maven APT plugin :
...
+        <configuration>
+            <annotationProcessors>
+                <annotationProcessor>com.querydsl.apt.jpa.JPAAnnotationProcessor</annotationProcessor>
+            </annotationProcessors>
+            <annotationProcessorPaths>
+                <path>
+                    <groupId>com.querydsl</groupId>
+                    <artifactId>querydsl-apt</artifactId>
+                    <version>${querydsl.version}</version>
+                    <classifier>jpa</classifier>
```
jwgmeligmeyling, line 37: The classifier here is unnecessairy.

Either you use:

```xml
<dependency>
  <groupId>com.querydsl</groupId>
  <artifactId>querydsl-apt</artifactId>
  <version>${querydsl.version}</version>
  <scope>provided</scope>
  <classifier>jpa</classifier>
</dependency>
```

as _dependency_. Which automatically registers the `com.querydsl.apt.jpa.JPAAnnotationProcessor` as APT processor through a service loader. This means it hooks onto the compiler directly, does not require additional plugins and also just works with compilation from your IDE. This plus its simplicity IMO makes this approach preferred over the plugin, however... 

If you use the plugin, then specifying a classifier is unnecessary there, because you have to provide the `annotationProcessor` class anyway.

querydsl-jpa/README.md
```diff
@@ -22,27 +22,34 @@ And now, configure the Maven APT plugin :
...
+                </path>
+                <path>
+                    <groupId>jakarta.persistence</groupId>
+                    <artifactId>jakarta.persistence-api</artifactId>
+                    <version>2.2.3</version>
+                </path>
+                <!-- Only necessary when using JDK 11+ -->
```
jwgmeligmeyling, line 44: Not at all necessairy in Querydsl 5.0.0 indeed

jwgmeligmeyling:
Don't forget to update the examples as well:

* https://github.com/querydsl/querydsl/blob/99e3ca936f46471bc61d9af6e1a3dbb16f3f12b3/querydsl-docs/src/main/docbook/en-US/content/tutorials/jpa.xml#L51-L66
* https://github.com/querydsl/querydsl/blob/99e3ca936f46471bc61d9af6e1a3dbb16f3f12b3/querydsl-docs/src/main/docbook/en-US/content/tutorials/jdo.xml#L47-L62
* https://github.com/querydsl/querydsl/blob/99e3ca936f46471bc61d9af6e1a3dbb16f3f12b3/querydsl-docs/src/main/docbook/en-US/content/general/codegen.xml#L228-L247
* https://github.com/querydsl/querydsl/blob/99e3ca936f46471bc61d9af6e1a3dbb16f3f12b3/querydsl-docs/src/main/docbook/en-US/content/tutorials/collections.xml#L167-L182

querydsl-jpa/README.md
devoto13: I see. Thanks for the explanation! 

Do you think we should add both approaches to the docs and let users decide which they prefer? Maybe present one as recommended? Or only document one? Which one? 

As far as I understand the drawback of the approach from this PR is that one would need to specify some extra dependencies depending on the processor they want to use so that the annotations are on the AP classpath. I assume it is not really a problem, since people will normally just copy-paste the snippet from the docs.

While the issue of using querydsl-apt as a provided dependency is that classes from it will appear in the completion. This is probably a far-fetched issue, but some people seem to [find it annoying](https://github.com/querydsl/querydsl/issues/2091#issuecomment-869964558). It surely requires less configuration, but IMO explicit is better than implicit.

We may want to document both as if the user already uses another annotation processor, then they should use the same approach for adding querydsl-apt as it will either disable the other annotation processor or the querydsl processor will do nothing upon adding it. The drawback is that we try to cover a lot of corner cases making documentation harder to follow.

I'm thinking to document the approach from this PR as recommended (explicit better than implicit) and mention the existing annotation processors' caveat pointing out the provided dependency approach as a solution to those. But let me know if you think differently.

jwgmeligmeyling: > Do you think we should add both approaches to the docs and let users decide which they prefer? Maybe present one as recommended? Or only document one? Which one?

Probably best to do both.

1) The approach using the Maven plugin allows for additional configuration to be set
2) The approach using the Maven plugin might be more recognisable for Spring Data MongoDB users, as Spring Data provides an extended Querydsl Annotation Processor for MongoDB.
3) The approach using the Maven plugin might be preferred in specific situations where multiple APT plugins are used and are meant to be executed in a particular order.

PR #3053, opened 2021-10-10
Maven profile introduced in order to be able to switch off jpa integration tests against certain databases.

dhaase:
Since not every contributor has installed all databases for which there are integration tests in the JPA module, it should be possible to switch the integration tests of certain databases on or off if necessary.

This pull request enables integration tests of certain databases to be switched on or off.


F43nd1r:
This PR is unlikely to be merged in its current state, because we have a docker file to start all required databases easily and it mixes a dependeny upgrade with a feature.

PR #3076, opened 2021-10-31
Improve Kotlin Codegen Spring Data Interop

F43nd1r:
Spring Data tries to access a field using reflection. Make it available by `@JvmField`


xenoterracide:
when will there be a release including this?

hamidsafdari:
I literally spent an entire day reaching this issue. Is there a version that includes this fix?

Issue #3109, opened 2021-12-07
Cannot set multiple QueryHints with same key


MelleD:
With the git commit "5abd478fc30e800e87065ffcc7b80d76672b2d02. Removed multimap guava occurrenes" it is not possible to use QueryHints with the same key, because the hashmap cannot save multiple keys.

However, this is necessary for some QueryHints. In addition, the JPA specification defines that a key has multiple values. For this reason, the multi map should be reintroduced in AbstractJPAQuery

## Steps to reproduce

Create a Entity with two 1:n relations. Every relation should/fetched in a single select with a BATCH in Query

```

   @QueryHints( {
         @QueryHint( name = BATCH_TYPE, value = "IN" ),
         @QueryHint( name = BATCH, value = "t.bar1" ),
         @QueryHint( name = BATCH, value = "t.bar2" ),
   } )
   @EntityGraph( value = "test", type = EntityGraph.EntityGraphType.LOAD )
   Page<Foo> findAll( Predicate predicate, Pageable pageable );
```

Querydsl version: 5.0.0

igormilina:
I can confirm that this is a regression bug introduced in Querydsl v5 ("Removed multimap guava occurrenes").

I originally reported issue: https://github.com/querydsl/querydsl/issues/348 and requested that query hints be stored in some sort of multimap, so that was the reason Timo introduced worked great.

MelleD:
How you see there is already an open PR. But nobody cares
Not sure if the project is dead?

igormilina:
Could someone please take a look at PR that MelleD committed? 
As I said, it's a regression bug, and the fix is not really complicated.
This bug is realy hurting the performance of the aplication as we cannot set query hints to adress n+1 problems.

igormilina:
This issue is still open and now even marked as stale? It's a regression bug. It is not hard to fix. The working PR is submitted a year ago, it has 10 lines of code. And there is nobody to review and merge it? I guess project is really dead, a shame.

This issue alone is the reason I forked this repo a year ago and started using my custom version of the library with this issue fixed.

igormilina:
Bump?

MelleD:
@igormilina maybe switch to https://github.com/OpenFeign/querydsl

PR #3110, opened 2021-12-07
Fix using multi query hints

MelleD:
#3109

### DIFF
--- a/querydsl-jpa/pom.xml
+++ b/querydsl-jpa/pom.xml
@@ -76,6 +76,12 @@
       <optional>true</optional>
     </dependency>
 
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+      <version>${guava.version}</version>
+    </dependency>
+    
     <!-- test -->
     <dependency>
       <groupId>org.hibernate.validator</groupId>
--- a/querydsl-jpa/pom.xml
+++ b/querydsl-jpa/pom.xml
@@ -76,12 +76,6 @@
       <optional>true</optional>
     </dependency>
 
-    <dependency>
-      <groupId>com.google.guava</groupId>
-      <artifactId>guava</artifactId>
-      <version>${guava.version}</version>
-    </dependency>
-    
     <!-- test -->
     <dependency>
       <groupId>org.hibernate.validator</groupId>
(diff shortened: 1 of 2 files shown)


jwgmeligmeyling:
Please use a collection as value in the map instead. We don’t want a guava dependency in JPA.

also, what is the use case for this? Afaik jpa itself does not support multiple query hints for the same key.

MelleD:
spring data and JPA support multiple values. 

Without multiple values the select looks different. Then just the last entry wins

In this example without multiple values you have n+1 issue. For first value looks like

First select n times
`select * from bar1 where id = ? `
`select * from bar1 where id = ? `
Second select 
`select * from bar2 where id IN (?,?...) `


With this fix both works (works currently in spring data)
First select 
`select * from bar1 where IN (?,?...) `
Second select 
`select * from bar2 where id IN (?,?...) `

In the parent pom there is already guava available, but ok i can remove the deps

MelleD:
Here you can check in EJBQueryImpl in EclipseLink that the key can have multiple values

`
       static DatabaseQuery apply(String hintName, boolean shouldUseDefault, Object hintValue, DatabaseQuery query, ClassLoader loader, AbstractSession activeSession) {
            Hint hint = (Hint)mainMap.get(hintName);
            if (hint == null) {
                // unknown hint name - silently ignored.
                return query;
            }

            Map<String, Object> existingHints = (Map<String, Object>)query.getProperty(QUERY_HINT_PROPERTY);
            if (existingHints == null){
                existingHints = new HashMap<String, Object>();
                query.setProperty(QUERY_HINT_PROPERTY, existingHints);
            }
            existingHints.put(hintName, hintValue);

            return hint.apply(hintValue, shouldUseDefault, query, loader, activeSession);
        }
`

MelleD:
@jwgmeligmeyling jwgmeligmeyling please have a look. Removed guava

MelleD:
@jwgmeligmeyling friendly reminder, please have a look :)

MelleD:
@jwgmeligmeyling friendly reminder, please have a look :)

MelleD:
@sangyongchoi
@jwgmeligmeyling 
@meatballhat 
@rocketraman 

can someone have look?

Issue #2223, opened 2017-11-23
rows().preceding not working on MS SQL


rgra:
SQLExpressions.sum(tr.value)
		.over()
		.partitionBy( tr.productId)
		.orderBy(tr.valueDate)
		.rows().preceding(3);

This creates the following SQL
sum( "value" ) over(partition by product_id order by value_date desc rows preceding 3)

This won't execute on MS SQL Server 2012+ because it has to be
rows 3 preceding 

Currently there  is no way to fix this with templates, because it is hard coded in WindowRows.

jwgmeligmeyling:
Accepting PRs.

allanhaywood:
@jwgmeligmeyling I'm looking into this. Can I be assigned?

allanhaywood:
Created PR: https://github.com/querydsl/querydsl/pull/3111

PR #3111, opened 2021-12-08
Fixing issue 2223

allanhaywood:
Updated tests to match desired sql syntax for preceding and following
Updated WindowRows to produce the proper syntax for preceding and
following

NOTE: I looked up several documents on the expected sql syntax for preceding and following, and all of them had the integers prior to the keyword.

Here is an example: https://www.sqlite.org/windowfunctions.html

### DIFF
--- a/querydsl-sql/src/main/java/com/querydsl/sql/WindowRows.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/WindowRows.java
@@ -58,8 +58,8 @@ public class WindowRows<A> {
 
         public BetweenAnd preceding(Expression<Integer> expr) {
             args.add(expr);
-            str.append(PRECEDING);
             str.append(" {").append(offset++).append("}");
+            str.append(PRECEDING);
             return new BetweenAnd();
         }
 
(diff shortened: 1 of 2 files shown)


allanhaywood:
This is for issue: https://github.com/querydsl/querydsl/issues/2223

jwgmeligmeyling:
Doesn't this just break it for other dialects?

It seems a rather strange "fix".

allanhaywood:
@jwgmeligmeyling I'm new to this code base, so any feedback would be appreciated. 

I did do a search for what the accepted syntax is, and didn't find any that were the other way around. Also, I couldn't find where either the tests or the windowsrows class had support for different behavior for different dialects.

If you could provide additional guidance on how to handle supporting multiple dialects (perhaps point me to another area of the codebase that has the existing pattern), and I'll look into it.

allanhaywood:
@jwgmeligmeyling I've  been looking through the code and trying to figure out how to templatize the changes.
This is what I've done so far, but it isn't entirely working yet.
https://github.com/allanhaywood/querydsl/tree/2223_WIP

For the WindowFunctionTest, the all test passes, it uses the SQLExpressions directly)
For the tests that use WIndowsRows, those are failing for unknown operation.

Any guidance would be appreciated.

Thanks!

PR #3117, opened 2021-12-10
Edit1

mz4987:


jwgmeligmeyling (changes requested):
Should take expressions instead of constants. Should not be added to `MathExpressions` but to a `DateExpressions` utility class.

PR #3215, opened 2022-03-23
Fixes #3212 (implement a code template for the @Generated annotation)

### LINKED ISSUE #3212: custom Generated annotations
## Why the new feature should be added

We're using JaCoCo to compute test-coverage.
JaCoCo supports ignoring generated code if the code has annotations named `@Generated` with `RetentionPolicy.CLASS` or `RUNTIME`.

Unfortunately, `javax.annotation.processing.Generated` (and also `javax.annotation.Generated` for that matter) has just  `RetentionPolicy.SOURCE` and thus cannot be seen by JaCoCo (or any other class-file based analyzer for that matter).
Configuration for ignoring QueryDSL Q-classes is non-standard, tedious and error prone since the only other option is filename-patterns.

So I'd like to use some of the other existing `@Generated` annotations or of one of our own.

## How the new feature should work
Implement a parameter for the source generator (and (e.g.) maven-plugin) that receives a fully qualified name to an annotation class.
If this parameter is not set, the generator could fallback to current behavior.

For a more generic option: allow for a source code template that might also allow filling in fields of the annotation.
Example: 
`generated-template=javax.annotation.processing.Generated(value="{{queryDSL.generatorClass}}", date="{{queryDSL.timestamp}}", comments="Version: {{queryDSL.version}}")`

HonoluluHenk:
This pr implements a configurable code template for the Generated annotations (see #3212 )

I'd reeeeally like to have custom annotations that do not follow the Single Value Annotation pattern.

Feel free to reject this pr, I had fun implementing it... which is rewarding in and of itself :)

### DIFF
--- a/querydsl-apt/src/main/java/com/querydsl/apt/DefaultConfiguration.java
+++ b/querydsl-apt/src/main/java/com/querydsl/apt/DefaultConfiguration.java
@@ -282,8 +282,8 @@ public class DefaultConfiguration implements Configuration {
 
         module.loadExtensions();
 
-        Class<? extends Annotation> generatedAnnotationClass = GeneratedAnnotationResolver.resolve(options.get(QUERYDSL_GENERATED_ANNOTATION_CLASS));
-        module.bindInstance(CodegenModule.GENERATED_ANNOTATION_CLASS, generatedAnnotationClass);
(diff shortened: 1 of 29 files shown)


jwgmeligmeyling:
Mind sharing a bit on your opinion regarding the Single Value Annotation pattern? I am wondering how this would be used. To implement a stereotype like annotation?

PR #3304, opened 2022-07-01
use isNull() when the only one elemnet is null

HLuff:
image

    public void testQuery() {
        List<Long> list = new ArrayList<>();
        list.add(null);
        try{
            roleRepository.findAll(QRole.role.id.in(list));
            logger.info("[null] success");
        }catch (Exception e){
            logger.info("[null] exception {}", e.getMessage());
        }

        list.add(null);
        try{
            roleRepository.findAll(QRole.role.id.in(list));
            logger.info("[null, null] success");
        }catch (Exception e){
            logger.info("[null, null] exception {}", e.getMessage());
        }
    }

in [null, null] is ok, but in [null] is error.


jwgmeligmeyling:
The code is error prone and makes false assumptions on the correlation between Java's `null` and SQL's `NULL` which are actually quite different.

PR #3346, opened 2022-08-19
kotlin-codegen: Improved Kotlin compatibility

zbin:
In Kotlin, `Long::class.java` is java Primitive Type.
In `kotlin-codegen`, Using ` T::class.java` will cause exception, because `Number` type is Primitive type.
```kotlin
// com.querydsl.kotlin.codegen.Extensions.kt
fun ClassName.asClassStatement() = CodeBlock.of("%T::class.java", this)

// com.querydsl.core.util.MathUtils.kt
    public static <D extends Number> D cast(Number num, Class<D> type) {
        D rv;
        if (num == null || type.isInstance(num)) {
            rv = type.cast(num);
        } else if (type.equals(Byte.class)) {
            rv = type.cast(num.byteValue());
        } else if (type.equals(Double.class)) {
            rv = type.cast(num.doubleValue());
        } else if (type.equals(Float.class)) {
            rv = type.cast(num.floatValue());
        } else if (type.equals(Integer.class)) {
            rv = type.cast(num.intValue());
        } else if (type.equals(Long.class)) {
            rv = type.cast(num.longValue());
        } else if (type.equals(Short.class)) {
            rv = type.cast(num.shortValue());
        } else if (type.equals(BigDecimal.class)) {
            rv = type.cast(new BigDecimal(num.toString()));
        } else if (type.equals(BigInteger.class)) {
            if (num instanceof BigDecimal) {
                rv = type.cast(((BigDecimal) num).toBigInteger());
            } else {
                rv = type.cast(new BigInteger(num.toString()));
            }
        } else {
            throw new IllegalArgumentException(String.format("Unsupported target type : %s", type.getSimpleName()));
        }
        return rv;
    }
```

see kotlin source code:
```kotlin
public val <T : Any> KClass<T>.javaObjectType: Class<T>
    get() {
        val thisJClass = (this as ClassBasedDeclarationContainer).jClass
        if (!thisJClass.isPrimitive) return thisJClass as Class<T>

        return when (thisJClass.name) {
            "boolean" -> JavaLangBoolean::class.java
            "char"    -> JavaLangCharacter::class.java
            "byte"    -> JavaLangByte::class.java
            "short"   -> JavaLangShort::class.java
            "int"     -> JavaLangInteger::class.java
            "float"   -> JavaLangFloat::class.java
            "long"    -> JavaLangLong::class.java
            "double"  -> JavaLangDouble::class.java
            "void"    -> Void::class.java
            else -> thisJClass
        } as Class<T>
    }
```

We need use `Long::class.javaObjectType`

### DIFF
--- a/querydsl-kotlin-codegen/src/main/kotlin/com/querydsl/kotlin/codegen/Extensions.kt
+++ b/querydsl-kotlin-codegen/src/main/kotlin/com/querydsl/kotlin/codegen/Extensions.kt
@@ -15,6 +15,7 @@ package com.querydsl.kotlin.codegen
 
 import com.querydsl.codegen.EntityType
 import com.querydsl.codegen.TypeMappings
+import com.querydsl.codegen.utils.model.ClassType
 import com.querydsl.codegen.utils.model.Type
 import com.squareup.kotlinpoet.ClassName
 import com.squareup.kotlinpoet.CodeBlock
@@ -49,6 +50,7 @@ fun Type.asClassName(): ClassName = when (this.fullName) {
     "long[]" -> LongArray::class.asClassName()
     "float[]" -> FloatArray::class.asClassName()
     "double[]" -> DoubleArray::class.asClassName()
+    "java.lang.Object" -> Any::class.asClassName()
     "java.lang.String" -> String::class.asClassName()
     "java.util.List" -> List::class.asClassName()
     "java.util.Map" -> Map::class.asClassName()
@@ -63,9 +65,17 @@ private fun Type.enclosingTypeHierarchy(): List<String> {
     return generateSequence { current?.also { current = it.enclosingType }?.simpleName }.toList().asReversed()
 }
 
-fun ClassName.asClassStatement() = CodeBlock.of("%T::class.java", this)
+fun ClassName.asClassStatementForJavaObjectType():CodeBlock =
+    CodeBlock.of("%T::class.javaObjectType", this)
+fun ClassName.asClassStatement(): CodeBlock =
+    CodeBlock.of("%T::class.java", this)
+fun ClassName.asClassStatement(classType: KotlinClassType): CodeBlock =
+    CodeBlock.of("%T::class.${classType.value}", this)
 
-fun Type.asClassNameStatement() = asClassName().asClassStatement()
+fun Type.asClassNameStatementForJavaObjectType() = asClassName().asClassStatement(KotlinClassType.JAVA_OBJECT_TYPE)
+fun Type.asClassNameStatement() = asClassName().asClassStatement(KotlinClassType.JAVA)
+
+fun Type.asClassNameStatement(classType: KotlinClassType) = asClassName().asClassStatement(classType)
 
 fun TypeMappings.getPathClassName(type: Type, model: EntityType) = getPathType(type, model, true).asClassName()
 
@@ -84,3 +94,10 @@ fun Collection<String>.joinToCode(
 fun Any.asCodeBlock(format: String = "%L") = CodeBlock.of(format, this)
 
 fun ParameterSpec.asCodeBlock(format: String = "%N") = (this as Any).asCodeBlock(format)
+
+enum class KotlinClassType(val value: String) {
+    JAVA("java"),
+    JAVA_CLASS("javaClass"),
+    JAVA_OBJECT_TYPE("javaObjectType"),
+    JAVA_PRIMITIVE_TYPE("javaPrimitiveType");
+}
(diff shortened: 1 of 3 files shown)


F43nd1r:
Hi @zbin looks good, but could you add a test case covering this issue?

zbin:
> Hi @zbin looks good, but could you add a test case covering this issue?

You're Right. I will add it later

zbin:
I am not sure if this covers the test cases correctly. 

But I have to say that APT-based integration tests are not really easy to write

IceBlizz6:
I generated kotlin Q classes.
Trying to run queries by comparing numbers.

`qUser -> qUser.id.gt(42)`

> Exception in thread "main" java.lang.IllegalArgumentException: Unsupported target type : int
> 	at com.querydsl.core.util.MathUtils.cast(MathUtils.java:86)
> 	at com.querydsl.core.types.dsl.NumberExpression.cast(NumberExpression.java:178)
> 	at com.querydsl.core.types.dsl.NumberExpression.gt(NumberExpression.java:337)

`eq` function works, but `gt` and `in` gives errors.
i get matching errors for Double and Long.
I suspect that maybe `eq` isn't using cast.

I believe that this PR probably solves this, so i won't make a new issue for it.
Would be great if this could be prioritized.
When can we expect this to be released?

IceBlizz6:
@zbin Are you able to take another look at this PR?
I think that it may not have been completed because it is reporting errors in the build process.
Would be really appreciated!

jwgmeligmeyling:
Yea I actually rerun the tests to see what happened. There is a binary incompatibility issue. However, the next release will be a major release anyway so we can ignore it. That can either be done through the plugin configuration or by bumping the Querydsl version to `6.0.0-SNAPSHOT` (to indicate that the next release will be a major one).

zbin:
@IceBlizz6 sorry for the late reply.
I tried to fix the unit test, but as @jwgmeligmeyling said, the binary compatibility change must wait until `6.0.0-SNAPSHOT`

One thing I don't understand is why the test outputs :
```
[ERROR] Failed to execute goal com.github.siom79.japicmp:japicmp-maven-plugin:0.15.3:cmp (default) on project querydsl-kotlin-codegen: There is at least one incompatibility: com.querydsl.kotlin.codegen.ExtensionsKt.asClassNameStatementForJavaObjectType(com.querydsl.codegen.utils.model.Type):METHOD_REMOVED,com.querydsl.kotlin.codegen.ExtensionsKt.asClassStatementForJavaObjectType(com.squareup.kotlinpoet.ClassName):METHOD_REMOVED -> [Help 1]
```

I didn't find any reference to it in the code, and now I had to add this code even though it doesn't have any references

meloning:
Hello, I also had the same issue with version 5.0.0.



It seems that the PR has not been reflected yet, so I temporarily solved it as follows:
```kotlin
private fun ltTableId(lastTableId: Long): BooleanExpression {
    return qTableEntity.id.lt(Expressions.asNumber(lastTableId))
}
```
[NumberExpression.kt]
```kotlin
    /**
     * Create a {@code this < right} expression
     *
     * @param <A>
     * @param right rhs of the comparison
     * @return {@code this < right}
     * @see java.lang.Comparable#compareTo(Object)
     */
    public final <A extends Number & Comparable<?>> BooleanExpression lt(A right) {
        return lt(ConstantImpl.create(cast(right)));
    }

    /**
     * Create a {@code this < right} expression
     *
     * @param <A>
     * @param right rhs of the comparison
     * @return {@code this < right}
     * @see java.lang.Comparable#compareTo(Object)
     */
    public final <A extends Number & Comparable<?>> BooleanExpression lt(Expression<A> right) {
        return Expressions.booleanOperation(Ops.LT, this, right);
    }
```
I solved it by changing the type to the Expression type through the `lt(Expression<A> right)` function.

Hope this helps anyone who is having problems.

zbin:
Sorry, I am no longer using the Lib and therefore will not be following up on this PR. 
Invite other interested friends to maintain or submit separately.

---

I may not have much time to continue open source maintenance as a developer working in China. Very hyped work has filled my life and I'm miserable.

IceBlizz6:
@meloning Would you be able to create a new PR with your changes?
seems like this one was closed now.

My team really needs a working QueryDSL for Kotlin so would be appreciated.

meloning:
Hello, @IceBlizz6 

The comment I posted was not a new change, but a temporary workaround by using the `lt(Expression<A> right)` function, an overload of the NumberExpression class.

It seems difficult to contribute to solve the issue right now. When I have free time, I think I can upload a PR if there is no person in charge.

PR #3367, opened 2022-09-02
changed EntityType#properties to use LinkedHashSet instead of TreeSet

lpandzic:
Type of Set on `EntityType#properties` impacts the order of properties in resulting Q class.
The original line dates back to 2010 and @Shredder121 explained to me that this was most likely due to desire to have deterministic behavior. `TypeElement#getEnclosedElements` is also deterministic so it's not exactly clear why this option was picked.The end result is that instead of having order of columns in Q class being in the same order as fields in original class, they instead end up being in ascending order.

This is problematic when, for instance, using constructor projections. IDE's, Lombok, records etc all respect the order of fields when generating full argument constructors and so parameters are in same order as their respective field counterparts. Because of this - Querydsl users that want to use constructor projections can't merely state they want it over some Q class - they need to make sure that either fields are in ascending order or they need to create constructors for which the parameters have ascending order.

### DIFF
--- a/querydsl-codegen/src/main/java/com/querydsl/codegen/EntityType.java
+++ b/querydsl-codegen/src/main/java/com/querydsl/codegen/EntityType.java
@@ -26,7 +26,6 @@ import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.Map;
 import java.util.Set;
-import java.util.TreeSet;
 import java.util.function.Function;
 import org.jetbrains.annotations.Nullable;
 /**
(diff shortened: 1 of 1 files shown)


lpandzic:
I've build a snapshot locally and run it against https://github.com/infobip/infobip-spring-data-querydsl tests succesfully.
I did have test failures locally with querydsl but I'm not sure is this caused by this change or my local setup so I'll check with CI results.

jwgmeligmeyling (approved):

lpandzic:
I decided to cancel this effort, as I don't have time right now to figure out why tests are exactly failing. I had a hard time figuring out where actual and expected Q class format actually came from - it seems both are generated from something on the fly.

jwgmeligmeyling:
The test failure is just comparison of produced code against a text file. You should be able to just replace the text file.

lpandzic:
But as far as I can see AntJPADomainExporterTest doesn't use any text files - it checks that files generated by AntJPADomainExporter match files that are generated in "../querydsl-jpa/target/generated-test-sources/java" which is also generated...

jwgmeligmeyling:
I see. Then probably we have some means of code generation that still suffers from this "bug". I'll have a look for you.

PR #3382, opened 2022-09-16
JDBC 4.2 support for JSR 310 objects.

wuwen5:
JDBC 4.2 for LocalTime, LocalDate & LocalDateTime


jwgmeligmeyling:
Fixed by #3402

PR #3400, opened 2022-09-30
Fix return type for ARRAY columns

rdicroce:
Currently, when using QueryDSL's SQL module with the H2 database, you get ClassCastExceptions when selecting ARRAY columns. The cause is that the JDBC spec is very ambiguous about what Array#getArray() returns, and QueryDSL is not handling all of the possible cases correctly. This change fixes ArrayType to handle those cases. I have tested the change in my application using H2 and have verified that it fixes the problem.


jwgmeligmeyling:
The fix looks correct, but is there a way we can test this scenario? To prevent a regression when this code is touched later on.

rdicroce:
I added unit tests for both getValue() and setValue() which I think cover all the possible scenarios.

jwgmeligmeyling (approved):

PR #3403, opened 2022-10-04
MS SQL Server: replace old, unmaintained jTDS driver with mssql-jdbc

rdicroce:
The jTDS driver has not been updated in many years and is not JDBC 4.2 (Java 8) compliant. Since Microsoft has open-sourced their official JDBC driver, it makes sense to switch to that instead.

This PR is a prerequisite for #3402, as mentioned in that PR.

In case there are any questions about the use of `sendTimeAsDatetime=false` in the connection string: this is necessary for tests related to `java.sql.Time` to work, as per footnote 1 in the [official MS docs](https://learn.microsoft.com/en-us/sql/connect/jdbc/using-basic-data-types?view=sql-server-ver16).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -47,7 +47,7 @@
     <postgresql.version>42.3.4</postgresql.version>
     <oracle.version>11.2.0.4</oracle.version>
     <mysql.version>8.0.29</mysql.version>
-    <jtds.version>1.3.1</jtds.version>
+    <mssql.version>11.2.1.jre8</mssql.version>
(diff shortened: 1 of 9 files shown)


rdicroce:
The new driver revealed a bug in UtilDateType. java.util.Date internally represents the date as milliseconds UTC, but externally the date is represented in the current timezone. Therefore I have changed UtilDateType to be consistent with TimezoneType.

jwgmeligmeyling (approved):

jwgmeligmeyling:
Thanks! 👍🏻

PR #3404, opened 2022-10-06
Oracle: upgrade JDBC driver to latest. Maven coordinates changed.

rdicroce:
Upgraded Oracle to ojdbc8. This is a prerequisite for #3402 as noted in that PR.


jwgmeligmeyling (approved):

rdicroce:
The failing test works fine on my machine. It looks like the problem has something to do with the JVM's time zone, which presumably is different in the GitHub runner. I'll try some workarounds, but it may take a few attempts since I can't reproduce the problem locally.

PR #3500, opened 2023-04-29
Add `clearGroupBy` and `clearHaving` to `QueryMetadata`

davin111:
There are `clearOrderBy()` and `clearWhere()`, but `clearGroupBy()` and `clearHaving()` don't exist.

When someone needs to manipulate `QueryMetadata`, this unbalance may cause inconvenience. (Actually, I suffered.) I declared new methods as `default` methods (although it is not pretty) because I'm afraid of causing compile errors to other existing subclasses.

Please give me advice if you have better idea.

### DIFF
--- a/querydsl-core/src/main/java/com/querydsl/core/DefaultQueryMetadata.java
+++ b/querydsl-core/src/main/java/com/querydsl/core/DefaultQueryMetadata.java
@@ -197,6 +197,16 @@ public class DefaultQueryMetadata implements QueryMetadata, Cloneable {
     }
 
     @Override
+    public void clearGroupBy() {
+        groupBy.clear();
+    }
+
+    @Override
+    public void clearHaving() {
+        having = new BooleanBuilder();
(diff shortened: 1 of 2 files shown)


blue-hope:
In addition, we also need a way to manipulate the Boolean predicate of the join on clauses (`joinCondition`) for the same reason. (like `clearJoinCondition` on `JoinExpression`)

jwgmeligmeyling:
Although these missing methods are definitely inconsistent, removing expressions from the query that were explicitly added earlier sounds quite error prone to me. Why dont you just copy the relevant parts of the query over to a new query instance?

davin111:
@jwgmeligmeyling I also think modifying the query is dangerous. However I can't find a proper way to copy `SQLQuery`, retaining its `conn` and modifying its `metadata`.  I want to write codes about applying some converters to predicates of queries. So I just need to modify the query string itself.

PR #3526, opened 2023-05-26
refactor: Easily change complex logic

birariro:
- Removed complex branching statements from getIdentifierColumns()
  - No need to check that "joins" size is 1

- Changed the logic of "handleSelect()" to be more readable

### DIFF
--- a/querydsl-sql/src/main/java/com/querydsl/sql/SQLSerializer.java
+++ b/querydsl-sql/src/main/java/com/querydsl/sql/SQLSerializer.java
@@ -115,41 +115,28 @@ public class SQLSerializer extends SerializerBase<SQLSerializer> {
      */
     @SuppressWarnings("unchecked")
     protected List<Expression<?>> getIdentifierColumns(List<JoinExpression> joins, boolean alias) {
-        if (joins.size() == 1) {
-            JoinExpression join = joins.get(0);
-            if (join.getTarget() instanceof RelationalPath) {
-                return ((RelationalPath) join.getTarget()).getColumns();
-            } else {
-                return Collections.emptyList();
-            }
+        List<Expression<?>> rv = new ArrayList<>();
+		int counter = 0;
 
-        } else {
-            List<Expression<?>> rv = new ArrayList<>();
-            int counter = 0;
-            for (JoinExpression join : joins) {
-                if (join.getTarget() instanceof RelationalPath) {
-                    RelationalPath path = (RelationalPath) join.getTarget();
-                    List<Expression<?>> columns;
-                    if (path.getPrimaryKey() != null) {
-                        columns = path.getPrimaryKey().getLocalColumns();
-                    } else {
-                        columns = path.getColumns();
-                    }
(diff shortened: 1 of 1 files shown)


querydsl-sql/src/main/java/com/querydsl/sql/SQLSerializer.java
```diff
@@ -115,41 +115,28 @@ public List<Path<?>> getConstantPaths() {
      */
     @SuppressWarnings("unchecked")
     protected List<Expression<?>> getIdentifierColumns(List<JoinExpression> joins, boolean alias) {
-        if (joins.size() == 1) {
-            JoinExpression join = joins.get(0);
-            if (join.getTarget() instanceof RelationalPath) {
-                return ((RelationalPath) join.getTarget()).getColumns();
```
jwgmeligmeyling, line 121: This optimisation is gone.

querydsl-sql/src/main/java/com/querydsl/sql/SQLSerializer.java
```diff
@@ -115,41 +115,28 @@ public List<Path<?>> getConstantPaths() {
...
 
-                } else {
-                    // not able to provide a distinct list of columns
-                    return Collections.emptyList();
-                }
-            }
-            return rv;
-        }
+			RelationalPath<?> path = (RelationalPath<?>)join.getTarget();
+			List<Expression<?>> columns =
```
jwgmeligmeyling, line 127: Ternary operators are arguably less readable

querydsl-sql/src/main/java/com/querydsl/sql/SQLSerializer.java
```diff
@@ -162,22 +149,22 @@ public void handle(String template, Object... args) {
...
-                        expr = ExpressionUtils.as(expr, "col__" + name + replacements.size());
-                    }
-                }
-                replacements.add(expr);
-            }
-            return handle(sep, replacements);
-        } else {
-            return handle(sep, expressions);
-        }
+        if (!inSubquery) {
```
jwgmeligmeyling, line 152: Negative conditionals are not considered more readable.

jwgmeligmeyling:
The code worsened in three places. Why fix whats not broken?

querydsl-sql/src/main/java/com/querydsl/sql/SQLSerializer.java
birariro: I didn't realize it was an optimization.

querydsl-sql/src/main/java/com/querydsl/sql/SQLSerializer.java
birariro: Negative conditional statements are not easier to read.
They just make the code easier to read by reducing its depth.

You might want to check out
https://go.dev/doc/effective_go#if
https://stackoverflow.com/questions/2928556/programming-style-should-you-return-early-if-a-guard-condition-is-not-satisfied

PR #3536, opened 2023-06-27
Make UtilUUIDType work more reliably

rdicroce:
mssql-jdbc knows how to return UUID, but if you don't specify a type, it returns String


rdicroce:
This is not stale. It just needs a maintainer to look at it.

PR #3537, opened 2023-06-28
Copy the connection provider when cloning AbstractSQLQuery

rdicroce:
Fixes bug when cloning a query that uses a connection provider. Without this change, the clone has neither a connection nor a connection provider, so the clone cannot be executed.


rdicroce:
This issue is also not stale and needs a maintainer to review it.

PR #3541, opened 2023-07-07
Support more flags in INSERT/UPDATE/DELETE

rdicroce:
This allows adding table hints and the OUTPUT clause in SQL Server


rdicroce:
Just like the other two, this is not stale and needs a maintainer to look at it.

PR #3675, opened 2024-02-02
Test against Java 21

johnktims:


dev-jonghoonpark:
https://github.com/dev-jonghoonpark/querydsl/actions/runs/8249327811

I tried modifying it to pass all tests in java 21.
And I succeeded in modifying it to pass the test in Java 21.

Now I'm modifying it to work with other older java versions.

PR #3718, opened 2024-04-06
Fixing performance issue related with querydsl spatial on hibernate.

dcuestam:
As the representation of the operations were doing an equal true (boolean check), the DB engine is not able to use the spatial index, and instead of performing the intersects with the entities using the spatial index and comparing only the ones indicated by this one, was applying the intersects operation to all the entities.

How was discovered? 

While migrating my application from native spatial queries to querydsl-spatial I found a big downgrade on the performance.  Analysing the differences between the native queries that I had and the ones generated by querydsl-spatial, I was able to indentify that queryDSL spatial was adding for most of the operations that I was interested on an = true clause. It is not needed and this comparison makes the db engine to not to use the spatial index on those queries. 

it can be easily analysed by an EXPLAIN clause, that would show that using = true/false, the spatial index is not used, but without this clause, it is used, having exactly the same result in the queries.

In my case, the performance issue was of having queries executed in few milliseconds to more than 10 minutes in some cases

### DIFF
--- a/querydsl-spatial/src/main/java/com/querydsl/spatial/hibernate/HibernateSpatialSupport.java
+++ b/querydsl-spatial/src/main/java/com/querydsl/spatial/hibernate/HibernateSpatialSupport.java
@@ -41,15 +41,15 @@ public final class HibernateSpatialSupport {
         ops.put(SpatialOps.BOUNDARY, "boundary({0})");
         ops.put(SpatialOps.EXTENT, "extent({0})");
 
-        ops.put(SpatialOps.EQUALS, "equals({0}, {1}) = true");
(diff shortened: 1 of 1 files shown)


jwgmeligmeyling:
The problem is dat a function that returns a boolean cannot be used as predicate in JPQL. So the queries that this produces are semantically not correct JPQL. HQL is a bit more flexible, but this also varies between versions of Hibernate. So there is a potential compatibility issue here (and we will have to do some testing against various versions of Hibernate).

dcuestam:
Hi @jwgmeligmeyling , but the current version of Hibernate in the project is 5.5.3.Final and from this version should be compatible, isn't it?

PR #3738, opened 2024-07-13
Simplify of deprecated usage of typeVisitor

kamilkrzywanski:
Since Java below version 7 is not supported by QueryDSL, I suggest simplifying this class to avoid using reflection. I discovered this workaround while trying to fix [this issue](https://github.com/eclipse-jdt/eclipse.jdt.core/issues/565) (I hope the fix will be effective)


velo:
This was an awesome cleanup change.

Got this sorted on 6.6 release querydsl active fork
https://github.com/OpenFeign/querydsl/releases/tag/6.6

Issue #3740, opened 2024-07-29
querydsl-kotlin-codgen uses Object instead of Any


allantodd:
## Observed vs. expected behavior

I have an entity with this attribute

```
    @Column(columnDefinition = "jsonb")
    var metadata: Map<String, Any> = mutableMapOf()
```

Which results in this attribute in the Q class

```
  public val metadata: MapPath<String, Object, SimplePath<Object>> = this.createMap<String, Object,
      SimplePath<Object>>("metadata", String::class.java, Object::class.java,
      SimplePath::class.java)
```

Resulting in this compiler warning which breaks our build

`This class shouldn't be used in Kotlin. Use kotlin.Any instead.`

The generated code should be

```
  public val metadata: MapPath<String, Any, SimplePath<Any>> = this.createMap<String, Any,
      SimplePath<Any>>("metadata", String::class.java, Any::class.java, SimplePath::class.java)
```
## Steps to reproduce

Use the code generator with the attribute shown above

## Environment

```
------------------------------------------------------------
Gradle 8.9
------------------------------------------------------------

Build time:    2024-07-11 14:37:41 UTC
Revision:      d536ef36a19186ccc596d8817123e5445f30fef8

Kotlin:        1.9.23
Groovy:        3.0.21
Ant:           Apache Ant(TM) version 1.10.13 compiled on January 4 2023
Launcher JVM:  21.0.2 (Eclipse Adoptium 21.0.2+13-LTS)
Daemon JVM:    /Library/Java/JavaVirtualMachines/temurin-21.jdk/Contents/Home (no JDK specified, using current Java home)
OS:            Mac OS X 14.5 aarch64
```

Querydsl version: 5.1.0

Querydsl module: querydsl-kotlin-codegen

Database: Postgres

JDK: 1.9.23

## Additional details

This can be fixed by adding this line to Extensions.kt.  I will raise a pull request shortly

```
    "java.lang.Object" -> Any::class.asClassName()
```

allantodd:
Created this pull request https://github.com/querydsl/querydsl/pull/3741

velo:
FWIW, got this sorted on 6.6 release querydsl active fork
https://github.com/OpenFeign/querydsl/releases/tag/6.6

PR #3741, opened 2024-07-29
fix: 3740 map Java Object to Kotlin Any

allantodd:
Fix for https://github.com/querydsl/querydsl/issues/3740

`java.lang.Object` should map to `Any`


velo:
FWIW, got this sorted on 6.6 release querydsl active fork
https://github.com/OpenFeign/querydsl/releases/tag/6.6

PR #3742, opened 2024-08-17
Add Jakarta classifier to the Maven integration docs

hazemessam:
Update the Maven integration docs to support Spring Boot 3.x which is based on Jakarta EE 9+ by using the Jakarta classifier.


hazemessam:
@johnktims Could you please this PR.

velo:
Project is abandoned... I would happily take the change, but there is no point as I don't have docs published yet

https://github.com/OpenFeign/querydsl/

PR #3767, opened 2024-12-21
Optimization of Computational Performance and Simplification of Type Extraction

dukbong:
Prevented unnecessary redundant operations and improved readability by aligning type extraction with the latest code practices.

### DIFF
--- a/querydsl-jpa/src/main/java/com/querydsl/jpa/impl/JPAUtil.java
+++ b/querydsl-jpa/src/main/java/com/querydsl/jpa/impl/JPAUtil.java
@@ -41,6 +41,7 @@ public final class JPAUtil {
             Object val = constants.get(i);
 
             if (val instanceof Param<?> param) {
+                Param<?> param = (Param<?>) val;
                 val = params.get(val);
                 if (val == null) {
                     throw new ParamNotSetException(param);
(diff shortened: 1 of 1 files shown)


querydsl-jpa/src/main/java/com/querydsl/jpa/impl/JPAUtil.java
```diff
@@ -40,20 +40,17 @@ public static void setConstants(Query query, List<Object> constants, Map<ParamEx
         for (int i = 0; i < constants.size(); i++) {
             Object val = constants.get(i);
 
-            if (val instanceof Param) {
-                Param<?> param = (Param<?>) val;
+            if (val instanceof Param<?> param) {
```
jwgmeligmeyling, line 43: This is a Java 16 feature and Querydsl targets Java 8 as minimum right now. While 8 is a bit low, I don't see Querydsl strechting it to Java 16 anytime soon, particularly as Hibernate 6.0 currently targets 11 and above.

dukbong: > This is a Java 16 feature and Querydsl targets Java 8 as minimum right now. While 8 is a bit low, I don't see Querydsl strechting it to Java 16 anytime soon, particularly as Hibernate 6.0 currently targets 11 and above.
> 
> Unless this is somewhere desugared in the compiler (losing whatever performance improvement is aimed here) I doubt this even compiles with the Maven build :)

Ah! In that case, it seems better to keep the type extraction as is. I’m curious about what you think of reducing unnecessary operations using else if?

jwgmeligmeyling: That change affects not just performance but also the logic, and I think its incorrect. Did you run the test suite?

dukbong: I apologize for the confusion. I will make sure to conduct tests before submitting the PR.

